PR #44, opened 2022-06-16
SIP-43 - Pattern matching with named fields

Jentsch:
State of implementation is here: https://github.com/lampepfl/dotty/pull/15437

### DIFF
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -1,0 +1,306 @@
+---
+layout: sip
+title: SIP-NN - Pattern matching with named fields
+status: submitted
+permalink: /sips/:title.html
+redirect_from: /sips/pending/2021-06-25-named-pattern-matching.html
+---
+
+## History
+
+| Date          | Version       |
+|---------------|---------------|
+| Feb 25th 2022 | Initial Draft |
+
+## Motivation
+
+An readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
+
+Link to work in progress implementation lives here: https://github.com/Jentsch/dotty
+
+## Motivating Examples
+
+Given one wants to use pattern matching on a case class:
+
+```scala
+case class User(name: String, age: Int, city: String)
+
+val user = User(name = "Anna", age = 10, city = "Berlin")
+
+val annasCity = user match
+  case User(name = "Anna", city = c) => c
+```
+
+The Deconstruction allows the same syntax as the construction and seems to be what people intuitively expect. See for example the very first post in [Scala Contributors Thread][contributors-thread] for this topic.
+
+Without names in patterns users have to use underscore a lot. The example above would be written as:
+
+//TODO: how offend does this pattern occur?
+
+```scala
+val annasCity = user match
+  case User("Anna", _, c) => c
+```
+
+This makes it hard which parameter means what, basically the same idea as for named arguments. Adding underscores until the compile is happy is also not a great experience. IDEs help here, by showing the names.
+
+In addition, the code breaks every time a field of `User` gets added, rearranged, or removed.
+In the worst case it breaks silently, if two fields with the same type switch places.
+
+My personal motivation comes from using [`matchPattern` in ScalaTest](https://www.scalatest.org/user_guide/using_matchers#matchingAPattern)
+and got bitten by it every time my data model changed slightly.
+
+## Counter-Examples
+
+This SIP doesn't aim to allow pattern where parameters go into the pattern, e.g:
+
+```scala
+val map = Map("Berlin" -> 10, "Paris" -> 5)
+
+map match
+  case Map("Paris" = five) => five
+```
+
+## Design
+
+THe goal is to allow named parameter in deconstruction as in construction.
+
+### Mixed usage
+
+Mixed patterns, with positional and named patterns are allowed to keep the similarity.
+But they have no motivational use case. Maybe they should be disallowed.
+
+```scala
+  case User("Anna", city = c) => // Mixed usage seems wired
+  case User(_, city = c) => // Leading underscore are especially useless
+```
+
+### Disallow same name twice
+
+```scala
+  case User(city = city1, city = city2) => a // error city is used twice
+  case User(name1, name = name2) => a // error city is used twice
+```
+
+The same should happen if the `User` had a field with a [deprecated name](https://www.scala-lang.org/api/current/scala/deprecatedName.html).
+
+### Order of name and term
+
+Normally an equal sign assigns the result of the right side to the left side. With the proposed syntax that's not the case. However, for most, if not all, people that's the correct order.
+
+This order seems the requires a single look ahead in the parser.
+
+### Desugaring
+
+> One important principle of Scala’s pattern matching design is that case classes should be abstractable. I.e. we want to be able to represent the abilities of a case class without exposing the case class itself. That also allows code to evolve from a case class to a regular class or a different case class while maintaining the interface of the old case class. [Martin Odersky](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/52)
+
+* Allow the same feature of case classes:
+  * Provide arbitrary names
+  * Enable [`@deprecatedName`](https://www.scala-lang.org/api/current/scala/deprecatedName.html), which leads to multiple names for a single field. Using both should be an error.
+* Maybe provide additional named fields. For example for our `User`:
+
+  ```scala
+  case User(ageInDays = 3650) => ...
+  ```
+
+  But this could be an awful feature.
+* Maybe leave fields without names
+
+This SIP proposes basically to use [shapeless records](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#extensible-records) for the specific purpose of pattern matching.
+For incomplete list of alternatives see [alternative desugaring](#alternative-desugaring) below.
+
+Pro:
+
+* with shapeless records we have a good understanding what we are doing, however we need to tweak the encoding a bit
+* with some helper definitions the usage could be quite short
+* reuse possible
+
+Con:
+
+* opens an can of worms, as this encoding could be useful for other requested features like [named tuples](named-tuple), which would require more thought-out design
+* no support for `@deprecatedName`
+* identifiers are represented with string literals
+
+Example for user:
+
+```scala
+object User:
+  def unapply(user: User): Option[
+    (String, Int, String) & 
+    { type Names = ("name", "age", "city") }
+  ] = Some((user.name, user.age, user.city)).asInstanceOf
+```
+
+// TODO: Describe why Names can also be use toplevel.
+
+The reason to pull the names out, instead of keeping them near to their type, is to prevent extractors to accidentally leaking the name.
+
+## Implementation
+
+If a pattern with a name is encountered, the compiler looks up list of provided names and places the trees accordingly.
+
+The list of names either provided by the return type of the unapply method or by the constructor list of the case class.
+
+Example:
+
+```scala
+// a match clause like
+case User(<tree1>, city = <tree2>) => ???
+
+// gets rewritten to:
+case User(
+  <tree1>, // because tree1 is positional pattern
+  _, // because name isn't mentioned
+  <tree2>, // because city is the third parameter of user
+)
+```
+
+This allows us to glace over the details of how pattern matching gets desugared further down the line.
+
+## Drawbacks
+
+### Allow skipping arguments
+
+Whenever a single named argument is used in pattern, the pattern can have fewer arguments than the unapply provides. This is driven by the motivation the make pattern matching extensible.
+But this leads to (arguably small) inconsistency, as pointed out by Lionel Parreaux in the [Scala Contributors Thread](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/44). This could lead users to use a named pattern, just to skip all parameters.
+
+```scala
+  case User(age = _) => "Just wanted to use the extractor, lol!"
+```
+
+### Reordering of user code
+
+
+```scala
+//TODO: Test this
+object User:
+  class UserExtractor(user: User):
+    def _1 = println("Got name"); user.name
+    def _2 = println("Got age"); user.age
+    def _3 = println("Got city"); user.city
+
+    type Names = ("name", "age", "city")
+// Ask for city, then for name
+val User(city = _, name = _) = user
+// But the execution shows:
+// Got name
+// Got age
+// Got city
+```
+
+## Alternatives
+
+### Without any changes to the language
+
+One alternative way of archiving most objectives, that is doable with current Scala, is to use specialized extractors.
+
+```scala
+object User:
+  object age:
+    def unapply(user: User): Option[Int] =
+      Some(user.age)
+
+user match
+  case User.age(y) => y
+```
+
+Libraries like [Monocle][monocle] could be extended to reduce the boilerplate, but some boilerplate would remain.
+In addition, this breaks the intuitive similarity between construction and deconstruction.
+
+### Alternative desugaring
+
+As the above described desugaring has its drawbacks.Here are some alternatives with other drawbacks, and maybe better trade-offs.
+
+#### Use underscore methods
+
+In the sprit of [name based pattern matching](https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html#name-based-match):
+
+```scala
+object User:
+  class UserMatching(user: User):
+    def _1 = user.name
+    def _name = user.name
+
+    @deprecatedName
+    def _oldName = user.name.toUpperCase
+    ...
+  
+  def unapply(user: User) = UserMatching(user)
+```
+
+Pro:
+
+* allows to have more named fields than positional fields
+* allows `@deprecatedName`
+
+Con:
+
+* How to detect that a name means the same as a position? Maybe detect simple patterns like the last line in the example?
+* It's long and verbose, without any shortcuts in sight.
+* An underscore at the beginning of a name isn't an unheard of pattern, even in Scala. This could accidentally expose fields, which weren't suppose to become fields.
+
+#### Annotated `unapply` method
+
+```scala
+object User:
+  @names("name", "age", "city")
+  def unapply(user: User): Some[(String, Age, String)] = ...
+```
+
+Pro:
+
+* simple to implement (was done in the first draft implementation)
+
+Con:
+
+* no clear way of encoding deprecated name
+* annotations are not very scalaish
+* no reuse possible
+
+#### Add type as vehicle
+
+```scala
+object User:
+  type Unapply = ("name", "age", "city")
+  def unapply(user: User): Some((String, Age, String)) = ...
+```
+
+Pro:
+
+* should be easy to implement
+* the type and the method can inherited from traits and allow some kind of reuse
+
+Con:
+
+* the type and the method can be defined in unrelated traits. Only at the use site of the can be checked if the type and the method agree on the arity of the pattern.
+* no clear way of encoding deprecated name
+
+#### Partial destructuring in guards
+
+Lionel Parreaux proposed a more powerful mechanism, where if guards of cases could them self contains destructuring patterns.
+
+```scala
+  case user: User if Age(years) <- user.age => years
+  case User(age = Age(years)) => years // both cases do the same thing
+```
+
+His proposal is strictly more powerful, but arguably less intuitive. Both, Pattern matching with named fields and Partial destructuring in guards could be implemented along each other. Named fields for simple patterns and destructuring in guards for complex patterns. However, they offer two ways to do the same thing and could lead to lots of bike shedding, if both got added to the language.
+
+## Open questions
+
+It could be useful to add extractors with just named fields to sealed traits and enums.
+
+What would reuse would look like? What is desirable?
+
+## References
+
+* [Scala Contributors Thread][contributors-thread]
+
+* [Monocle][monocle]
+* [Named Tuple Arguments / Anonymous Case Classes][named-tuple]
+* [Partial Destructuring in Guards][partial-destructuring-in-guards]
+
+[monocle]: https://www.optics.dev/Monocle/ "Monocle"
+[named-tuple]: https://contributors.scala-lang.org/t/named-tuple-arguments-anonymous-case-classes/4352
+[contributors-thread]: https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/20 "Scala Contributors thread"
+[partial-destructuring-in-guards]: http://lptk.github.io/programming/2018/12/12/scala-pattern-warts-improvements.html#-partial-destructuring-in-guards
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -107,7 +107,7 @@ This order seems the requires a single look ahead in the parser.
 * Maybe leave fields without names
 
 This SIP proposes basically to use [shapeless records](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#extensible-records) for the specific purpose of pattern matching.
-For incomplete list of alternatives see [alternative desugaring](#alternative-desugaring) below.
+For an incomplete list of alternatives, see [alternative desugaring](#alternative-desugaring) below.
 
 Pro:
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -50,17 +50,6 @@ In the worst case it breaks silently, if two fields with the same type switch pl
 My personal motivation comes from using [`matchPattern` in ScalaTest](https://www.scalatest.org/user_guide/using_matchers#matchingAPattern)
 and got bitten by it every time my data model changed slightly.
 
-## Counter-Examples
-
-This SIP doesn't aim to allow pattern where parameters go into the pattern, e.g:
-
-```scala
-val map = Map("Berlin" -> 10, "Paris" -> 5)
-
-map match
-  case Map("Paris" = five) => five
-```
-
 ## Design
 
 THe goal is to allow named parameter in deconstruction as in construction.
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -79,7 +79,7 @@ But they have no motivational use case. Maybe they should be disallowed.
 
 ```scala
   case User(city = city1, city = city2) => a // error city is used twice
-  case User(name1, name = name2) => a // error city is used twice
+  case User(name1, name = name2) => a // error name is used twice
 ```
 
 The same should happen if the `User` had a field with a [deprecated name](https://www.scala-lang.org/api/current/scala/deprecatedName.html).
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -283,9 +283,7 @@ His proposal is strictly more powerful, but arguably less intuitive. Both, patte
 
 ## Open questions
 
-It could be useful to add extractors with just named fields to sealed traits and enums.
-
-What would reuse would look like? What is desirable?
+May, search for TODO in this file.
 
 ## References
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -259,9 +259,11 @@ object User:
 
 Pro:
 
-* allows to have more named fields than positional fields
+* have more named fields than positional fields
 * allows `@deprecatedName`
 * enabled meaningful names in variadic patterns
+* lazy evaluation patterns where bad field could have type `Nothing` or would throw could, but the pattern would match, as long the bad field isn't mentioned. (This would have many consequences)
+* It's easy to add extractors for maps in another SIP. At least on the encoding side.
 
 Con:
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -91,7 +91,8 @@ For the case that the extractor is a mixure of [product and sequence match](http
 
 ```
 case class Country(name: String, cities: String*)
-val Country(cities = "Berlin") = ???  // ok
+val Country(cities = "Berlin", "Hamburg") = ???  // ok
+val Country(cities = "Berlin", _*) = ??? // ok
 val Country(cities = "Berlin", "Hamburg", name = "Germany") = ??? // not okay
 ```
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -12,13 +12,20 @@ redirect_from: /sips/pending/2021-06-25-named-pattern-matching.html
 |---------------|---------------|
 | Feb 25th 2022 | Initial Draft |
 
-## Motivation
+# Summary
 
-A readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
+With the first [first SIP ever named and default arguments](https://docs.scala-lang.org/sips/named-and-default-arguments.html) a feature disparity between case class constructors and case class deconstructors was introduced. This was an long outstanding issue. (for example https://github.com/scala/bug/issues/6524)
 
-Link to work in progress implementation lives here: https://github.com/Jentsch/dotty
+This SIP introduces a readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
 
-## Motivating Examples
+A WIP PR can be found here: https://github.com/Jentsch/dotty
+
+This SIP is for now structured in three sections: 
+Motivation to visualize the desired outcome. 
+Desgin to enumarate all goals and constrains for this SIP.
+Encoding describe how the design can be implemented, focused on the encoding choosen in the PR.
+
+# Motivating
 
 Given one wants to use pattern matching on a case class:
 
@@ -35,7 +42,7 @@ The Deconstruction allows the same syntax as the construction and seems to be wh
 
 Without names in patterns, users have to use underscore a lot. The example above would be written as:
 
-//TODO: how offend does this pattern occur?
+//TODO: how often does this pattern occur?
 
 ```scala
 val annasCity = user match
@@ -51,11 +58,11 @@ In the worst case, a pattern breaks silently, if two fields with the same type s
 My personal motivation comes from using [`matchPattern` in ScalaTest](https://www.scalatest.org/user_guide/using_matchers#matchingAPattern)
 and got bitten by it every time my data model changed slightly.
 
-## Design
+# Design
 
 The goal is to allow named parameter in deconstruction as in construction.
 
-### Mixed usage
+## Mixed usage
 
 Mixed patterns, with positional and named patterns are allowed to keep the similarity.
 But they have no motivational use case. Maybe they should be disallowed.
@@ -65,7 +72,7 @@ But they have no motivational use case. Maybe they should be disallowed.
   case User(_, city = c) => // Leading underscore are especially useless
 ```
 
-### Disallow same name twice
+## Disallow same name twice
 
 ```scala
   case User(city = city1, city = city2) => a // error city is used twice
@@ -74,13 +81,26 @@ But they have no motivational use case. Maybe they should be disallowed.
 
 The same should happen if the `User` had a field with a [deprecated name](https://www.scala-lang.org/api/current/scala/deprecatedName.html).
 
-### Order of name and term
+## Order of name and term
 
 Normally, an equal sign assigns the result of the right side to the left side. With the proposed syntax, that's not the case. However, for most, if not all, people that's the correct order.
 
 This order seems to require a single look ahead in the parser.
 
-### Desugaring
+## Ignoring of unused parameters
+
+To allow case class to become more extensible, all unused parameters should be ignored. So when a field is added to the cases class old patterns stay source compatible.
+
+```scala
+  case User(city = "Paris") => // is the same as
+  case User(_, _, "Paris") =>
+```
+
+TODO: Find consent here
+
+## Syntax
+
+## Desugaring
 
 > One important principle of Scala’s pattern matching design is that case classes should be abstractable. I.e. we want to be able to represent the abilities of a case class without exposing the case class itself. That also allows code to evolve from a case class to a regular class or a different case class while maintaining the interface of the old case class. [Martin Odersky](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/52)
 
@@ -95,6 +115,8 @@ This order seems to require a single look ahead in the parser.
 
   But this could be an awful feature.
 * Maybe leave fields without names
+
+# Encoding
 
 This SIP proposes basically to use [shapeless records](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#extensible-records) for the specific purpose of pattern matching.
 For an incomplete list of alternatives, see [alternative desugaring](#alternative-desugaring) below.
@@ -113,7 +135,7 @@ Con:
 * no meaningful way to use names within variadic patterns 
   (the encoding enforces to use the names in order. A name like `last` in `Seq` could be implemented with this encoding.)
 
-Example for user:
+Example for an extractor using option and tuple:
 
 ```scala
 object User:
@@ -125,7 +147,20 @@ object User:
     ]
 ```
 
-// TODO: Describe why Names can also be use toplevel.
+Example for an name based extractor:
+
+```scala
+object User:
+  def unapply(user: User) = MyUserExtractor(user)
+
+class UserExtractor(user: User) extends Product:
+  def _1: String = user.name
+  def _2: Int = user.age
+  def _3: String = user.city
+  type Names = ("name", "age", "city")
+  ...
+```
+The `Names` can be pulled either direcly from the return type of the `unapply` method, for name based extractors, or from the type of the returned `Option`. (The case where no tuple, but just a single type is return, e.g. `Option[Int]`, also can be expressed as `Option[Int & { Names = ("age") }]`.)
 
 The reason to pull the names out, instead of keeping them near to their type, is to prevent extractors to accidentally leaking the name.
 
@@ -270,7 +305,7 @@ Con:
 * the type and the method can be defined in unrelated traits. Only at the use site of the can be checked if the type and the method agree on the arity of the pattern.
 * no clear way of encoding deprecated name
 
-#### Partial destructuring in guards
+### Partial destructuring in guards
 
 Lionel Parreaux proposed a more powerful mechanism, where if guards of cases could them self contains destructuring patterns.
 
@@ -281,11 +316,11 @@ Lionel Parreaux proposed a more powerful mechanism, where if guards of cases cou
 
 His proposal is strictly more powerful, but arguably less intuitive. Both, pattern matching with named fields and Partial destructuring in guards could be implemented along each other. Named fields for simple patterns and destructuring in guards for complex patterns. However, they offer two ways to do the same thing and could lead to lots of bike shedding, if both got added to the language.
 
-## Open questions
+# Open questions
 
-May, search for TODO in this file.
+Search for TODO in this file.
 
-## References
+# References
 
 * [Scala Contributors Thread][contributors-thread]
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -117,7 +117,9 @@ object User:
   def unapply(user: User): Option[
     (String, Int, String) & 
     { type Names = ("name", "age", "city") }
-  ] = Some((user.name, user.age, user.city)).asInstanceOf
+  ] = Some((user.name, user.age, user.city)).asInstanceOf[
+      Option[(String, Int, String) & { type Names = ("name", "age", "city") }]
+    ]
 ```
 
 // TODO: Describe why Names can also be use toplevel.
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -45,7 +45,8 @@ val annasCity = user match
 This makes it hard to see which parameter means what. The same issue led to named arguments. Adding underscores until the compiler is happy is also not a great experience. IDEs help here, by showing the names.
 
 In addition, the code breaks every time a field of `User` gets added, rearranged, or removed.
-In the worst case it breaks silently, if two fields with the same type switch places.
+In the best case the broken patterns fail to compile and needs to fixed manually.
+In the worst case a pattern breaks silently, if two fields with the same type switch places.
 
 My personal motivation comes from using [`matchPattern` in ScalaTest](https://www.scalatest.org/user_guide/using_matchers#matchingAPattern)
 and got bitten by it every time my data model changed slightly.
@@ -109,6 +110,8 @@ Con:
 * opens an can of worms, as this encoding could be useful for other requested features like [named tuples](named-tuple), which would require more thought-out design
 * no support for `@deprecatedName`
 * identifiers are represented with string literals
+* no meaningfull way to use names within variadic patterns 
+  (the encoding enforces to use the names in order. A name like `last` in `Seq` could be implmented with this encoding.)
 
 Example for user:
 
@@ -223,6 +226,7 @@ Pro:
 
 * allows to have more named fields than positional fields
 * allows `@deprecatedName`
+* enabled meaningfull names in variadic patterns
 
 Con:
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -18,7 +18,7 @@ With the first [first SIP ever named and default arguments](https://docs.scala-l
 
 This SIP introduces a readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
 
-A WIP PR can be found here: https://github.com/Jentsch/dotty
+A WIP PR can be found here: https://github.com/Jentsch/dotty/tree/pattern-matching-with-named-fields
 
 This SIP is for now structured in three sections: 
 Motivation to visualize the desired outcome. 
@@ -109,9 +109,9 @@ To allow case class to become more extensible, all unused parameters should be i
   case User(_, _, "Paris") =>
 ```
 
-TODO: Find consent here
-
 ### Syntax
+
+
 
 ### Desugaring
 
@@ -145,7 +145,7 @@ Con:
 * opens a can of worms, as this encoding could be useful for other requested features like [named tuples](named-tuple), which would require more thought-out design
 * no support for `@deprecatedName`
 * identifiers are represented with string literals
-* no meaningful way to use names within variadic patterns 
+* no meaningful way to use names for [pure variadic patterns](https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html#sequence-match)
   (the encoding enforces to use the names in order. A name like `last` in `Seq` could be implemented with this encoding.)
 
 Example for an extractor using option and tuple:
@@ -216,7 +216,7 @@ But this leads to (arguably small) inconsistency, as pointed out by Lionel Parre
   case User(age = _) => "Just wanted to use the extractor, lol!"
 ```
 
-#### UNexpected execution of extractor code
+#### Unexpected execution of extractor code
 
 
 ```scala
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -146,7 +146,7 @@ case User(
 )
 ```
 
-This allows us to glace over the details of how pattern matching gets desugared further down the line.
+This allows us to gloss over the details of how pattern matching gets desugared further down the line.
 
 ## Drawbacks
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -128,7 +128,7 @@ The reason to pull the names out, instead of keeping them near to their type, is
 
 ## Implementation
 
-If a pattern with a name is encountered, the compiler looks up list of provided names and places the trees accordingly.
+If a pattern with a name is encountered, the compiler looks up the list of provided names and places the trees accordingly.
 
 The list of names either provided by the return type of the unapply method or by the constructor list of the case class.
 
--- a/content/2022-02-25-pattern-matching-with-named-fields.md
+++ b/content/2022-02-25-pattern-matching-with-named-fields.md
@@ -14,7 +14,7 @@ redirect_from: /sips/pending/2021-06-25-named-pattern-matching.html
 
 ## Motivation
 
-An readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
+A readable, extensible, and intuitive way to deconstruct case classes in pattern matching.
 
 Link to work in progress implementation lives here: https://github.com/Jentsch/dotty
 
@@ -33,7 +33,7 @@ val annasCity = user match
 
 The Deconstruction allows the same syntax as the construction and seems to be what people intuitively expect. See for example the very first post in [Scala Contributors Thread][contributors-thread] for this topic.
 
-Without names in patterns users have to use underscore a lot. The example above would be written as:
+Without names in patterns, users have to use underscore a lot. The example above would be written as:
 
 //TODO: how offend does this pattern occur?
 
@@ -44,16 +44,16 @@ val annasCity = user match
 
 This makes it hard to see which parameter means what. The same issue led to named arguments. Adding underscores until the compiler is happy is also not a great experience. IDEs help here, by showing the names.
 
-In addition, the code breaks every time a field of `User` gets added, rearranged, or removed.
-In the best case the broken patterns fail to compile and needs to fixed manually.
-In the worst case a pattern breaks silently, if two fields with the same type switch places.
+Additionally, the code breaks every time a field of `User` gets added, rearranged, or removed.
+In the best case, the broken patterns fail to compile and needs to fixed manually.
+In the worst case, a pattern breaks silently, if two fields with the same type switch places.
 
 My personal motivation comes from using [`matchPattern` in ScalaTest](https://www.scalatest.org/user_guide/using_matchers#matchingAPattern)
 and got bitten by it every time my data model changed slightly.
 
 ## Design
 
-THe goal is to allow named parameter in deconstruction as in construction.
+The goal is to allow named parameter in deconstruction as in construction.
 
 ### Mixed usage
 
@@ -76,9 +76,9 @@ The same should happen if the `User` had a field with a [deprecated name](https:
 
 ### Order of name and term
 
-Normally an equal sign assigns the result of the right side to the left side. With the proposed syntax that's not the case. However, for most, if not all, people that's the correct order.
+Normally, an equal sign assigns the result of the right side to the left side. With the proposed syntax, that's not the case. However, for most, if not all, people that's the correct order.
 
-This order seems the requires a single look ahead in the parser.
+This order seems to require a single look ahead in the parser.
 
 ### Desugaring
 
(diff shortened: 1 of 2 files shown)


content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+* identifiers are represented with string literals
+
+Example for user:
+
+```scala
+object User:
+  def unapply(user: User): Option[
+    (String, Int, String) & 
+    { type Names = ("name", "age", "city") }
+  ] = Some((user.name, user.age, user.city)).asInstanceOf
```
sjrd, line 131: Please don't write `.asInstanceOf` without an explicit type. This is bad practice.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+
+// gets rewritten to:
+case User(
+  <tree1>, // because tree1 is positional pattern
+  _, // because name isn't mentioned
+  <tree2>, // because city is the third parameter of user
+)
+```
+
+This allows us to glace over the details of how pattern matching gets desugared further down the line.
```
sjrd, line 158:
```suggestion
This allows us to gloss over the details of how pattern matching gets desugared further down the line.
```
?

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+### Allow skipping arguments
+
+Whenever a single named argument is used in pattern, the pattern can have fewer arguments than the unapply provides. This is driven by the motivation the make pattern matching extensible.
+But this leads to (arguably small) inconsistency, as pointed out by Lionel Parreaux in the [Scala Contributors Thread](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/44). This could lead users to use a named pattern, just to skip all parameters.
+
+```scala
+  case User(age = _) => "Just wanted to use the extractor, lol!"
+```
+
+### Reordering of user code
```
sjrd, line 171: I don't think this is an issue, because Extractors are supposed to be pure. This is already specified today: the pattern matcher is under no obligation to respect the order in which extractors are evaluated, nor how many times they will be evaluated, if any. I don't think it's a problem to extend that requirement to the `def _x` accessors.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+
+The same should happen if the `User` had a field with a [deprecated name](https://www.scala-lang.org/api/current/scala/deprecatedName.html).
+
+### Order of name and term
+
+Normally an equal sign assigns the result of the right side to the left side. With the proposed syntax that's not the case. However, for most, if not all, people that's the correct order.
+
+This order seems the requires a single look ahead in the parser.
+
+### Desugaring
```
sjrd, line 93: It's not clear what happens for extractors with `unapplySeq` methods (and case classes with variadic arguments).

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+```
+
+This allows us to glace over the details of how pattern matching gets desugared further down the line.
+
+## Drawbacks
+
+### Allow skipping arguments
+
+Whenever a single named argument is used in pattern, the pattern can have fewer arguments than the unapply provides. This is driven by the motivation the make pattern matching extensible.
+But this leads to (arguably small) inconsistency, as pointed out by Lionel Parreaux in the [Scala Contributors Thread](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/44). This could lead users to use a named pattern, just to skip all parameters.
```
sjrd, line 165: This particular aspect troubles me. This will cause to silently ignore fields if we forget some.

I strongly believe we should have explicit syntax to deliberately ignore the remaining fields, if that's what we want. Perhaps something like
```scala
case User(name, city = c, _*) =>
```
with the generalized understanding that `_*` deliberately ignores what's left, whether it's single fields or variadic fields.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+Libraries like [Monocle][monocle] could be extended to reduce the boilerplate, but some boilerplate would remain.
+In addition, this breaks the intuitive similarity between construction and deconstruction.
+
+### Alternative desugaring
+
+As the above described desugaring has its drawbacks.Here are some alternatives with other drawbacks, and maybe better trade-offs.
+
+#### Use underscore methods
+
+In the sprit of [name based pattern matching](https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html#name-based-match):
```
sjrd, line 216: Speaking of name-based pattern matching, the main proposal should better describe how it interacts with name-based extractors.

It should also describe how it interacts with single-argument extractors, which don't get a `Tuple` of the elements.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+  case User(age = Age(years)) => years // both cases do the same thing
+```
+
+His proposal is strictly more powerful, but arguably less intuitive. Both, Pattern matching with named fields and Partial destructuring in guards could be implemented along each other. Named fields for simple patterns and destructuring in guards for complex patterns. However, they offer two ways to do the same thing and could lead to lots of bike shedding, if both got added to the language.
+
+## Open questions
+
+It could be useful to add extractors with just named fields to sealed traits and enums.
+
+What would reuse would look like? What is desirable?
```
sjrd, line 293: I don't understand what these mean.

sjrd (commented):
Here are some first comments, without digging *too* much into the details.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+
+// TODO: Describe why Names can also be use toplevel.
+
+The reason to pull the names out, instead of keeping them near to their type, is to prevent extractors to accidentally leaking the name.
+
+## Implementation
+
+If a pattern with a name is encountered, the compiler looks up list of provided names and places the trees accordingly.
+
+The list of names either provided by the return type of the unapply method or by the constructor list of the case class.
```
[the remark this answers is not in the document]
Jentsch: This would simplify things. But I have two concerns:
* This would change the signature of all generated `unapply`s. For most cases this should be fine, because the new type is more specific than the old type. I'm unsure about libraries, like slick, which rely on the unapply method. But thats why we have the community build, right?
* The current implementation would retroactively add named pattern for libraries compiled with an older scala version. That's maybe a good thing.

I'll give a try and see how much breaks.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Counter-Examples
+
+This SIP doesn't aim to allow pattern where parameters go into the pattern, e.g:
+
+```scala
+val map = Map("Berlin" -> 10, "Paris" -> 5)
+
+map match
+  case Map("Paris" = five) => five
```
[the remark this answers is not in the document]
Jentsch: It's removed

Jentsch (commented):
@sjrd Thank you very much for your review. Now I have something to chew on.

content/2022-02-25-pattern-matching-with-named-fields.md
gabro: @sjrd I did not understand what troubles you, normally one destructures to use the extracted value, so what is a case in which forgetting fields causes an issue?

content/2022-02-25-pattern-matching-with-named-fields.md
Jentsch: With the given encoding I see no meaningful names that could be given to variadic argument. (Only `first`, `second`, `third` ... would be possible, but not `last`).
I added this constrain down below to the [Con list, line 113](https://github.com/scala/improvement-proposals/pull/44/files#diff-c3cfd3b89047c3905b5786ac634ef23cbab2b40c4f5b2e3913f17342fb5afebcR235)

content/2022-02-25-pattern-matching-with-named-fields.md
Jentsch: It was an idea in the back of my head, to add unapply methods to sealed trait with the names of the fields of the trait. I removed this section. Maybe the idea will come up again.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,306 @@
...
+
+```scala
+object User:
+  def unapply(user: User): Option[
+    (String, Int, String) & 
+    { type Names = ("name", "age", "city") }
+  ] = Some((user.name, user.age, user.city)).asInstanceOf
+```
+
+// TODO: Describe why Names can also be use toplevel.
```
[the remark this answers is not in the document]
Jentsch: Here I need to describe the mechanism for name-based extractors. Preview from a test:

```scala
object MyUser:
  def unapply(user: User) = MyUserExtractor(user)

class MyUserExtractor(user: User) extends Product:
  def _1: String = user.name
  def _2: Int =user. age
  def _3: String = user.city
  type Names = ("name", "age", "city")
```

dwijnand:
Thanks @Jentsch for the SIP and impl proposal!  This would be an exciting feature to have in the language.

Here's a concern I've read about way back when (https://grokbase.com/t/gg/scala-debate/12ad0n33b0/default-and-named-arguments-in-extractors#20121015a7udamnwcyasnybzqfsq7kmvlq) about this, which I'll reproduce:

> One potential point of contention: should a missing parameter with a default argument be treated as the _ pattern, or as equal to the default arg? I can see the intuitiveness of _, but I wonder if the broken symmetry is worth it:
> 
>```scala
>case class Foo(a: Int, b: Int = 42)
>
>assert(Foo(a = 1) == Foo(a = 1, b = 42))
>assert(Foo(a = 1) != Foo(a = 1, b = 0))
> 
>val Foo(a = x) = Foo(a = 1, b = 42)
>val Foo(a = x) = Foo(a = 1, b = 0)
>```

I'm really on the fence on this one.  If we can find one, I think it would be good to find a way to break this similarity/difference.  I'm thinking maybe with some sort of "and ignore the rest" syntax, as in `Foo(a = x, <ignore the rest>)`, but what to use? `Foo(a = x, _)` looks like one element, `Foo(a = x, _*)`, looks like the last varargs element.  So must it be a new token in the language?

sjrd:
> `Foo(a = x, _*)`, looks like the last varargs element

IMO this is a plus. The varargs element can also be read as "and ignore the rest". So we're truly generalizing it; not inventing a different concept.

dwijnand:
> > `Foo(a = x, _*)`, looks like the last varargs element
> 
> IMO this is a plus. The varargs element can also be read as "and ignore the rest". So we're truly generalizing it; not inventing a different concept.

Yeah, I buy that.  Nice.  So that would mean:

>```scala
>case class Foo(a: Int, b: Int = 42)
>
>assert(Foo(a = 1) == Foo(a = 1, b = 42))
>assert(Foo(a = 1) != Foo(a = 1, b = 0))
> 
>val Foo(a = x, _*) = Foo(a = 1, b = 42)
>val Foo(a = x, _*) = Foo(a = 1, b = 0)
>```

sjrd:
I think so. I see no reason why not. It can even work with no field at all: `case Foo(_*) => ...` (which would be sort of equivalent to `case _: Foo =>`, except it also work if `Foo` is a an extractor and not a type).

SethTisue:
Note that there's been a Scala 2 ticket on this for a long time, with some discussion on it: https://github.com/scala/bug/issues/6524

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,299 @@
...
+```scala
+// a match clause like
+case User(<tree1>, city = <tree2>) => ???
+
+// gets rewritten to:
+case User(
+  <tree1>, // because tree1 is positional pattern
+  _, // because name isn't mentioned
+  <tree2>, // because city is the third parameter of user
+)
```
julienrf, line 149: I am not a reviewer but I would like to add that this solution would make it impossible to manually implement named field patterns.

Currently, it is possible to implement an `unapply` method in the companion object of a (non-case) class to implement an extractor supporting a constructor pattern:

~~~ scala
class Rational(val num: Int, val denom: Int)

object Rational:
  def unapply(rational: Rational): Option[(Int, Int)] = ...
~~~

However, with your solution, there does not seem to be a simple way to manually implement an extractor that supports named field patterns. This could be useful, e.g. for data types that can’t be case classes because they need to support evolutions in a binary compatible way. Or, maybe for extracting things from a Map:

~~~ scala
httpRequest match
  case GET -> Root / "foo" :? QueryParameter(q = query) => println(query)
~~~

content/2022-02-25-pattern-matching-with-named-fields.md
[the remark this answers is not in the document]
Jentsch: I hope the current text explains it and the reasoning as well.

content/2022-02-25-pattern-matching-with-named-fields.md
Jentsch: Maybe I got you wrong, but it is possible. Your example would be:
```
object Rational:
  def unapply(rational: Rational)
      : Option[(Int, Int) & { type Names = ("num", "denom") }] = ...
```

Regarding the possibility of extracting things from a map: I tried to keep it out of this SIP, but keep a way open to add it later - maybe. A hidden design goal, as people got confused - I think.

As I think about this now, the now proposed encoding doesn't fit that well in this regards. I'll add it to the cons list below. Thanks for bringing it up.

dwijnand:
> @sjrd and @dwijnand let me see if I got your concerns regarding skipping of arguments right.
> 
> 1. Given the user case class of the SIP
> 2. A programmer writes a function like:
>    ```scala
>    def postalAddress(user: User) =
>      val User(name = name, city = city) = user
>      name + "\n" + city
>    ```
> 3. Another programmer adds the field `street` to the user
> 4. Nobody see that `street` needs to added to postalAddress and a bug was introduced
> 5. Everybody is unhappy
> 
> Is that what you had in mind? (I'll interpret any kind of positive emoji as yes)

Well, no.

Given the User case class of the SIP, the following:

```scala
case class User(name: String, age: Int, city: String)

val User(name = name, city = city) = user // error: missing age
val User(name = name, age = age, city = city) = user // ok
val User(name = name, city = city, _*) = user // ok
```

Error meaning compile error, not as in throws.

Jentsch:
Okay, but why should this be an error?
```scala
val User(name = name, city = city) = user // error: missing age
```

What kind of programmer interaction do you have in mind, where this error would be beneficial? (The compiler doesn't need it.)

Jentsch:
Thank you for your answers.

For me exhaustivity warnings prevent accidentally missing case that could popup in during runtime. That would lead to an exception and very likely to a bug.

But ignoring a field can't lead to an exception.¹ So I don't see much similarities here.

What I'm missing here?

¹ If the extractor throws when all fields are used it throws also if some fields got ignored. All cases below are semantically identical with current SIP.

```scala
case User(age = a, city = "Amsterdam") =>
case User(age = a, name = _, city = "Amsterdam") =>
case User(_, a, "Amsterdam") =>
```

content/2022-02-25-pattern-matching-with-named-fields.md
julienrf: > Maybe I got you wrong, but it is possible. Your example would be:
> 
> ```scala
> object Rational:
>   def unapply(rational: Rational)
>       : Option[(Int, Int) & { type Names = ("num", "denom") }] = ...
> ```

I think you understood my question well. However, it is not clear to me how, from this `unapply` signature, the compiler would be able to perform the rewrite you showed, which translates the whole thing into positional patterns.

content/2022-02-25-pattern-matching-with-named-fields.md
sjrd: Basically, refactoring.

If I already have some `case User(name, city = c) =>` in my code, and then later I change the definition of `User` to add a new field, there is a pretty good chance that there are at least some `case User(...)` that will have to take this field into account. In many cases, for me, most if not all of my `case User` will need to take it into account. I want the compiler to help me with that refactoring. If we silently ignore any omitted field, refactoring becomes risky.

sjrd:
Here are other reasons for which ignoring remaining fields by default is a bad idea. It will lead to inconsistencies and hard-to-spec behaviors. Consider the definition
```scala
case class User(name: String, age: Int, city: String)
```
If I only use positional arguments, like this:
```scala
case User(n, a) => ... // error
```
the compiler flags it as an error. But if I do
```scala
case User(name = n, age = a) => ...
```
then it's fine. OK, then what if I write:
```scala
case User(n, age = a) => ... // ?
```
should this be flagged as missing arguments? Unclear, since we have a mix of both.

One could say mixing shouldn't be allowed to begin with, but I think that rules out very real use cases like:
```scala
case class VarDef(name: String, mutable: Boolean, tpe: Type)
...
tree match {
  case VarDef(name, mutable = false, tpe) => ...
  case VarDef(name, mutable = true, tpe) => ...
}
```
Even if you disagree with that, there remains the problem of *empty* patterns:
```
case User() => ...
```
Is that 0 positional arguments, in which case it should be an error? Or is that 0 *named* arguments, in which case we should ignore all fields of `User`? There is no correct answer to that question, so this is one more inconsistency.

Basically, ignoring by default under some-but-not-all circumstances (here, depending on whether there are named arguments or not) leads to a series of inconsistencies that will inevitably make the spec, learning experience, and day-to-day reliability more complicated and less robust.

sjrd:
I had an off-line discussion with @julienrf about this proposal just now. He was pushing me to find a way for name-based extractors to, somehow, allow binary-compatibly-evolvable extractors. That would have to be a `def unapply` for which you can add a new field in the future, and it wouldn't binary-break existing `case Extractor(...)` patterns using it. (Note that this is for user-defined extractors, not case classes; case classes are hopelessly non-binary-evolvable anyway.)

We came up with an idea, so I'm just putting it here as a record, but we haven't thought it through all the way.

Assume one can define members (fields and nullary defs) that are somehow marked as usable for named extraction. I'll use the keyword `case` for this:
```scala
class Foo(x: Int, y: String):
  case val age: Int = x
  case def name: String = y
```
A `case val` or `case def` (or even `case lazy val` or `case var`) behaves like a normal `val`/`def` in all situations. The only difference is that it can be used in Product- or name-based pattern matching, as follows:
```scala
object Extractor:
  def unapply(foo: Foo): Some[Foo] = Some(foo)

(??? : Any) match
  case Extractor(age = a, name = n) => ...
```
It works as follows:

1. `Extractor.unapply` returns a `Some[Foo]`, which has the methods `isEmpty: Boolean` and `get: Foo`, so it qualifies for name-based pattern matching, through `Foo`
2. `Foo` has a `case val age` and a `case def name`, so the subpatterns `age = a` and `name = n` are valid.
3. The compilation scheme is something like  `val a: Int = foo.age; val n: String = foo.name`.

A later version of `Foo` can add a third `case val city` without breaking the above desugaring. (it does break *source* compatibility since the pattern is not exhaustive anymore---see my other comments---but not binary)

This scheme has several advantages:

* It uses true members as the names used for name-based extractors, which is simpler to learn than type-level singleton types.
* It automatically supports `@deprecatedName`-style deprecations, as we can define an `@deprecated case def oldName = newName`.
* The encoding for named extractors does not depend on the order or number of `case` members, so they can be added and reordered without breaking use sites.
* We can expressly forbid positional extractors by *only* defining `case` members, and not the `_1`..`_N` methods of positional extraction.
* `case class`es are extended to this scheme in a straightforward way: all fields of the first argument list become `case val`s instead of just `val`s.

It has two significant drawbacks.

First, exhaustivity of a pattern is more difficult to specify. Here are some thoughts:

* If all subpatterns are named, we could say that we must read all `case` members.
* Problem: what about these `@deprecated case` members? They shouldn't be counted, but they must still allow to replace their old names without causing an error. Perhaps it's better to use `@deprecatedName("oldName") case val newName = ...` instead.
* What happens if we mix positional and named? How do we check exhaustivity? Can we mark two members, for example `_1` and `age`, as corresponding to each other somehow?

The second drawback is that this *would* have an impact on the internal implementation in a way that reaches further than `typer`. The named extractors would have to survive until the `patmat` transformation in some form.

soronpo:
This proposal is missing specifying what happens in interaction with a varargs case class or unapply argument.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,336 @@
...
+  (the encoding enforces to use the names in order. A name like `last` in `Seq` could be implemented with this encoding.)
+
+Example for an extractor using option and tuple:
+
+```scala
+object User:
+  def unapply(user: User): Option[
+    (String, Int, String) & 
+    { type Names = ("name", "age", "city") }
+  ] = Some((user.name, user.age, user.city)).asInstanceOf[
```
odersky, line 145: I think this implementation scheme is unacceptably complex. If want to make named-based pattern matching universally available, there should be no penalty to pay for writing extractors. In particular I would hope that the scheme we come with:

 - needs no `asInstanceOf`
 - needs no separate string-based field tuple

odersky (changes requested):
I agree with the motivation: it would be _great_ to have named pattern matching in Scala. 

Some remarks on details:

 - Having `_*` is brilliant. It would help enormously also for code that uses positional matches.  So that's a keeper for sure.
 - At the same time, I am not sure we should _require_ `_*` to be present for non-exhaustive named matches. It seems a bit pedantic to me. The benefit arises only in additional aid when evolving a program, but even that is very limited. Once you have already added a `_*` since you match not on all fields, adding a new field to the class will not alert you anymore. So one cannot bank on the fact that, if one adds a field to a case class, one will automatically get errors for all matches against that class. And if a guarantee is not absolute, what is its value, after all?
 - As stated below I think the proposed implementation strategy is a non-starter. I'll outline a different implementation strategy that comes with different tradeoffs in a separate comment.

odersky:
Here's another more limited and far simpler implementation scheme: Allow named pattern matches only if the pattern is elaborated as a product match over a case class, or an enum case (which expands to a case class), or something that leads via `isEmpty` and `get` to the former.

For case classes we already have all the info we need since we know the correspondence between parameter names and case accessors `_N`. So everything works out of the box. 

Here is the `User` example under the alternate representation scheme:

First, direct variant:
```scala
  class User(val name: String, val age: Int, val city: String)
  object User:
    def unapply(user: User) = UserExtractor(user.name, user.age, user.city)

  case class UserExtractor(name: String, age: Int, city: String)
```
Or, add an `Option`:
```scala
  object User:
    def unapply(user: User) : Option[User] = Some(UserExtractor(user.name, user.age, user.city))
```
Note that there's no specialized ceremony whatsoever. If you want to support named pattern matches, you use case classes or `unapply`s that return case classes.

The _only_ problem is in the case where we have just one parameter. That has nothing to do with named patterns but with the fact how pattern matching is defined today. In fact if we contract `User` to a new single field class `Person`, then the following would work:
```scala
  class Person(val name: String)
  object Person:
    def unapply(person: Person) = PersonExtractor(person.name)
  case class PersonExtractor(name: String)

  @main def Test =
    val p = Person("Peter")
    (p: Any) match
      case Person(n) => println(s"Hello $n") // prints "Hello Peter"
```
But the following variant would print "Hello PersonExtractor(Peter)":
```scala
  object Person:
    def unapply(person: Person): Option{Person] = Some(PersonExtractor(person.name))
```
Again, that has nothing to do with named patterns, it's just (seemingly) irregularity of how pattern matching worrks today in Scala. Basically, pattern matching only performs one decomposition step (which would decompose `Option` in the second case), unless forced by matching against a product pattern of arity > 1.

But the fact remains that if we _wanted_ to support named patterns for single field cases returned from extractors, we can't do that. It's because the rules that allow us to define such named patterns will cause us to match against the container rather than the field. 

On the other hand, one can take a step back an ask: is that so bad? The rules as given are consistent. It's not an arbitrary implementation restriction. And I'd argue it's not a fundamental loss of expressiveness either. If we wanted to write an extractor that returns just a Person's name then we always have the option to call it `PersonNamed`:
```scala
  object PersonNamed:
    def unapply(person: Person): Option{String] = Some(person.name)
```
Then if we match against
```scala
   val PersonNamed(name) = ...
```
there would be no temptation to write this as
```scala
  val PersonNamed(name = name)
```
So I think in practice, the loss of expressiveness would not matter, and in theory we get a cleaner system overall.

julienrf:
> If we wanted to write an extractor that returns just a Person's name then we always have the option to call it `PersonNamed`:
> 
> ```scala
>   object PersonNamed:
>     def unapply(person: Person): Option{String] = Some(person.name)
> ```

It’s less convenient than named-field patterns. How would you implement the equivalent of the following?

~~~ scala
case Person(name = "Julien", email = "julien.richard-foy@epfl.ch", _*) =>
~~~

If I understand correctly your proposal, you would have to implement several extractors and combine them:

~~~ scala
object PersonName:
  def unapply(person: Person): Some[String] = Some(person.name)

object PersonEmail:
  def unapply(person: Person): Some[String] = Some(person.email)

object &:
  def unapply[A](a: A): (A, A)

// and then

(???: Any) match
  case PersonName("Julien") & PersonEmail("julien.richard-foy@epfl.ch") =>
~~~

> Allow named pattern matches only if the pattern is elaborated as a product match over a case class, or an enum case (which expands to a case class), or something that leads via `isEmpty` and `get` to the former.

I think this is a sad restriction. I believe it would be valuable to support named-field patterns with data structures that are _not_ products or case classes. It would be especially valuable with [data types that can evolve in a binary compatible way](https://contributors.scala-lang.org/t/pre-sip-structural-data-structures-that-can-evolve-in-a-binary-compatible-way/5684).

odersky:
@julienrf I think I was not clear. As soon as your case class has more than one member it's no problem. And as as soon as your unapply returns just the case class, even with one member, it's no problem either. The issue is that in Scala there is no way to arrange for `unapply` returning an `Option` of a case class that the pattern will then match the field of the case class.That's simply excluded by the system.  It will match the case class itself. 

But I believe for a user the gap in expressiveness will not even be visible. The rule will be: If the pattern is a product match, and the matched product is a case class, you can use the parameters of the case class as names. There's a nice analogy in that named parameters `(n = ...)` can be used exactly where you could use a field selector `.n` on an `unapply` result.

odersky:
In the context of this PR, I'd like to make another suggestion. Whateven we do, we will have to work this into the pattern matching spec. https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html

The spec is currently not very easy to read, and it would very likely become even less intelligible by adding more stuff to it. So it would be very important to clarify this spec. That's an difficult but essential effort that needs to be taken on by someone of we want to move this forward.

Maybe it should even be a fist step. Since, without an understandable spec what we have, how can we make decisions how to extend it?

chrisandrews-ms:
Hi @odersky - your first and second points here seem contradictory to me:

> * Having `_*` is brilliant. It would help enormously also for code that uses positional matches.  So that's a keeper for sure.
> * At the same time, I am not sure we should _require_ `_*` to be present for non-exhaustive named matches. It seems a bit pedantic to me. The benefit arises only in additional aid when evolving a program, but even that is very limited. Once you have already added a `_*` somewhere since you do not match on all fields, adding a new field to the class will not alert you anymore. So one cannot bank on the fact that, if one adds a field to a case class, one will automatically get errors for all matches against that class. And if a guarantee is not absolute, what is its value, after all?

If we wouldn't require `_*` for non-exhaustive named matches, then what's the rationale for requiring it for non-exhaustive positional matches? Why not just allow them to also compile without an error? Does it change the meaning of previously valid programs? (FWIW I think requiring `_*` for any non-exhaustive match is probably better)

Jentsch:
> Is that 0 positional arguments, in which case it should be an error? Or is that 0 _named_ arguments, in which case we should ignore all fields of `User`? There is no correct answer to that question, so this is one more inconsistency.

@sjrd I see the point of inconsistency here and I'm also not happy about it (but would accept it). I still want to keep incomplete patterns, because they are the main appeal for this SIP. (It's the _extensible_ part in the motivation.) If we throw _intuitive_ syntax under the bus, we could distinguish positional and named patterns with syntax. For example:

```scala
user match {
  case User(_, _, _, _) { name = n } => n
  case User { name = n } => n // here I want to ignore the number of parameters
  case User { name.toUpperCase = "ANNA" } => "anna" // invoke methods on fields or the extractor. This would be become very close to Partial destructuring in guards as proposed by @lptk
}
```

This would solve all inconsistencies regarding positional and named patterns, I think. (This would have many of the pros and cons of "Use underscore methods" or the `case val` desugaring  you proposed above. (they are basically the same idea, aren't they?))

I think many people would prefer the so far proposed syntax more, as the syntax is the same as in method with named arguments. But that leads to inconsistency, or we lose the extensibility. None of the options sound great to me and I don't have an option so far.

Jentsch:
> * Having _* is brilliant. It would help enormously also for code that uses positional matches. So that's a keeper for sure.

I don't see `_*` as necessary so far and as sole author¹ of this SIP and the implementation (up to now) I would defer it to an other SIP. (My thoughts so far: The use case seems to marginal and I don't thinks is a generalization.)

> Maybe it should even be a fist step. Since, without an understandable spec of what's there, how can we make decisions how to extend it?

I'm skeptical about having such a vague precondition for this SIP. Especially nobody (so far) volunteered for this task.

¹ I know, in many cases I just collect bits and pieces from various conversations.

content/2022-02-25-pattern-matching-with-named-fields.md
Jentsch: This scheme isn't a starter, but my third attempt. (An attempt of a pun)

At least `asInstanceOf` can be circumvented with some helpers:

```scala
def unapply(user: User): (String, Int, String) & { type Names = ("name", "age", "city") } =
  ("name", user.name) +:: ("age", user.age) +:: ("city", user.city) +:: empty
```

(Full example here: [runNamedPatternMatching.scala](https://github.com/lampepfl/dotty/blob/01c79b1ca942550ac865cd3cdf267c2c7ec93b79/tests/run/runNamedPatternMatching.scala#L29))

Up to now I've seen two options to encode the names of the pattern in the return type of the extractor: case classes, as you describe below and the fields of the object returned by the extractor (is also described [above by sjrd](https://github.com/scala/improvement-proposals/pull/44#issuecomment-1191235753)). The first one was ruled out very early and the the second one would be bigger undertaking. I'm open other encodings.

Jentsch:
> Here's another more limited and far simpler implementation scheme: Allow named pattern matches only if the pattern is elaborated as a product match over a case class, or an enum case (which expands to a case class), or something that leads via `isEmpty` and `get` to the former.

I fact it's so simple it was the fist thing I build. But I though it wouldn't match you concerns here:

> One important principle of Scala’s pattern matching design is that case classes should be abstractable. I.e. we want to be able to represent the abilities of a case class without exposing the case class itself. That also allows code to evolve from a case class to a regular class or a different case class while maintaining the interface of the old case class. [@odersky](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/52)

Technical your scheme / encoding fits those criteria, but I'm not sure if we want that. E.g. If I remember correctly, option less pattern matching was also introduced to have more performant pattern matching. With this scheme / encoding one would always need to create a case class for pattern matching.

I don't have an opinion here, I just want point a possible conflict.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,299 @@
...
+```scala
+  case User(city = city1, city = city2) => a // error city is used twice
+  case User(name1, name = name2) => a // error name is used twice
+```
+
+The same should happen if the `User` had a field with a [deprecated name](https://www.scala-lang.org/api/current/scala/deprecatedName.html).
+
+### Order of name and term
+
+Normally, an equal sign assigns the result of the right side to the left side. With the proposed syntax, that's not the case. However, for most, if not all, people that's the correct order.
```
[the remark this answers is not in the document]
Jentsch: > ```scala
> case '{ $m: Mirror.ProductOf[T] { type MirroredElemTypes = elementTypes } } =>
>   // introduces the names `m` and `elementTypes`
> ```

Maybe this or [Scala Type-Level Operations – Matt Bovel](https://www.youtube.com/watch?v=6OaW-_aFStA) inspired the newly proposed syntax [down below](https://github.com/scala/improvement-proposals/pull/44#issuecomment-1195881165). Thanks

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,299 @@
...
+
+The Deconstruction allows the same syntax as the construction and seems to be what people intuitively expect. See for example the very first post in [Scala Contributors Thread][contributors-thread] for this topic.
+
+Without names in patterns, users have to use underscore a lot. The example above would be written as:
+
+//TODO: how offend does this pattern occur?
+
+```scala
+val annasCity = user match
+  case User("Anna", _, c) => c
```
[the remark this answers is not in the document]
Jentsch: I added some examples further down.

content/2022-02-25-pattern-matching-with-named-fields.md
```diff
@@ -0,0 +1,386 @@
...
+  case User(_, city = c) => // Leading underscore are especially useless
+```
+
+### Case classes with sequences
+
+For the case that the extractor is a mixure of [product and sequence match](https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html#product-sequence-match) the name of the sequence can only be used at the last position.
+
+```
+case class Country(name: String, cities: String*)
+val Country(cities = "Berlin") = ???  // ok
```
julienrf, line 94: Would the following be valid as well?

~~~ scala
val Country(cities = "Berlin", "Hamburg") = ???
~~~

It is not clear that `"Hamburg"` would be part of the named field pattern, in my opinion.

I don’t really have a better suggestion (other than not supporting named field patterns in sequences), though.

Maybe the following?

~~~ scala
val Country(cities = Seq("Berlin", "Hamburg")*) = ???
~~~

But that looks heavy.

julienrf:
There have been several discussions and propositions going on in this thread. My feeling is that someone would probably need to run some experiments with the possible design options (e.g. do we want to require an explicit `_*` or not, to discard the fields we are not interested in). @Jentsch would you be interested in running such experiments? I also see several “TODO” both in the content of the proposal and in your implementation, what could we do to fix them?

@odersky, you proposed to start with a less ambitious scope (support named field patterns on case classes only), but that raised the following question:

> > One important principle of Scala’s pattern matching design is that case classes should be abstractable. I.e. we want to be able to represent the abilities of a case class without exposing the case class itself. That also allows code to evolve from a case class to a regular class or a different case class while maintaining the interface of the old case class. [@odersky](https://contributors.scala-lang.org/t/pattern-matching-with-named-fields/1829/52)
> 
> Technical your scheme / encoding fits those criteria, but I'm not sure if we want that. […] With this scheme / encoding one would always need to create a case class for pattern matching.

@odersky, what is your final take on this?

On the other hand, @Jentsch it seems that your proposal does not fully address the comment:

> I think this implementation scheme is unacceptably complex. If we want to make named-based pattern matching universally available, there should be no penalty to pay for writing extractors. In particular I would hope that the scheme we come up with:
> 
>   * needs no `asInstanceOf`
> 
>   * needs no separate string-based field tuple

Since your workaround does not seem to be working? See:

https://github.com/lampepfl/dotty/blob/01c79b1ca942550ac865cd3cdf267c2c7ec93b79/tests/run/runNamedPatternMatching.scala#L38

content/2022-02-25-pattern-matching-with-named-fields.md
Jentsch: I think it should be the first syntax, because the syntax is the same for construction.

```scala
val Country(cities = "Berlin", "Hamburg") =
  Country(name = "Germany", cities = "Berlin", "Hamburg")
```

I expanded the example. I played around the pattern below should be valid as well, for the same reason.

```scala
val Country(name = "Germany", "Berlin", "Hamburg") =
  Country(name = "Germany", "Berlin", "Hamburg")
```

Jentsch:
> There have been several discussions and propositions going on in this thread. My feeling is that someone would probably need to run some experiments with the possible design options (e.g. do we want to require an explicit `_*` or not, to discard the fields we are not interested in). @Jentsch would you be interested in running such experiments? I also see several “TODO” both in the content of the proposal and in your implementation, what could we do to fix them?

In principle I'm like the idea of running experiments to better understand a topic. But here I don't see any easy experiment, that would convince anyone, because the concerns where, as far as I can tell, about the long-term effects. Do you have something in mind? @julienrf

For a prototype to experiment with I would:
- mark the feature as experimental or use a [feature flag](https://github.com/lampepfl/dotty/pull/15437/files#diff-45b85c3e7925cdd6cb6619f2aadd274ab6f465ff556e37ffaad4feca1da975dfR74)
- add oderskys encoding with just case classes by changing https://github.com/lampepfl/dotty/pull/15437/files#diff-8893c00063ce469a750bdbdbca40265550e0901a39a8a45e2f83aa6d51230aa0R1429
- fix the recursive bug below
- implement what we want to test

Many ToDos in the implementation are questions, her I would be very glad if someone with with knowledge about the compiler could answer them. 
Most other ToDos are simple tasks which I deferred, until this discussion is over. (E.g. proper error messages.)
One ToDo stands out: sometimes the compiler shows an error about a recursive definition without type annotation and I have no idea why. [See here](https://github.com/lampepfl/dotty/blob/01c79b1ca942550ac865cd3cdf267c2c7ec93b79/tests/neg/negNamedPatternMatching.scala#L54)

> On the other hand, @Jentsch it seems that your proposal does not fully address the following comment:
> 
> > I think this implementation scheme is unacceptably complex. If we want to make named-based pattern matching universally available, there should be no penalty to pay for writing extractors. In particular I would hope that the scheme we come up with:
> > 
> >   * needs no `asInstanceOf`
> > 
> >   * needs no separate string-based field tuple
> 
> Since your workaround does not seem to be working? See:
> 
> https://github.com/lampepfl/dotty/blob/01c79b1ca942550ac865cd3cdf267c2c7ec93b79/tests/run/runNamedPatternMatching.scala#L38

I wanted to show that I'm also not super happy about the encoding, it's just the best thing I came up with. To hide it, I tinkered around to build something like records from shapeless. Yes, this is not done yet. But I'm confident it can work. Afterall records where already implemented in shapeless. It could become part of the stdlib, if we stick with the encoding.

Nevertheless it already works without `asInstanceOf`, addressing one of the two concerns of odersky.

dragos:
While initially I liked the idea of named fields in patterns, I have a few concerns that bother me about this proposal:

1. The more I read the examples, the less readable I find them. Except for when matching against literals, the syntax feels "boilerplatey" because one needs to find another name for every parameter, but often the best name is already taken by the parameter itself:

```
  case User(name = userName, city = userCity) => ..
```

When using positions one can just use the parameter name (and IDEs should make this easy to type):

```
  case User(name, city) => ..
```

2. The most compelling proposal is the ability to ignore fields. However, that has the drawbacks that @sjrd mentioned, and introduces another asymmetry: `User(name = "Joe", city = "Lausanne")` is valid in a pattern match, but not as a constructor (unless the `age` parameter has a default value). 

On the other hand, I do see `_*` is a useful addition to explicitly ignore everything else. I even see _some_ potential use when dealing with JSON coming from REST APIs that may add fields between releases, and often one needs only a small subset of everything that's sent over the wire. However, despite all that, I'd probably be very cautious and add a lint rule to _forbid_ `_*` in any large codebase (because of all the reasons @sjrd mentioned above), and require an explicit decision to allow it.

Bottom line, I don't see pattern matching with named fields as a feature I'd use often at all (but I see some benefits for ignoring fields explicitly).

chrisandrews-ms:
> 1. ... Except for when matching against literals, the syntax feels "boilerplatey" because one needs to find another name for every parameter, but often the best name is already taken by the parameter itself

@dragos, I've been assuming so far that we could use the same names (just like we can when using named arguments on a method call), e.g.

```
 case User(name = name, city = city) => ..
```

@Jentsch I don't see anything in the proposal which explicitly bans or allows using the same names - WDYT?

chrisandrews-ms:
> 2. The most compelling proposal is the ability to ignore fields. However, that has the drawbacks that @sjrd mentioned, and introduces another asymmetry: `User(name = "Joe", city = "Lausanne")` is valid in a pattern match, but not as a constructor (unless the `age` parameter has a default value).

@dragos I agree that it's a kind of asymmetry but it seems ok to me. Anyway I think it would be misleading to mentally associate default parameters with missing arguments in patterns, because given `case class User(name: String, city: String, age: Int = 0)`, we intend `case User(name = "Joe", city = "Lausanne")` to mean `case User("Joe", "Lausanne", _)` not `case User("Joe", "Lausanne", 0)`

dragos:
> @dragos, I've been assuming so far that we could use the same names (just like we can when using named arguments on a method call), e.g.
> 
> ```
>  case User(name = name, city = city) => ..
> ```
@chrisandrews-ms this doesn't read much better, but at least it doesn't matter which one is the bound name :)

The only case that becomes clearer is where the inner pattern is not a variable. I can see some appeal in

```
case User(name = "Joe",..) =>
case User(email = Email(host = "gmail.com", _)) => etc
```

Anything else is a net loss in readability, as far as I can tell.

> @dragos I agree that it's a kind of asymmetry but it seems ok to me. Anyway I think it would be misleading to mentally associate default parameters with missing arguments in patterns, because given case class User(name: String, city: String, age: Int = 0), we intend case User(name = "Joe", city = "Lausanne") to mean case User("Joe", "Lausanne", _) not case User("Joe", "Lausanne", 0)

That's a very good point, and unfortunately, another reason I find this proposal less than optimal (i.e. one more exception that needs to be explained to new users).

My remark on asymmetry was based on the opening paragraph which aims to fix a _feature disparity_ introduced in the named-args SIP. I loosely interpreted that disparity as being an asymmetry, and I found it surprising that it introduces another one.

If we ignore the symmetry argument, I am still trying to figure out what is the *fundamental* improvement that this proposal brings to the language? I haven't seen a very compelling use case yet, readability is at best dubious and the effects of ignoring unmatched parameters are very serious.

Here's my concern: a programmer will interpret `_*` (or, named parameters without an explicit syntax for it) as "I can save a few characters when matching on this case class", but what it _really means_ is: "I will never be interested in any other parameters of this case class, no matter how it evolves in the next decade". In a large codebase with various teams evolving various parts, people often rely on the type system for safe refactoring. This offers a loop hole that I see as a bad trade off in almost any case I can think of. Do we want this in the language? Are there any other benefits than saving a few characters (and trading one asymmetry for another) that I am missing?

soronpo:
> If we ignore the symmetry argument, I am still trying to figure out what is the _fundamental_ improvement that this proposal brings to the language? I haven't seen a very compelling use case yet, readability is at best dubious and the effects of ignoring unmatched parameters are very serious.

In large case classes, I may care about specific fields for pattern matching, but not others.
```scala
case class Foo(arg1: Int, arg2: String, .... arg10 : Int)

f match
  case Foo(_, _, _, 5, _, 20, _, _, _, _) =>
```
I hate all the underscores instead of naming the arguments I care about. So for me, it's mainly being able to pattern match on fields I need and nothing else.

mruzekw:
I second @soronpo's point. 

Positional pattern matching doesn't scale.

I'd very much appreciate a feature like this.

julienrf:
> In principle I like the idea of running experiments to better understand a topic. But here I don't see any easy experiment, that would convince anyone, because the concerns were, as far as I can tell, about the long-term effects. Do you have something in mind? @julienrf

Ideally, it would be good to test scenarios that are closer to the real world. Here are some suggestions. You could look at some existing programs or libraries and count the places where using named fields patterns would be useful. You could write several variants of a match expression, using the feature or not, or some variants of the feature that have been discussed here (e.g. with / without `_*`), show them to a group of people and ask them which ones are the most readable. If you have a concrete example of situation in mind where that feature would have been useful, use it as a detailed scenario to showcase that the feature provides great benefits.

Jentsch:
> 2. The most compelling proposal is the ability to ignore fields. However, that has the drawbacks that @sjrd mentioned, and introduces another asymmetry: User(name = "Joe", city = "Lausanne") is valid in a pattern match, but not as a constructor (unless the age parameter has a default value).

@dragos I don't see that as an issue either. Patter matching and constructors share some similarities, but in the end you need all fields for a constructor but not for a pattern match. Enforcing this symmetry would hinder pattern matching. `Some(_)` is already a valid pattern, but not a valid constructor. Only a subset of all patterns are constructors, but all constructors are valid patterns. (If you ignore constructors with multiple arguments (that's an asymmetry I would like to have fixed))

> ```scala
>   case User(name = name, city = city) => ..
> ```

@chrisandrews-ms you are right, this line is valid. I avoided that in the SIP, because it becomes unclear what's the field of the case class and what's the newly introduced name. (Rust and TypeScript allow writing something like `User(name, city)` in such cases, but that syntax is already used for positional patterns in Scala.)

> > If we ignore the symmetry argument, I am still trying to figure out what is the _fundamental_ improvement that this proposal brings to the language? I haven't seen a very compelling use case yet, readability is at best dubious and the effects of ignoring unmatched parameters are very serious.
> 
> In large case classes, I may care about specific fields for pattern matching, but not others.
> 
> ```scala
> case class Foo(arg1: Int, arg2: String, .... arg10 : Int)
> 
> f match
>   case Foo(_, _, _, 5, _, 20, _, _, _, _) =>
> ```
> 
> I hate all the underscores instead of naming the arguments I care about. So for me, it's mainly being able to pattern match on fields I need and nothing else. (large isn't necessarily that large, and even 4 arguments is enough for me to choose naming the arguments if I could have).

Thank you @soronpo, this what I have in mind, too. The SIP uses shorter examples to explain what and how Scala could change, but case classes with many fields are the reason _why_ names would be useful in pattern matching.

Reusing your example we can now choose between:

```scala
f match
  // 1: current scala
  case Foo(_, _, _, 5, _, 20, _, _, _, _) =>
  // 2: this proposal, like Typescript
  case Foo(arg4 = 5, arg6 = 20) =>
  // 3: like Rust without _*
  case Foo(arg4 = 5, arg6 = 20, arg1 = _, arg2 = _, arg3 = _, arg4 = _, arg5 = _, arg7 = _, arg8 = _, arg9 = _, arg10 = _) =>
  // 4: like Rust with ..
  case Foo(arg4 = 5, arg6 = 20, _*) => 
```

I think it's understandable, why I strongly dislike pattern 3. Even the positional pattern looks better, because the important signals `5` and `20` get lost between all the unused argument names in pattern 3. It's not just a few characters, but around 90 for this small case class with short names.

That leaves me with pattern 2 and 4, which I would prefer. But why than having `_*` in the first place? I don't gain any security if the escape hook `_*` is widely used, because most developer would dislike patter 3 (that's an assumption).

This leaves me only with pattern 2, the current proposal. Or don't change pattern matching at all.

Documentation from Rust: https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#destructuring-structs Is here any Rust programmer how can share his/her knowledge?

soronpo:
> But why than having `_*` in the first place? I don't gain any security if the escape hook `_*` is widely used, because most developer would dislike patter 3 (that's an assumption).

I think that `_*` should only be used for varargs, to remove ambiguity for how many values we pattern match against.
Consider the following example:
```scala
case class Foo(a: Int, b: Int, v: Int*)
val f1234 = Foo(1, 2, 3, 4)
f1234 match
  case Foo(1, 2, 3) => println("This should NOT be printed")
  case Foo(1, 2, 3, _*) => println("This SHOULD be printed")
```
For varargs, if we don't use `_*`, then we are saying to the compiler "we have exactly X number of parameters and these are their values". If we use `_*`, then we are saying "the first X parameters have this values, and the rest we don't care about".

mruzekw:
I agree that 2 is the best option with _* used only for varargs.

Is there a way to add an extension to the Scala compiler to test out?

Jentsch:
> Ideally, it would be good to test scenarios that are closer to the real world. Here are some suggestions. You could look at some existing programs or libraries and count the places where using named fields patterns would be useful. You could write several variants of a match expression, using the feature or not, or some variants of the feature that have been discussed here (e.g. with / without `_*`), show them to a group of people and ask them which ones are the most readable. If you have a concrete example of situation in mind where that feature would have been useful, use it as a detailed scenario to showcase that the feature provides great benefits.

@julienrf I took a first look into the community project and greped for `_, _`. Skimming through the results the most common case was to ignore all fields. But nothing concrete for today. That's a bit unsettling. With around 5000 matches, including many many false positives, I think this feature won't be that important for libraries.

I suspect that application with a rich model of the real world would benefit more. Unfortunately I don't know/ have access to such a application right now. Could someone give me pointer or volunteer? @soronpo, @mruzekw maybe?

Jentsch:
> Is there a way to add an extension to the Scala compiler to test out?

I don't know what you mean with extension, but there is working draft https://github.com/lampepfl/dotty/pull/15437. That's only the compiler, so no IDE support for know.

 @julienrf is it easy to ship / package branches for experiments?

soronpo:
> Could someone give me pointer

https://github.com/DFiantHDL/DFiant/blob/students/compiler/stages/src/main/scala/DFiant/compiler/stages/ExplicitPrev.scala

julienrf:
> is it easy to ship / package branches for experiments?

The easiest way is to locally build and publish the compiler, and then use it.

Otherwise, you have to convince the compiler team to merge your feature behind a flag.

chrisandrews-ms:
> I suspect that application with a rich model of the real world would benefit more.

**Proprietary example**

In the (closed source, Scala 2) application I work on, we have >1,200 occurrences of four consecutive placeholders in pattern matches. Here's an example monstrosity (class names altered to avoid even the tiniest leakage of proprietary code, but syntax unchanged):

```
case Foo(Bar(_, Baz(_, _, _, _, _, _, _), _, _, , _, true, _), _, v) =>
```

With the proposed syntax including `_*` to indicate missing fields, it would be:

```
case Foo(thing = Bar(flavor = Baz(_*), switch = true, _*), value = v, _*) =>
```

I think it's much clearer, especially since there are multiple Boolean parameters to Foo and it's very helpful as a reviewer to know which one is being checked.

If we don't use `_*` to indicate missing fields, it would be:

```
case Foo(thing = Bar(flavor = Baz(_, _, _, _, _, _, _), switch = true), value = v) =>
```

Now `Foo` and `Bar` are more readable but don't have a nice way to write "any `Baz`". Perhaps we'd just write:

```
case Foo(thing = Bar(flavor = _: Baz, switch = true), value = v) =>
```

(And of course we could have done that for `Baz` even without this SIP). But people are reluctant to do that because we all know that type tests are a little bit Javaish, and of course it wouldn't work if Baz was some custom extractor.

**Open source example**

I found plenty of examples of multiple placeholders in [this dusty old compiler](https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/typechecker/Typers.scala) 😄 

Here's one example:

```
case ctor @ DefDef(_, _, _, vparamss, _, cbody @ Block(cstats, cunit)) =>
```

Arguably it's fairly readable already to anyone who works on the compiler because vparamss matches the name of the field, and  we know `Block` is a tree so that last part must be the `rhs`. But I still think it's clearer as:

```
case ctor @ DefDef(vparamss = vparamss, rhs = cbody @ Block(cstats, cunit)) =>
```

**Hypothetical example:**

It's not that unusual to have multiple fields with the same type, e.g. suppose we have `case class Color(red: Double, green: Double, blue: Double)`, then `case Color(1.0, 1.0, 0.0)` conveys a lot less information than `case Color(red = 1.0, green = 1.0, blue = 0.0)`. After all, if we don't think that names are important to see at usage sites, why do we have named parameters for method calls? I guess it could be argued that we should separate introduce types for `Red`, `Green`, and `Blue` but it seems a little overkill.

phiSgr:
Should there be an explicit "ignore other fields" like in Rust (`, ..`) and OCaml (`; _`)?

sjrd:
Hey, sorry for the radio silence here. I've been meaning to reply here for a while, but never got around to doing it.

We had an offline discussion a while ago with @odersky on how to make progress here. Some principles that guided the reasoning were:

* We should not compromise on the property of `case class`es that they can be explained as a desugaring into regular extractors; so no `case class`-only magic.
* We don't want annotations to affect type checking.

I also rehashed some of the ideas with @adpi2 this morning.

The conclusion I've reached from all of that is that we could design things as follows:

* It's based on the idea of `case val`s I exposed earlier. They are looked up within the type `U` (`Product`-based) or `S` (name-based) used [in the reference](https://docs.scala-lang.org/scala3/reference/changed-features/pattern-matching.html).
* We would probably need a different modifier than `case`, because it would be very confusing that a `case val` and a `case object` don't mean the same thing at all. I'll keep using `case` for now.
* Stick to the *order* and number of `case` definitions in a class to match them up with *positional* arguments.
* If a class does *not* define `_N` fields, it cannot be extracted with positional arguments; it can only be extracted with named arguments.
* If it *does* define `_N` fields, the number of `case` defs must match the number of `_N` fields, and their types must corresponding pairwise. This way, we know the link between both for the purposes of exhaustivity.
* The idea of "deprecated `case` defs" is dropped. It's still possible to keep a former `case` def as an `@deprecated` *non*-`case` def to preserve binary and TASTy compatibility; but source compatibility is broken in that situation.
* Named arguments coming before positional arguments in a pattern must correspond to the positional arguments that would be positioned where they are.
* A named argument cannot be used for the variadic components of a variadic extractor.

All of that is orthogonal to whether patterns should enforce exhaustivity by default or not. I have nothing to say about that that I haven't said before.

Example:
```scala
class User(case val lastName: String, case val age: Int, case val city: String) extends Product:
  def _1: String = lastName
  def _2: Int = age
  def _3: String = city

object User:
  def unapply(u: User): User = u // Product-based extractor

x match
  case User(n, a, c) => // all positional
  case User(lastName = n, age = a, city = c) => // all named
  case User(age = a, city = c, lastName = n) => // all named, order changed
  case User(n, age = a, c) => // mix; named arg 'age' is used at its corresponding position
  case User(n, city = c, a) => // illegal; named arg city comes before positional arg 'a' and does not correspond to its position
```
If later, we want to deprecate `lastName` in favor of `name`, we can do:
```scala
class User(case val name: String, case val age: Int, case val city: String) extends Product:
  def _1: String = name
  def _2: Int = age
  def _3: String = city

  @deprecated
  def lastName: String = name
end User
```
while retaining backward binary and TASTy compatibility.

Some illegal definitions:
```scala
// Not same number of `case` defs versus `_N` defs
class User(case val name: String, case val age: Int):
  def _1: String = name
  def _2: Int = age
  def _3: String = city

// Type mismatch between `case` def and `_N` def
class User(case val name: String, case val age: Int):
  def _1: Int = age
  def _2: String = name
```

soronpo:
> * We would probably need a different modifier than `case`, because it would be very confusing that a `case val` and a `case object` don't mean the same thing at all. I'll keep using `case` for now.

How about `product val`? (Soft keyword)

lrytz:
@sjrd IIUC, in your proposal, fields cannot be omitted in a match? To me, allowing to skip the `_`s is probably the most valueable part of this SIP.

sjrd:
We can omit fields, but only if we add a `, _*` at the end. Which we can do whether we use named arguments or not, by the way.

I really don't want to loose the safety of matching of all fields just because I write an explicit name for a `Boolean` field.

Jentsch:
I'm also sorry for the radio silence. Here a bulk of thoughts:

Keywordwise: Could we use `match` as modifier keyword? Like:
```scala
class User(match val name: String, ...)
```

I'll using that syntax for now, to see how much it hurts the eyes.

Pairing the name and the position by their position seems a brittle. Checking the types helps, but even the `User` example here has two fields with the same type and boolean fields are another driver for this SIP. It also ties constructor parameters to the number of the fields in the pattern, which is maybe not always wanted. (I'm unsure about it.)

So here an alternative: If a `match def` or `match val` just references an other `match` it becomes an alias of the referenced definition. This is position independend and allows multiple `match def` to references the same field, including deprecated ones. (The main motivation here is to make the schema position independend, deprecated names are just a benefit.)

```scala
class User(match val name: String, match val age: Int, match val city: String):
  def _1: String = name
  def _2: Int = age
  def _3: String = city

  @deprecated
  match def lastName: String = name
```

The first position (`_1`) and `lastName` is an alias for `name`, because it's just calling `name` / `this.name`. We could a other modifier to make this rule more explicit, but that becomes a bit verbose:

```
class User(match val name: String, ...):
  alias def _1: String = name
  @deprecated
  alias match def lastName: String = name
```

> * If a class does not define _N fields, it cannot be extracted with positional arguments; it can only be extracted with named arguments.

Regardless how we recognize aliases, with the option of having fields without corresponding positional arguments we need have to think about the space engine for exhaustiv checking of patterns. Would it be okay to leave here room for future improvement?

```scala
class User(match val name: String, ...):
  match def firstName: String = name.split(" ")(0)

user match:
  User(firstName = "Anna", ...) => // is equivalent to
  u @ User(...) if u == "Anna" =>  // so exhaustiv checking is limited
```

> * Named arguments coming before positional arguments in a pattern must correspond to the positional arguments that would be positioned where they are.
> * A named argument cannot be used for the variadic components of a variadic extractor.

@sjrd What the reasoning for the first rule? Especially because it seems to enforce the second rule / limitation.

Regarding using the token `_*`: I'm still skeptical about it, because of its two different meanings at the same position. Maybe we can use the same token as Rust here: `..`?

And now to the spicy bit: exhaustivity of names

I have a question: If we make `_*` mendatory, we can still phase out the restriction in a later release? If yes, we could make it part of the SIP and then wait for feedback from the community using the feature.

Or another approach: Instead of deciding at the usage side, would it be better to decide on the declaration side? Like:

```
object User:
  match_all_names def unapply(user: User): User = user
```

@gabro, @chrisandrews-ms, @soronpo: Sorry for not coming back to your examples. I couldn't come up with a questionary that doesn't @sjrd concerns justice. I think every short self containing example is extremly biased against enforcing the usage of all names.

julienrf:
The reviewers of this proposal are now @smarter, @raulraja, and @chrisandrews-ms. I’ve removed @sjrd from the team of reviewers because he will work actively on the proposal with @Jentsch.

anatoliykmetyuk:
@sjrd any progress on this one?

sjrd:
Ouch ... another one that keeps slipping away. :(

> Keywordwise: Could we use `match` as modifier keyword?

Why not? I don't think it collides with any existing syntax, since `match` cannot currently start a statement.

> Pairing the name and the position by their position seems a brittle. [...]
> 
> So here an alternative: If a `match def` or `match val` just references an other `match` it becomes an alias of the referenced definition. This is position independend and allows multiple `match def` to references the same field, including deprecated ones. (The main motivation here is to make the schema position independend, deprecated names are just a benefit.)

The main issue with that is that it makes the *body* of a val/def part of the public API. In terms of library evolution, that is a big problem. We would need a publicly visible way to ensure that.

Regarding brittleness, perhaps we can enforce that *if* the positions match, *then* the bodies are aliasing. That enforces internal consistency, while still not exposing the body to the public API. It still does not allow deprecated things, however.

> Regardless how we recognize aliases, with the option of having fields without corresponding positional arguments we need have to think about the space engine for exhaustiv checking of patterns. Would it be okay to leave here room for future improvement?
> 
> ```scala
> class User(match val name: String, ...):
>   match def firstName: String = name.split(" ")(0)
> 
> user match:
>   User(firstName = "Anna", ...) => // is equivalent to
>   u @ User(...) if u == "Anna" =>  // so exhaustiv checking is limited
> ```

If we can indeed recognize aliases, I don't see why we would need to downgrade exhaustivity like that. What am I missing?

> > * Named arguments coming before positional arguments in a pattern must correspond to the positional arguments that would be positioned where they are.
> > * A named argument cannot be used for the variadic components of a variadic extractor.
> 
> @sjrd What the reasoning for the first rule? Especially because it seems to enforce the second rule / limitation.

The rationale is that it makes sure that a positional argument at position 2 unambiguously correspond to position 2 in the definitions. The same restriction exists when mixing named and positional arguments in calls:
```scala
scala> def foo(a: Int, b: Int, c: Int) = (a, b, c)
def foo(a: Int, b: Int, c: Int): (Int, Int, Int)
                                                                                               
scala> foo(1, b = 2, 3)
val res0: (Int, Int, Int) = (1,2,3)
                                                                                               
scala> foo(1, c = 2, 3)
-- [E171] Type Error: ----------------------------------------------------------
1 |foo(1, c = 2, 3)
  |^^^^^^^^^^^^^^^^
  |missing argument for parameter b of method foo: (a: Int, b: Int, c: Int): (Int, Int, Int)
1 error found
```

> And now to the spicy bit: exhaustivity of names
> 
> I have a question: If we make `_*` mendatory, we can still phase out the restriction in a later release? If yes, we could make it part of the SIP and then wait for feedback from the community using the feature.

Yes, absolutely. We can keep it mandatory at this point, and it would be backward compatible to remove that restriction in the future if so desired.

soronpo:
Is it worth to introduce just `_*` as a separate SIP? It's already worthwhile having, even without named parameters?

SethTisue:
@lrytz @dwijnand are looking into it

odersky:
Another scheme would be possible if we had named tuples. 

Named pattern matches would be supported for

 - case classes
 - extractors that return named tuples

Example:
```scala
 object Address:
   unapply(x: Any): Option[(city: String, zip: Int, street, String)]

   x match
     case Address(city, zip) => ...
```
I am against an explicit marker for named fields. Object deconstruction is usually by selecting fields, and there's no check that we have selected all the fields. Named pattern matches should be analogous. In my mind, it's a _feature_ that we can add more fields to a class and not go through the ceremony of updating all patterns. In a case where we want to insist on complete matches, just define an unapply that returns a regular tuple. If that gets too large and you want to use names, I'd argue your spec is unreasonable. You should not have a great number of fields and at the same time insist that all fields are matched. So in balance I think I prefer partial named matches. 

Note that [F#](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/pattern-matching#record-pattern) does the same thing:

> The record pattern is used to decompose records to extract the values of fields. The pattern does not have to reference all fields of the record; any omitted fields just do not participate in matching and are not extracted.

Dessix:
Coming from Rust and Haskell, I keep running into exactly this syntactic limitation, so I'm definitely glad to see recent activity on it.

The only thing I'd suggest given recent code snippets is that it should also be possible to name the binding separately from the match- for example in cases of an `Option` field being bound only when it's `Some`, or to destructure the field for further conditionality on content without creating any actual bindings.

PR #46, opened 2022-07-15
SIP-46 - Use Scala CLI to implement the 'scala' command

romanowski:
A proposal to use Scala CLI (or its subset) as the new `scala` command.

### DIFF
--- a/content/scala-cli.md
+++ b/content/scala-cli.md
@@ -38,7 +38,7 @@ projects, nor to be extended via a task system. The main advantages of SBT and M
 
 Mill and SBT uses turing complete configuration for build so the complexity of build scripts in theory is unlimited. Scala CLI is configuration-only and that limits the complexity what put a hard cap how complex Scala CLI builds can be.
 
- In our opinion, `scala` command should be first and foremost a command line tool. Requirements for a certain project structure or presence configuration files limit SBT and Mill usability certain use cases related to command line. 
+`scala` command should be first and foremost a command line tool. Requirements for a certain project structure or presence configuration files limit SBT and Mill usability certain use cases related to command line. 
 
 One of the main requirements for the new `scala` commands was speed, flexibility and focus on command-line use cases. Initially, we were considering improving SBT or Mill as well as building Scala CLI on top one. We have quickly realized that getting Mill or SBT to reply within Milliseconds (for cases where no hard work like compilation is require) would be pretty much out of reach. Mill and SBT's codebases are too big to compile them to native image using GraalVM, not to mention problems with dynamic loading and reflection. Adding flexibility when in comes to input sources (e.g. support for Gists) and making the tool that can accept most of the configuration using simple command-line parameters would involve writhing a lot of glue code. That is why we decided to build the tool from scratch based on existing components like coursier, bloop or scalafmt. 
 
@@ -48,13 +48,13 @@ We propose to gradually replace the current `scala`, `scalac` and `scaladoc` com
 
 The complete set of `scala-cli` features can be found in [its documentation](https://scala-cli.virtuslab.org/docs/overview).
 
-Scala CLI brings many features like testing, packaging, exporting to sbt / Mill or upcoming support for publishing micro-libraries. Initially, we may want to limit the set of features available in the `scala` command by default. Scala CLI is a relatively new project and we should battle-proof some of its features before we commit to support them as part of the offical `scala` command. 
+Scala CLI brings many features like testing, packaging, exporting to sbt / Mill or upcoming support for publishing micro-libraries. Initially, we propose to limit the set of features available in the `scala` command by default. Scala CLI is a relatively new project and we should battle-proof some of its features before we commit to support them as part of the official `scala` command. 
 
-Scala CLI offers [multiple native ways to be installed](https://scala-cli.virtuslab.org/install#advanced-installation) so most users should find a suitable method. We would like these packages to become the default `scala` package in most repositories, often replacing existing `scala` packages.
+Scala CLI offers [multiple native ways to be installed](https://scala-cli.virtuslab.org/install#advanced-installation) so most users should find a suitable method. We propose that these packages to become the default `scala` package in most repositories, often replacing existing `scala` packages but the fact how new `scala` command would be installed is not intended to be a part of this SIP.
 
 ### High-level overview
 
-Let us show a few examples where adopting Scala CLI as `scala` command would be a significant improvement ofer current scripts. For this, we have assumed a minial set of features. Each additional  Scala CLI feature included, such as `package`, would add more and more use cases.
+Let us show a few examples where adopting Scala CLI as `scala` command would be a significant improvement over current scripts. For this, we have assumed a minimal set of features (described as MUST have and SHOULD have). Each additional  Scala CLI feature included in the future, such as `package`, would add more and more use cases.
 
 **Using REPL with a 3rd-party dependency**
 
@@ -66,7 +66,7 @@ With Scala CLI, starting a REPL with a given dependency is as simple as running:
 scala-cli repl --dep com.lihaoyi::os-lib:0.7.8
 ```
 
-Compared to Ammonite, default Scala REPLs provided by Scala 2 and 3 - that Scala CLI uses by default - are somewhat limited. However, Scala CLI also offers to start Ammonite instead of the default Scala REPL, by passing `--ammonite` (or `--amm`) option to `scala-cli repl`.
+Compared to Ammonite, default Scala REPLs provided by Scala 2 and 3 - that Scala CLI uses by default - are somewhat limited. However, Scala CLI also offers to start Ammonite instead of the default Scala REPL, by passing `--ammonite` (or `--amm`) option to `scala-cli repl` but we do not propose to include `--ammonite` to the `scala` command not to commit to its maintenance.
 
 Additionally, `scala-cli repl` can also put code from given files / directories / snippets on the class path by just providing their locations as arguments. Running `scala-cli repl foo.scala baz` will compile code from `foo.scala` and the `baz` directory, and put their classes on the REPL class path (including their dependencies, scalac options etc. defined within those files).
 
@@ -74,9 +74,9 @@ Compilation (and running scaladoc as well) benefit in a similar way from the abi
 
 ** Providing reproductions of bugs **
 
-Currently, when reporting a bug in the compiler (or any other Scala-related) repository, users need to provide depencencies, compiler options etc. in comments, create a repository containing a projet with a Mill / sbt configuration to reproduce. In general, testing the reporoduction or working on further minization is not straightworwad.
+Currently, when reporting a bug in the compiler (or any other Scala-related) repository, users need to provide dependencies, compiler options etc. in comments, create a repository containing a projet with a Mill / sbt configuration to reproduce. In general, testing the reproduction or working on further minimization is not straightforward.
 
-"Using directives", provided by Scala CLI give the ablity to include the whole configuration in single file, for example:
+"Using directives", provided by Scala CLI give the ability to include the whole configuration in single file, for example:
 
 ```scala
 //> using platform "native"
@@ -92,49 +92,55 @@ Moreover, Scala CLI provides the ability to run GitHub gists (including multi-fi
 
 ** Watch mode **
 
-When working on a piece of code, it is often useful to have it compiled/run everytime the file is changed, and build tools offer a watch mode for that. This is how most people are using watch mode through a build tool. Scala CLI offers a watch mode for most of its commands (by using `--watch` / `-w` flags).
+When working on a piece of code, it is often useful to have it compiled/run every time the file is changed, and build tools offer a watch mode for that. This is how most people are using watch mode through a build tool. Scala CLI offers a watch mode for most of its commands (by using `--watch` / `-w` flags).
 
 
 ### Specification
 
- In order to be able to expand the functionality of Scala CLI and yet use the same core to power the `scala` command, we propose to include both `scala` and `scala-cli` commands in the installation package. Scala CLI already has a feature to limit accessible sub-commands based the binary name (all sub-commands in `scala-cli`, and a curated list in `scala`). Later, we can include more and more features from `scala-cli` into `scala`. 
+ In order to be able to expand the functionality of Scala CLI and yet use the same core to power the `scala` command, we propose to include both `scala` and `scala-cli` commands in the installation package. Scala CLI already has a feature to limit accessible sub-commands based the binary name (all sub-commands in `scala-cli`, and a curated list in `scala`). On later date, more features from `scala-cli` could be included into `scala` command by additional SIPs or similar processes.
 
-The sub-commands necessary to include in the `scala` to match the functionalities of current commands:
- - `compile`
- - `run`
- - `repl`
- - `doc`
+These sub-commands MUST be included in the the specification of the new `scala` command:
 
-On top of that, we think that the following user-facing sub-commands should also be included:
- - `clean` - to rebuild project from start without any cached steps
- - `setup-ide` - to control the setup of BSP for IDE integration
- - `doctor` - to analyze if everything is installed properly
+ - compile: Compile Scala code
+ - doc: Generate Scaladoc documentation
+ - repl: Fire-up a Scala REPL
+ - run: Compile and run Scala code.
+ - shebang: Like `run`, but more handy from shebang scripts
 
-We also suggest to include additional sub-commands by default: 
- - `fmt` - to format the code using scalafmt
- - `test` - to run tests included in the code
- - `package` - to package the code into various package formats: JAR, "fat"  be, executable  be or even native application or docker image
- - `shebang` - a command useful for scripting, designed to be included in the "shebang" section of scripts
- - `export` - transform current project to sbt / Mill project using the same settings as provided. Useful to evolve prototypes into bigger projects
+These sub-commands SHOULD be included in the the specification of the new `scala` command:
 
-Each of these commands expands what the current scripts offer and can be discussed separately. We can even open a dedicated SIP for each of them.
+ - fmt: Format Scala code
+ - test: Compile and test Scala code
+ - version: Print `scala-cli` version
 
-Beyond that, `scala-cli` offers multiple sub-commands needed to manage itself (e.g. `update`) or its components (e.g. the Bloop server). In most cases,
-these are not user-facing, but still handy. We can elaborate in more detail on what those commands are and why we need them, if necessary.
+
+The subcommand that MAY be included in the specification of the new `scala` command. Those sub-commands are specific to implementation of Scala CLI and provide important, user-facing features like integration with IDE or cleaning up incremental compilation state:
+
+ - about: Print details about this application
+ - bsp: Start BSP server
+ - clean: Clean the workspace
+ - doctor: Print details about this application
+ - help: Print help message
+ - install-completions: Installs completions into your shell
+ - install-home: Install `scala-cli` in a sub-directory of the home directory
+ - setup-ide: Generate a BSP file that you can import into your IDE
+ - uninstall: Uninstall scala-cli - only works when installed by the installation script
+ - uninstall-completions: Uninstalls completions from your shell
+ - update: Update scala-cli - only works when installed by the installation script
+
+Last section of this proposal is the list of options that each sub-command MUST HAVE and SHOULD HAVE for each sub-commands that MUST or SHOULD be included in the specification of the new `scala` command. The options that are specific to the implementation (MAY have) as well as options for implementation specific sub-commands (MAY have) are included in [full specification](https://romanowski.github.io/scala-cli/docs/reference/scala-command/runner-specification).
+ 
+
 
 Scala CLI can also be configured with ["using directives"](https://scala-cli.virtuslab.org/docs/guides/using-directives) - a comment-based configuration syntax that should be placed at the top of Scala files. This allows for self-containing examples within one file since most of the configuration can be provided either from the command line or via using directives (command line has precedence). This is a game changer for use cases like scripting, reproduction, or within the academic scope.
 
 We have described the motivation, syntax and implementation basis in the [dedicated pre-SIP](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700). Currently, we recommend to write using directives as comments, so making them part of the language specification is not necessary at this stage. Moreover, the new `scala` command could ignore using directives in the initial version, however we strongly suggest to include comment-based using directives from the start. 
 
-#### Commands, options and directives
-
-We have generated a complete lists of supported [commands](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/commands.md), [CLI options](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/cli-options.md) and [directives](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/directives.md) that we propose for new `scala` command. Lists contain many options and commands that are marked as `experimental` and/or `internal` and as such, we could change them at any point.
-
-The lists was generated automatically and starting from next release of Scala CLI such documentation will be included on our website. Using that lists, it will be easy to track and spot any changes in exposed APIs.
+Last section of this proposal contains a sumamry of Using Directives syntax as well as list of directives that MUST and SHOULD be supported.
 
 ### Compatibility
 
-Adopting Scala CLI as the new `scala` command, as is, will change some of the behaviour of today's scripts. Some examples:
+Adopting Scala CLI as the new `scala` command, as is, will change some of the behavior of today's scripts. Some examples:
 
 - Scala CLI recognizes tests based on the extension used (`*.test.scala`) so running `scala compile a.scala a.test.scala` will only compile `a.scala`
 - Scala CLI has its own versioning scheme, that is not related to the Scala compiler. Default version used may dynamically change when new Scala version is released. Similarly to Scala 3, we intend for Scala CLI to be backward compatible and this should help mitigate this risk.
@@ -143,14 +149,14 @@ Adopting Scala CLI as the new `scala` command, as is, will change some of the be
 
 ### Other concerns
 
-Scala CLI brings [using directives](https://scala-cli.virtuslab.org/docs/guides/using-directives) and  [conventions to mark the test files](https://scala-cli.virtuslab.org/docs/commands/test#test-sources). We are not sure if both can be accepted as a part of this SIP or we should have seperate SIPs for both (we have opened a [pre-SIP for using directives](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700/15))
+Scala CLI brings [using directives](https://scala-cli.virtuslab.org/docs/guides/using-directives) and  [conventions to mark the test files](https://scala-cli.virtuslab.org/docs/commands/test#test-sources). We suggest to accept both accepted as a part of this SIP but we are ready to open dedicated SIPs for both (we have opened a [pre-SIP for using directives](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700/15))
 
 Scala CLI is an ambitious project and may seem hard to maintain in the long-run.
 
 
 ### Open questions
 
-The main open question for this proposal is wich commands/features should be included by default in the `scala` command. Another aspect is the release cadence: should the new `scala` command follow the current release cadence for Scala CLI (every 2 weeks) or stick to Scala one (every 6 weeks)?
+The release cadence: should the new `scala` command follow the current release cadence for Scala CLI (every 2 weeks) or stick to Scala one (every 6 weeks)?
 
 ## Alternatives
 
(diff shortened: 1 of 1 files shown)


content/scala-cli.md
```diff
@@ -0,0 +1,148 @@
...
+
+Let us show a few examples where adopting Scala CLI as `scala` command would be a significant improvement ofer current scripts. For this, we have assumed a minial set of features. Each additionnal  Scala CLI feature included, such as `package`, would add more and more use cases.
+
+**Using REPL with a 3rd-party dependency**
+
+Currently, to start a Scala REPL with a dependency on the class path, users need to resolve this dependency with all its transitive dependencies (coursier can help here) and pass those to the `scala` command using the `--cp` option. Alternatively, one can create an sbt project including a single dependency and use the `sbt console` task. Ammonite gives a better experience with its magic imports. 
+
+With Scala CLI, starting a REPL with a given dependency is as simple as running:
+
+```scala-cli repl --dep com.lihaoyi::os-lib:0.7.8```
```
bishabosha, line 52:
```suggestion
`scala-cli repl --dep com.lihaoyi::os-lib:0.7.8`
```

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+### Other concerns
+
+Scala CLI brings [using directives](https://scala-cli.virtuslab.org/docs/guides/using-directives) and  [conventions to mark the test files](https://scala-cli.virtuslab.org/docs/commands/test#test-sources). We are not sure if both can be accepted as a part of this SIP or we should have seperate SIPs for both (we have opened a [pre-SIP for using directives](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700/15))
+
+Scala CLI is an ambitious project and may seem hard to maintain in the long-run. // TODO
+
+
+### Open questions
+
+The main open question for this proposal is wich commands/features should be included by default in the `scala` command. Another aspect is the release cadence: should the new `scala` command follow the current release cadence for Scala CLI (every 2 weeks) or stick to Scala one (every 6 weeks)?
```
bishabosha, line 136:
```suggestion
The main open question for this proposal is which commands/features should be included by default in the `scala` command. Another aspect is the release cadence: should the new `scala` command follow the current release cadence for Scala CLI (every 2 weeks) or stick to Scala one (every 6 weeks)?
```

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+## Motivation
+
+The current default Scala scripts: `scala`, `scalac`, `scaladoc` are quite limited. Beside that, the `scala` command that starts the Scala REPL is hardly used by non-power users. 
+
+The current scripts are lacking basic features such as support for resolving dependencies, incremental compilation or support for outputs other than JVM. This forces any user that wants to do anything more than just basic things to learn and use SBT, Mill or an other build tool and that adds to the complexity of learning Scala. 
+
+We observe that the current state of tooling in Scala is limiting creativity, with quite a high cost to create e.g. an application or a script with some dependencies that target Node.js. Many Scala developers are not choosing Scala for their personal projects, scripts, or small applications and we believe that the complexity of setting up a build tool is one of the reasons.
```
ckipp01, line 28: *NOTE*: Not part of the SIP committee, just a passerby with a comment.

> We observe that the current state of tooling in Scala is limiting creativity ... we believe that the complexity of setting up a build tool is one of the reasons

Firstly, I really enjoy `scala-cli`, and I believe it's a much needed tool. I'm really excited to see this here. However, I also think we should try to avoid hyperbole here. This is listed as one of the main points up here in motivation with the crux of the argument being what I quoted up above.

_Here is the minimal build tool setup in IDE including a dependency_

```
ThisBuild / scalaVersion := "3.1.3"

lazy val myproject = project
  .settings(
    libraryDependencies += "com.lihaoyi" %% "pprint" % "0.7.3"
  )
```
A single sbt command can also get you a fresh hello project that a user can just do `sbt run` and see `hello`. The same exists in Mill as well, which is arguably even easier.

_Here is the same with scala-cli_

```
//> using scala "3.1.1"
//> using lib "com.lihaoyi::pprint:0.7.3"
```
Is the difference above really what is limiting creativity in Scala? Is learning the basic sbt structure somehow more complex than learning using directives in scala-cli?

You do mention some good points here:

> with quite a high cost to create e.g. an application or a script with some dependencies that target Node.js

No doubt cross platform support is confusing in build tools, but seeing that this is a giant change in the ecoystem by changing the way someone first interacts with the language I think the SIP should include a couple things.

1. Why was this route chosen rather than improving some of these pain points in the existing tooling?
2. How is learning new concepts introduced by scala-cli such as learning directives and BSP easier to digest than other concepts in tools like sbt or Mill which you mention as a roadblock to creativity.

Mainly because investing time into a brand new tool is also an extremely high cost. Especially in a tooling ecosystem already starved of maintainers and contributors.

None of this means we shouldn't do it, but it just means we should be crystal clear about the _why_, and seeing that this is listed as one of the main motivations of this SIP I'd expect the questions up above also to be crystal clear in the document, and after reading this SIP they weren't answered for me.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+- Scala CLI can also be configured via using directives. Command line options have precedence over using directives, however using directives override defaults. Compiling a file starting with `//> using scala 2.13.8`, without providing a Scala version on the command line, will result in using `2.13.8` rather than the default Scala version. We consider this a feature. However, technically, this is a breaking change.
+
+### Other concerns
+
+Scala CLI brings [using directives](https://scala-cli.virtuslab.org/docs/guides/using-directives) and  [conventions to mark the test files](https://scala-cli.virtuslab.org/docs/commands/test#test-sources). We are not sure if both can be accepted as a part of this SIP or we should have seperate SIPs for both (we have opened a [pre-SIP for using directives](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700/15))
```
sjrd, line 129: I believe this SIP should at least specify the `using` directives that are recognized and applied by the commands available in `scala`, and that affect program semantics. So that includes `using scala`, `using options`, probably the platforms, etc.

Likewise, if we do introduce the `test` command, then we need to decide about the test files convention. (I am a bit worried about that one regarding the `.test.scala` since, as you point out above, it flat-out disregards files provided on the command-line of `scala compile`.)

content/scala-cli.md
sjrd: Given that Scala CLI will dynamically resolve various Scala versions, I don't see the point in tying its release schedule to that of dotc.

sjrd (commented):
If we accept this SIP, then the regular `scala` command that people install on their machines cannot be built by the release process of scala/scala or lampepfl/dotty anymore. How do you see the release processes evolving to adapt to this?

Also, how can we confidently run the `scalac`/`dotc` compiler in standalone mode, without any wrapper around it, to reproduce a compiler issue? Should we still distribute a separate "app" `scala-compiler` that is only a launcher for the standalone compiler?

content/scala-cli.md
romanowski: I am also all for the independent release scheme but @julienrf and others shared some concerns about this, and that is why I have included it as an open question.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+
+Currently, to start a Scala REPL with a dependency on the class path, users need to resolve this dependency with all its transitive dependencies (coursier can help here) and pass those to the `scala` command using the `--cp` option. Alternatively, one can create an sbt project including a single dependency and use the `sbt console` task. Ammonite gives a better experience with its magic imports. 
+
+With Scala CLI, starting a REPL with a given dependency is as simple as running:
+
+```
+scala-cli repl --dep com.lihaoyi::os-lib:0.7.8
+```
+
+Compared to Ammonite, default Scala REPLs provided by Scala 2 and 3 - that Scala CLI uses by default - are somewhat limited. However, Scala CLI also offers to start Ammonite instead of the default Scala REPL, by passing `--ammonite` (or `--amm`) option to `scala-cli repl`.
```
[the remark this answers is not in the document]
romanowski: That is a good point. That is why we were considering making some of the options non-stable so we can drop them at any point, and ammonite may be such an opinion.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Compatibility
+
+Adopting Scala CLI as the new `scala` command, as is, will change some of the behaviour of today's scripts. Some examples:
+
+- `scala repl` needs to be run to start a REPL instead of just `scala`
+- with `scala compile` and `scala doc`, some of the more obscure (or brand new) compile options need to be prefixed with `-O`
+- Scala CLI recognizes tests based on the extension used (`*.test.scala`) so running `scala compile a.scala a.test.scala` will only compile `a.scala`
+- Scala CLI has its own versioning scheme, that is not related to the Scala compiler. Default version used may dynamically change when new Scala version is released.
+- By default, Scala CLI manages its own dependencies (e.g. scalac, zinc, Bloop) and resolves them lazily. This means that the first run of Scala CLI resolves quite some dependencies. Moreover, Scala CLI periodically checks for updates, new defaults accessing online resources (but it is not required to work, so Scala CLI can work in offline environment once setup)
```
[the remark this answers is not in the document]
romanowski: If no version of Scala is provided, Scala CLI will pick the latest supported version, and that can change overnight (usually after the Scala is released).  The supported Scala versions per version of Scala CLI are defined [here](https://github.com/VirtusLab/scala-cli-scala-versions/blob/main/scala-versions-v1.json), and Scala CLI periodically checks it. 

We do not promise any specific behaviour with regard to dynamic versions so that we can change it. I can an option that once code is incrementally compiled, it will not change the version (so there are some files present indicating that the version was picked). In such a case (when we have an indication that code was once compiled) the warning will be printed if a new version is detected. I am not saying this is a good approach, just one of the options we can adopt if we want to increase stability.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+- Scala CLI can also be configured via using directives. Command line options have precedence over using directives, however using directives override defaults. Compiling a file starting with `//> using scala 2.13.8`, without providing a Scala version on the command line, will result in using `2.13.8` rather than the default Scala version. We consider this a feature. However, technically, this is a breaking change.
```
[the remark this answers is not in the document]
romanowski: Scala CLI is cmd-first, and that is why cmd args take precedence. For us, a situation when a user tries to compile code providing a given version of Scala and another is picked would be confusing.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+
+We have described the motivation, syntax and implementation basis in the [dedicated pre-SIP](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700). Currently, we recommend to write using directives as comments, so making them part of the language specification is not necessary at this stage. Moreover, the new `scala` command could ignore using directives in the initial version, however we strongly suggest to include comment-based using directives from the start. 
+
+### Compatibility
+
+Adopting Scala CLI as the new `scala` command, as is, will change some of the behaviour of today's scripts. Some examples:
+
+- `scala repl` needs to be run to start a REPL instead of just `scala`
+- with `scala compile` and `scala doc`, some of the more obscure (or brand new) compile options need to be prefixed with `-O`
```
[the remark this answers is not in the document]
romanowski: We decided to curate the list of options ourselves (with keeping the list of compiler options) instead than fall back on the compiler since we have a lot of options (e.g. `--dep`) that are not forwarded, and we want to pick a types etc. in such options early especially that not all commands may call a compiler.

So this is a case where we decided that the experience of most users (especially the newcomers) is more important than 100% compatibility, especially in the really obscure cases.

romanowski:
>Also, how can we confidently run the scalac/dotc compiler in standalone mode, without any wrapper around it, to reproduce a compiler issue? Should we still distribute a separate "app" scala-compiler that is only a launcher for the standalone compiler?

We offer `--sever:no` to turn off Bloop (and zinc) and the ability to provide a classpath with `--jars` option. In most cases, scala-cli is an binary so we start a JVM for the compiler, so the classpath etc. is also not affected. In such cases, we are relatively close to the current `scalac` / `dotc` so the question is if we really need to be any closer.

Also, I don't mind keeping current scripts as long as they are labelled as advanced and not advertised to average users.

content/scala-cli.md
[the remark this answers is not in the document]
sjrd: I think I wasn't clear before. Command-line args should definitely take precedence over file configuration. This is a universal good practice among all software.

But `using` clauses should be presented as the *normal* way of specifying options. Command-line args should only be used to punctually *override* the configured `using` clauses, but not as the regular way of doing things.

content/scala-cli.md
julienrf: Currently, the setup instructions tell you to run `scala -version` to test your setup:

Screenshot from 2022-07-20 13-38-30

What do we want `scala -version` to return when `scala` is implemented via `scala-cli`? I thought it could make sense to build a `scala` command with every release of Scala, and which would default to the Scala version of that release.

Otherwise, we can decide that the version of `scala` has nothing to do with an actual version of Scala, and in that case, we should document it explicitly.

bjornregnell:
> Should we still distribute a separate "app" `scala-compiler` that is only a launcher for the standalone compiler?

Could not the good old `scalac` command still co-exits with the new Scala-CLI-based `scala`-command?

bjornregnell (changes requested):

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+
+We have described the motivation, syntax and implementation basis in the [dedicated pre-SIP](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700). Currently, we recommend to write using directives as comments, so making them part of the language specification is not necessary at this stage. Moreover, the new `scala` command could ignore using directives in the initial version, however we strongly suggest to include comment-based using directives from the start. 
+
+### Compatibility
+
+Adopting Scala CLI as the new `scala` command, as is, will change some of the behaviour of today's scripts. Some examples:
+
+- `scala repl` needs to be run to start a REPL instead of just `scala`
```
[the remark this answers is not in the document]
bjornregnell: (Forgot to write that I think Scala CLI is awesome!!! :heart: )

content/scala-cli.md
bjornregnell: I can imagine that we in some future could have an official list of curated libs so that we could write something like:
```
//> using scala "3.1.1"
//> using pprint "0.7.3"
```
and *that* level of simplicity would be *really* nice from a beginner perspective, and this SIP is taking a big step towards that level of quality UX. (or is the new cool buzzword rather DX now :smile:  )

content/scala-cli.md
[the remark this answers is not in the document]
odersky: I also wonder why we can't have `scala` as a shorthand for `scala repl`. It's very important that the initial experience is a simple as possible. Many newcomers would not even know what a "repl" was, that's already expert speakl

content/scala-cli.md
ckipp01: > To avoid strange % and %% and : and :: that learners get intimidated by...

O for sure, but again, this doesn't address the main question I had about the effort going into a new tool, vs addressing these pain points in the existing tools. Take the example you have up above, and there actually are discussions about them both in [Mill](https://github.com/com-lihaoyi/mill/pull/1916) and in [sbt](https://github.com/sbt/sbt/discussions/6736). We shy away from talking about it, but in Scala we have a handful of tools that almost all have single maintainers, and sometimes barely that. A lot of pains points we feel _could_ be addressed given focus and people time. Imagine adding up all the hours of various devs that have gone into making `scala-cli`, and imagine putting all those hours towards exiting issues and problems. What would have a greater impact on the ecosystem?

Again, I want to reiterate that I'm not saying this necessarily as a reason _against_ `scala-cli`, I'm happily using `scala-cli`, but it's just a big glaring question mark still, is all the effort going into `scala-cli` (which also forces other tooling to put in effort to support this new tool) a net positive vs if all that effort was put into existing problems and tools. Maybe the answer is yes, starting from a clean slate offers us the greatest impact to make a change for the better in welcoming newcomers, addressing exiting problems, and pushing the boundaries of the existing tooling ecosystem. If that's the case, **then I 110% behind it**. Sign me up. For me personally, as someone who spends a lot of my time trying to help maintain other tools, answering the above question is most what I'm curious about.*

If this was just a project some person or company was working on because it was fun, or something they deemed worth it for them, cool. But I'm harping on the above so much because what we're proposing is a fundamental change to how someone is introduced to Scala, both from the command line and the editor. It's a big deal.

*<sub>I want to recognize that this isn't an easy question, or a new one. Balancing older stable tools with pain points and addressing them vs pushing the boundaries with new shiny ideas that don't fit within the confines of the existing tools. However I believe this just justifies putting the decision and answer to this question right at the forefront to show that it's really been discussed and thought about.</sub>

ckipp01: > or is the new cool buzzword rather DX now 😄

😆 I'm so happy to see this being used here.

sjrd:
To the best of my understanding, here is a summary of the points that should be addressed in order for this SIP to make progress:

* Make it clearer why scala-cli instead of investing into simplifying sbt/Mill (https://github.com/scala/improvement-proposals/pull/46#pullrequestreview-1040234897)
* Propose a list of `using` clauses that would actually be standardized along with this SIP (https://github.com/scala/improvement-proposals/pull/46#discussion_r924630713)
* Make it so that `scala` without any subcommand starts the REPL (https://github.com/scala/improvement-proposals/pull/46#discussion_r925787060)

And I will add that, IMO, the SIP should at least propose a list of subcommands (and their options) that would go into the default `scala`. I understand it's an open question, but we can hardly *decide* about it *without* having any proposed list in the first place.

(This summary doesn't mean that the other, localized comments shouldn't be addressed.)

content/scala-cli.md
julienrf: I would like to emphasize this point with a concrete example of “build definition” for Scala CLI:

https://github.com/VirtusLab/typed-frames/blob/5b6ca450d9c14a0c668283f00e1847298a9af050/publish-conf.scala#L1-L8

I don’t think this is significantly better or simpler than the equivalent with sbt, gradle, mill, etc. It just looks the same: you supply the information to put in the pom.xml that will ultimately be published.

So, my suggestion would be to rephrase the sentence that raised this discussion, to position Scala CLI differently. I don’t think it improves much the landscape of build tools, but I think it improves significantly the current Scala interpreter (the one we get with the `scala` command line tool) because that one can’t be configured with directives.

content/scala-cli.md
sjrd: IMO, `scala -version` should show the version of Scala that will be used if I use `scala` (i.e., `scala repl`) or `scala compile` in the current directory (which means taking `using` clauses into account). The version of the launcher is irrelevant in 99% of use cases, and probably 100% of *beginner* use cases.

For the Python lovers, compare with `py -version`/`py -3 -version` which show "Python 3.10.5" and `py -2 -version` which shows "Python 2.7.18": I get the version of Python that will be used if I run `py`/`py -3` or `py -2`. The version of `py` itself is irrelevant.

content/scala-cli.md
```diff
@@ -0,0 +1,150 @@
...
+these are not user-facing, but still handy. We can elaborate in more detail on what those commands are and why we need them, if necessary.
+
+Scala CLI can also be configured with ["using directives"](https://scala-cli.virtuslab.org/docs/guides/using-directives) - a comment-based configuration syntax that should be placed at the top of Scala files. This allows for self-containing examples within one file since most of the configuration can be provided either from the command line or via using directives (command line has precedence). This is a game changer for use cases like scripting, reproduction, or within the academic scope.
+
+We have described the motivation, syntax and implementation basis in the [dedicated pre-SIP](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700). Currently, we recommend to write using directives as comments, so making them part of the language specification is not necessary at this stage. Moreover, the new `scala` command could ignore using directives in the initial version, however we strongly suggest to include comment-based using directives from the start. 
+
+### Compatibility
```
julienrf, line 116: I believe another incompatibility that should be mentioned here for the sake of completeness is that it may not be possible to have distinct Scala CLI scripts in the same directory, whereas this is possible with the current `scala` command line tool.

If I understand correctly, Scala CLI loads all the .scala files of the current directory, which means that it is impossible to have, in the same directory, one script with `using scala 3.1.3` and another script with `using scala 2.13.8`. They would have to live in separate directories.

SethTisue: fwiw, I haven't observed that to be the case with my `scala-cli shebang` based scripts.

content/scala-cli.md
julienrf: As far as I know, it lacks some features, see e.g. https://github.com/com-lihaoyi/Ammonite/issues/1093

content/scala-cli.md
romanowski: > If I understand correctly, Scala CLI loads all the .scala files of the current directory

This is the case only if `.` is provided as one of the inputs. When user runs `scala-cli A.scala B.scala` the effect is the same regardless what sits in the current directory.

To sum up: it is possible to have `A.scala` using Scala `2.13` and `B.scala` using Scala `3.1` in the same directory without any conflicts as long as both are not run together (using e.g. `scala-cli .`).

content/scala-cli.md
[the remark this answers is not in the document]
romanowski: I have opened an issue to change the default: https://github.com/VirtusLab/scala-cli/issues/1249

There is slight problem with the behaviour of the `scala-cli --help` but I think we should be able to resolve it.

content/scala-cli.md
julienrf: Great, thank you for clarifying, I’m marking the discussion as resolved :)

content/scala-cli.md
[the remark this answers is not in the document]
bjornregnell: Great! I made a comment on the help text in the new issue. (scala help etc should give the help text for the whole tool IMO).

bjornregnell:
If the issues above are addressed, including making the `scala` command as backward compatible as possible/reasonable, with good deprecation/help messages for the things that does not work as before, and keeping `scalac` (and `scaladoc` and `scalap`) as a separate commands as before, and installed by cs, then I'm recommending the committee to vote **accept** on this SIP.

content/scala-cli.md
romanowski: Good point, I created an issue for it: https://github.com/VirtusLab/scala-cli/issues/1250

content/scala-cli.md
soronpo: > @ckipp01 I can imagine that we in some future could have an official list of curated libs so that we could write something like:
> 
> ```
> //> using scala "3.1.1"
> //> using pprint "0.7.3"
> ```

Related: https://contributors.scala-lang.org/t/why-doesnt-scala-have-library-listing/5696

chrisandrews-ms:
I can see big benefits of being able to embed dependency declarations and compiler options in _single-file scripts_ for educational, experimental or bug reproduction use cases. It would also be great to be able to start the repl with libraries automatically resolved from Maven* (the proposed `--dep` flag). I would support this proposal for those use cases.

However, once we're talking about multi-file or multi-directory codebases I think we're just implementing a new build tool here, as others have mentioned. We will end up having to create syntax to deal with all of the usual problems around dependency and compiler flag management (library version conflicts, inconsistent flags etc.)**. For example, what happens right now if two files specify different compiler flags? They then can't be compiled together, so is that an error condition, or do they compile in two separate compiler runs (which means they cannot be bidirectionally interdependent)? What if one specifies flags but the other says nothing? Do the flags apply to both?

If this proposal is accepted, will the using directives become part of the language and subject to SIP review in future? If so, that's a lot of work for the SIP committee (build tools are complicated!), and if not, it seems weird for them to be in Scala source files. Also, will Scala be expected to apply those using directives when invoked via another build tool? Presumably not, but then will they be ignored or will it be an error?

I also notice that there's already support for directives to configure publishing to Maven, which is very useful, but feels a long way from just being a tool to make scripting easier.

----

(* One thing which would really help corporate users is easy configuration of internal Maven replicas which need credentials - or is that kind of usage more advanced that we want to support?)
(** I say all of this as someone who contributed to building an internal corporate build tool 😄 )

bjornregnell:
Yes, we don't want to re-implement sbt... :smile:. As I see it, the scope of this SIP is to make the currently already implemented scala-cli the new `scala`. And it is not a build tool (if we by that mean the build tool use cases that are out-of-scope of scala-cli as explained under "We do not call Scala CLI a build tool" here https://scala-cli.virtuslab.org/ ).

> using directives become part of the language
> [...] or will it be an error?

That's why scala-cli a while ago reverted from directives in the language or annotations (all kind of solutions were discussed extensively if I remember correctly) to a "magic" comment `//>` and thus will not affect other (build) tools. So I don't see an issue here as it is `scala-cli` "as is" that we are discussing, with some comments to be addressed e.g. making it reasonably backwards compatible with old `scala`. 

This way the general aim here is to provide a better developer experience compared to old `scala` for all those who don't need or want a full-fledged build tool to run their code. So I'm very happy to go forward with this as soon as possible.

@romanowski Is my recollection of the scoping discussion summarized ok above, or did I miss something?

content/scala-cli.md
mpkocher: > After a discussion with Sébastien, here is a suggestion about this point: when someone invokes `scala -version`, simply prints both the version of the runner (ie, scala-cli), and the _current_ version of Scala. The _current_ version of Scala would depend on the working directory (e.g., if there is a Scala project that configures a specific Scala version with a “using” directive, that version should be printed).
> 
> Example:
> 
> ```
> $ scala -version .
> Scala code runner version: 1.2.3
> Scala version (set in ./my-program.scala:2): 3.1.3
> ```
> 
> Or:
> 
> ```
> $ scala -version
> Scala code runner version: 1.2.3
> Default Scala version: 3.2.0
> ```

What is the motivation for mixing up the concerns of the version of `scala-cli` and "inspecting" a project/file? I suspect having multiple versions displayed in the output would create some confusion for new users. 

It would be useful for the default scala version (`--scala-version` option) to  be displayed in the output of `--help` for the relevant subcommands (e.g., `run`, `compile`, etc...). Currently, there's no default shown.

```bash
$> scala-cli compile --help | grep scala-version
  -S, --scala, --scala-version version           Set the Scala version
```

However, a useful error message is displayed if an unsupported value is provided.

```bash
$> scala-cli run --scala-version 7 Example.scala
[error]  Cannot find matching Scala version for '7'
You can only choose one of the 3.x, 2.13.x, and 2.12.x. versions.
The latest supported stable versions are 3.1.3, 2.13.8, 2.12.16.
In addition, you can request compilation with the last nightly versions of Scala,
by passing the 2.nightly, 2.12.nightly, 2.13.nightly, or 3.nightly arguments.
Specific Scala 2 or Scala 3 nightly versions are also accepted.
```

Ideally, the defaults of other command line flags, such as `--jvm`, would be communicated in the output of `--help` as well.

I agree with your suggestion that "inspecting" a project would be a welcomed addition. Perhaps this use case would be better as a new subcommand that could provide more information about a project/file. For example, the configured scala version, the `--scalac-options` used, the resolved dependencies, or a list of main classes. Having an "inspect" functionality  decoupled from `--version` and captured in a subcommand would also enable adding more fine grained information about a project/file over time. Overloading `--version` with this info would probably not age well. 

Currently, I find `scala-cli` sometimes a little tricky to debug when I've configured it incorrectly (e.g., main class), or for cases where it's unclear what default values are being used by `scala-cli`. It could be useful to get feedback from the community to evaluate if an "inspect" functionality could improve the overall user experience.

Overall, I'm finding `scala-cli` to be a very positive addition to the scala ecosystem.

julienrf: > What is the motivation for mixing up the concerns of the version of scala-cli and "inspecting" a project/file?

I guess someone who installed Scala would be very likely to try `scala -version` to know which version of Scala is installed. But since `scala` will be a runner able to run any version of Scala, it makes sense to also show to the user the version of the CLI that is installed.

A similar situation is when someone starts a REPL session. We show both the JDK version and the Scala version because both are useful.

romanowski:
I am sorry for such a long radio silence from my side, but recently,  some personal and work-related things took pretty 120% of my time, leaving no time for open source. Hopefully, from now on, everything is back to normal.

> @romanowski Is my recollection of the scoping discussion summarized ok above, or did I miss something?

Yes, the recollection is really good. I would only add that many projects are somewhere between single-file and corporate projects or open source libraries (e.g. our [own tool](https://github.com/VirtusLab/scala-cli/tree/main/sclicheck) to validate documentation). 

> One thing which would really help corporate users is easy configuration of internal Maven replicas which need credentials - or is that kind of usage more advanced that we want to support?

Yes, we would like that Scala CLI be useful for users in more corporate settings, including internal Maven replicas or some specific proxies. For we have a basic [a page](https://scala-cli.virtuslab.org/docs/reference/proxy-authentication) that describes briefly how to configure proxies, and we [plan](https://github.com/VirtusLab/scala-cli/issues/1109) to have more advance guide in the future (based on an existing feature in Scala CLI and Maven).

As for @sjrd list of points that need to address, I've included the lists of [commands](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/commands.md), [CLI options](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/cli-options.md) and [directives](https://github.com/romanowski/scala-cli/blob/6b7e9614078070ce8ff6cbcb0635b78901228749/website/src/pages/scala-command/directives.md) with a comment in the text of the proposal. Also, next release of Scala CLI should default to `REPL` when running `scala-cli` or `scala` (the PR: https://github.com/VirtusLab/scala-cli/pull/1305).

romanowski:
I have also realized that I should mention that Scala CLI will go into 'restricted' mode that is described in this SIP not only when run as `scala` but also when run as `scala-cli-sip`. The recommended way to test and use Scala CLI, as it will be `scala` command is to create a link to the existing Scala CLI binary named `scala-cli`  and add it to PATH. Unfortunately, I've just discovered that we do not detect properly if Scala CLI is installed as scala when using coursier. We will fix that in the near future (https://github.com/VirtusLab/scala-cli/issues/1307).

romanowski:
Since the latest release, Scala CLI, when run as `scala`, we have limited options and directives (commands were limited for a while). Also, we are hosting [reference](https://scala-cli.virtuslab.org/docs/reference/scala-command/) of [commands](https://scala-cli.virtuslab.org/docs/reference/scala-command/commands), [options](https://scala-cli.virtuslab.org/docs/reference/scala-command/cli-options) and [directives](https://scala-cli.virtuslab.org/docs/reference/scala-command/directives) available when run as `scala`.

ckipp01:
I see that a vote is now already requested.

> To the best of my understanding, here is a summary of the points that should be addressed in order for this SIP to make progress:

> Make it clearer why scala-cli instead of investing into simplifying sbt/Mill (https://github.com/scala/improvement-proposals/pull/46#pullrequestreview-1040234897)

I'd still love to get some insight on this.

romanowski:
@ckipp01 I've included some explanation within the [proposal itself](https://github.com/scala/improvement-proposals/pull/46/commits/572e002997a077a8b1593dcce51e2198627716b3).

content/scala-cli.md
sideeffffect: May I humbly suggest to extend the section as
> Why decided to work on Scala CLI rather then improve existing tools like sbt, Mill _or Ammonite_?

For similar reasons as with sbt or Mill, Ammonite is an already existing tool that shares some similarities with `scala-cli`. It would be helpful to the wider Scala community to write down the reasons why it was better to start from scratch instead of building on top of Ammonite. Thank you :)

bjornregnell:
I recommend to accept on this.

sjrd:
Overall, I recommend to **accept** this proposal. However, I would like to discuss two amendments with the committee:

* I still strongly believe that the `using` directives must *not* be comments, and must instead be specified. They are *binding*; ignoring them results in different semantics! Comments should be comments, that never affect the semantics of a program. Therefore, `using` directives should not be comments.
* The convention for test sources as `*.test.scala` comes "out of the blue" compared to the Scala ecosystem; there is no precedent for this convention. Moreover, as the proposal points out, it causes confusing situations where `scala compile A.scala B.test.scala` does *not*, in fact, compile `B.test.scala`! I think we should rule out this special case, and stick to a) directory-based tests and/or b) `using target "test"`.

soronpo:
> * The convention for test sources as `*.test.scala` comes "out of the blue" compared to the Scala ecosystem; there is no precedent for this convention. Moreover, as the proposal points out, it causes confusing situations where `scala compile A.scala B.test.scala` does _not_, in fact, compile `B.test.scala`! I think we should rule out this special case, and stick to a) directory-based tests and/or b) `using target "test"`.

Good point, but considering that directory adds a little bit drag, maybe a convention like `B$test.scala` is something we can live with.

bjornregnell:
Well, the trade-off includes tools working out of the box as magic comments are ignored. I'm prepared to live with this until a compiler-aware `using bla "bla"` is developed e.g. in another SIP. Let's discuss it and I guess we could accept this but including in the decision a specific request for future enhancements.

soronpo:
In general, regarding the process, I think all reservations should be addressed **before** we discuss the SIP in the committee. We have no process in conditionally approving a SIP. We either approve, reject or send it back to address (new) reservations.

bjornregnell (approved):

sjrd:
> Well, the trade-off includes tools working out of the box as magic comments are ignored.

Except they don't. They're just pretending to, while ignoring semantically meaningful program pieces, altering semantics in the process.

With the same line of reasoning that says that tools "work out of the box" with `//> using`, I could come with a new feature for, say, conditional compilation, and pretend that tools "work":
```scala
//! ifdef JS
println("on JS")
//! else
println("on JVM")
//! endif
```
Or what about assertions à la Java?
```scala
val x = -1
//! assert x >= 0
```
Sure, syntax highlighting would "work", in the sense that it wouldn't panic, but nothing else would. It would be all wrong.

bjornregnell:
I see your point and I agree. But this is a matter if we want to live with it for now, or stall this until there is a (new SIP) proposal for specifying the compilation context in the language.  I can live with this for now.

sjrd:
This SIP was on the agenda for voting today at the SIP meeting. It was voted but was *neither* accepted nor rejected, and is therefore sent back here for revisions.

We don't have a simple one-liner for the feedback regarding what we think should be addressed. So please forgive us while we take a few days before posting again here.

romanowski:
@sjrd would it be possible to provide us with at least some feedback on what needs to be clarified so we can make the proposal ready before the next SIP meeting?

bjornregnell:
> @sjrd would it be possible to provide us with at least some feedback on what needs to be clarified so we can make the proposal ready before the next SIP meeting?

@julienrf Has set a deadline for tomorrow for the SIP committee to clarify feedback and then @sjrd will summarize soon after that. Is that timeline ok with you @romanowski ?

romanowski:
> Is that timeline ok with you @romanowski ?

Yes, it is actually way better then I expected :)

sjrd:
Here we go. Sorry for the delay.

First, I will reiterate that, overall, the committee wants to accept this SIP. In particular, those who have been actively using scala-cli have been emphasizing how much scala-cli is of an improvement on the overall Scala experience. And there is a desire not to drag this SIP any longer than necessary.

That said, at some point, we must take a vote whose answer can only be Yes or No. To that effect, our feedback is the following: we would like the core of this SIP to contain a *specification* for what "the Scala runner" should *at least* support. This would not prevent the Scala runner to offer *more* features. With such a specification, we would know what is the common basis that we can rely on, notably for documentation, and particularly for "permanent" documentation like books.

A specification would have:

* A definitive list of the commands that MUST be supported (MUST as in the RFC style) -- the prevailing opinion would be that a good list would be the "necessary to include" and the "should also be included" from the current proposal, but not the "we also suggest" list.
* For the selected commands, a list of the command-line arguments that MUST be supported, and their meaning. This may be copy-pasted from selected parts of the current documentation of scala-cli, if necessary.
* A specification of the `//> using` directives that MUST be supported by the selected commands, and what is their meaning. Again, this may be copy-pasted from the current documentation of scala-cli.

We hope that the above won't prove too difficult to get to, since we assume that all the actual material is present in the documentation, if only scattered and perhaps surrounded by nice-to-have options that would not be part the SIP.

Other than that, there have been no further comments. In particular, I will note that the following aspects are good as is:
* the motivation is clear now;
* standardizing on the `//>` comments, instead of non-comment `using` clauses, is the favored solution by the majority of the committee.

Don't hesitate to ask for clarifications, if needed.

bjornregnell:
In addition there was also a discussion on if scripts should be included in the spec of a minimal new Scala runner, and my view is that it should, both running `.sc`-files without a main and the excellent "shebang"-scripting. A rationale for including this in the new runner for Scala in this SIP is that the old Scala runner can run scripts (sort of). What do you think @SethTisue ? (referring to our previous discussions, and a ping if there is more to add as input to @romanowski here).

bjornregnell:
@romanowski The next SIP meeting is October 21 and there should be some days (~5-7) for the members to study the proposal before the meeting. I am a volunteered to summarize the pre-discussions at the meeting before the vote and would like to prepare a presentation. What is your time-plan to update the proposal and address the issues summarized above by @sjrd? Don't hesitate to pose any questions here about the requested update as outlined above. (I'm really excited by having Scala CLI out there as the new `scala` :rocket: )

SethTisue:
> What do you think @SethTisue ? 

agree

romanowski:
I was finally able to put out the requested specification. At this moment, I need to confirm with the team if I correctly assigned the categories (`MUST have`, `SHOULD have` and `Implementation detail`) but I do not expect any drastic changes. I was quite conservative in terms of `MUST have` vs `SHOULD have` (e.g. all JS/Native related options are marked as `SHOULD have`).

The specification contains:
 - [Command line specification](https://romanowski.github.io/scala-cli/docs/reference/scala-command/runner-specification) for Scala Runner that includes all commands with all supported options categorized as `MUST have`, `SHOULD have` and `Implementation detail`. This document contains a lot of duplication option-wise since, e.g. `run` supports all options from `compile`
 - [Using directives](https://romanowski.github.io/scala-cli/docs/reference/scala-command/directives) categorized as `MUST have` and `SHOULD have`
 - [A list of options groups](https://romanowski.github.io/scala-cli/docs/reference/scala-command/cli-options) with options categorized as `MUST have` and `SHOULD have`
 - A [list of commands](https://romanowski.github.io/scala-cli/docs/reference/scala-command/commands) categorized as `MUST have` and `SHOULD have` with links to supported options groups.
 
Documents in markdown form, as well as tooling to generate them based on the annotation in the code, are kept on [a branch in my repository](https://github.com/romanowski/scala-cli/tree/runner-specification/website/docs/reference/scala-command) but I can also include the text in the proposal (or some kind of attachment section of this repo).
 
 @sjrd , @bjornregnell please let me know if such specification will allow voting for the SIP.
 
 @bjornregnell, @SethTisue - the specification marks `shebang` as `SHOULD have` but I can also include 3rd category (`NICE TO have` or `SUGGESTED`).

bjornregnell:
@romanowski Thanks for the update! If you merge your branch with the updates soon I think we can vote, or what do you think @julienrf ?

I think the idea from @sjrd was that we will vote on the MUST stuff only, and creating another category "nice to have" will only complicate things. This is the first time in the new SIP process that we vote for a tooling thing and not just a language thing, so forgive us for not being clear on this. We could vote first on the MUST and then on the SHOULD or we could ignore the SHOULD for now. What do you think @sjrd ?

@romanowski   I think the shebang stuff should be in the MUST as it otherwise will be a regression from Scala 2 if no scripts can be run...

julienrf:
> I think we are in time for a vote next friday, or what do you think @julienrf ?

I would like the team of reviewers (you, @sjrd, and @chrisandrews-ms) to agree on the changes before we can put it on the agenda of the next SIP meeting.

romanowski:
@bjornregnell 
>  If you merge your branch with the updates soon (today)?

The documentation is generated dynamically so its content in Scala CLI master may change (e.g. new option may be added). Moreover, it will not be a part of official doc until the next release (since we update the documentation during the release so it is up-to-date with latest release). That is why I proposed to attach the specification in this proposal, so it is fixed. Before that, I am happy to apply any suggestions.

I have also changed `shebang` command to `MUST have`. Do we need any additional for it to be a part of the SIP? (links should reflect that).

bjornregnell:
@romanowski Ok if it is not suitable to merging the branch as it as autogenerated stuff, I think you should change these things in the "Specification" section to implement our requests (that were summarized on a high level by @sjrd  above):

* "The sub-commands necessary to include in the scala to match the functionalities of current commands"  ->  "These sub-commands MUST be included in the `scala` command:" And I also think you should specify here the default behavior of leaving out the sub-command -> starting the REPL etc.

*  "On top of that, we think that the following user-facing sub-commands should also be included:" ->   "These sub-commands are RECOMMENDED to be included in the `scala` command:"

*  "We also suggest to include additional sub-commands by default:" -> I propose to delete this line and just make the subsequent sub-commands part of the recommended commands.

* "We can even open a dedicated SIP for each of them." -> I suggest to remove this sentence and instead include **some kind of specification of each command**, perhaps you can just copy-paste stuff from the generated spec so that we have a fixed and self-contained SIP.  

* "Scala CLI can also be configured with ["using directives"](https://scala-cli.virtuslab.org/docs/guides/using-directives)"  -> As @sjrd asked for above you should include (instead of a link to the docs) some kind of specification here in the SIP of (parts) of the `using` syntax and semantics so that this SIP becomes more of a self-contained specification that can be a documentation of this new approach for the future, e.g. for tooling vendors etc to read about the syntax and semantics in this SIP.

So the over-all intention of our requests is to remove the "conversation with the committee"-phrases like "we think" etc and instead make this SIP more of a self-contained spec that can be a documentation of (any implementation) of the `scala` command from now, and potential future SIPS can relate to this version. Are you following my attempt to make this more concrete? @romanowski 

@sjrd @chrisandrews-ms Please correct me or add stuff if you think I have made some incomplete/disagreeable suggestions on the implementation proposal above.

bjornregnell:
@romanowski And the latest commit that I see here is currently from July so perhaps your changes are not live yet or am I missing something?

lrytz:
@romanowski the updates are looking good to me, thanks!

I have some questions about directives vs command-line options. First, which takes precedence?

Some options are also available as directives, others are not - how is that designed?

Would it make sens to have a `using runnerOpt [option...]` directive? Then the directives could be cut down to the most important ones, like `using lib`. Some directives like `using jsModuleSplitStyleStr` seem rather exotic.

On the other hand, there's no directive equivalent for `--jvm`, is that intended?

Finally I have a few questions about the cli options:
  - [The doc page](https://romanowski.github.io/scala-cli/docs/reference/scala-command/cli-options) doesn't seem to mention `-O` / `--scala-opt`? Could that be included at "options forwarding"?
  - Would it make sense to also forward `-deprecation`, `-feature`, `-release` to the compiler?
  - Could scala-cli support `-Dkey=value` style java properties?
  - Does scala-cli support argument files (`@file` in the compiler)?

content/scala-cli.md
odersky: > Is the difference above really what is limiting creativity in Scala? Is learning the basic sbt structure somehow more complex than learning using directives in scala-cli?

Since it is very much a first expression I believe it matters, yes. It's similar to arguing that `static public void main(args: String[])` is of so much of an issue, which is true, but since it's the first thing a newcomer sees it matters.

odersky (approved):
I also recommend accept

romanowski:
@lrytz 

> I have some questions about directives vs command-line options. First, which takes precedence?

Command line options. I've included that in the updated text.

> Some options are also available as directives, others are not - how is that designed?

We tried to add directives for all options for that it make sense (e.g. I do not see how a directive for `--watch` should work) but we may miss something. 

> Would it make sens to have a using runnerOpt [option...] directive? Then the directives could be cut down to the most important ones, like using lib. Some directives like using jsModuleSplitStyleStr seem rather exotic.

I am cannot clearly at this moment see use cases for `runnerOpt` and I fear it may cause confusion when with multiple main being used. We are open for discussion and I think we can follow up in the [Scala CLI](https://github.com/virtuslab/scala-cli) repo (as discussion or an issue).

> On the other hand, there's no directive equivalent for --jvm, is that intended?

That is something we need to add, there is [an issue](https://github.com/VirtusLab/scala-cli/issues/1337) for it.


> [The doc page](https://romanowski.github.io/scala-cli/docs/reference/scala-command/cli-options) doesn't seem to mention -O / --scala-opt? Could that be included at "options forwarding"?
> Would it make sense to also forward -deprecation, -feature, -release to the compiler?

It is included in the latest version of te proposal


> Could scala-cli support -Dkey=value style java properties?

I think we could but it may cause some conflicts. Again, I think we can proceed in Scala CLI repo.

> Does scala-cli support argument files (@file in the compiler)?

Not now and our idea was to use directives for that. But from technical point of view it could be added.

romanowski:
@bjornregnell 

> And the latest commit that I see here is currently from Sept so perhaps your changes are not live yet or am I missing something?

I've included the required specification as a comment, which was not ideal. Now, I've included a compact version of the specification in the text of the proposal.

lrytz:
> > Would it make sens to have a using runnerOpt [option...] directive? Then the directives could be cut down to the most important ones, like using lib. Some directives like using jsModuleSplitStyleStr seem rather exotic.
> 
> I am cannot clearly at this moment see use cases for `runnerOpt` and I fear it may cause confusion when with multiple main being used. We are open for discussion and I think we can follow up in the [Scala CLI](https://github.com/virtuslab/scala-cli) repo (as discussion or an issue).

Maybe I was not clear and you mis-understood? My idea with `runnerOpt` is to pass command-line flags to scala-cli, I didn't mean flags passed to the main method of the program. Then you could reduce `using` directives to the most common ones.

content/scala-cli.md
```diff
@@ -165,3 +171,539 @@ Scala CLI has many alternatives. The most obvious ones are sbt, Mill, or other b
...
+ - `-release`
+ - `-color`
+ - `-nowarn`
+ - `-feature`
+ - `-deprecation`
+
+
+ Additionally, all options that start with:
+- `-g`
+- `-language`
```
bishabosha, line 192: I think `-source` should also be adopted by default, currently you must do e.g. `-O -source:3.2`

bishabosha:
Hello, again not on the committee, however I noticed that in the updated spec, it says the `--ammonite` flag should not be maintained - but ammonite style imports are still supported (at least in 0.1.16). e.g.
```scala
import $ivy.`com.lihaoyi::fansi:0.4.0`

@main def hello = println(fansi.Color.Blue("Hello, World!"))
```
still compiles fine when using `scala-cli` under `scala-cli-sip`/`scala` modes. To me this seems perhaps a bug as it is not mentioned anywhere in the SIP to support this feature.

bjornregnell:
> I've included the required specification as a comment, which was not ideal. Now, I've included a compact version of the specification in the text of the proposal.

Ok great, now I see it in the file tab of this PR. 

I'll try to present the updated version at the SIP committee meeting in 15 min...

sjrd (approved):
Thank you for all the changes, @romanowski. At this point I can recommend it to be Accepted.

julienrf:
The Committee observed that the specification lacked details (e.g. what the “test” command does is not explained) on the one hand, but was maybe too detailed on the optional “MAY” parts. In order to move forward, the Committee voted to move the proposal to the “implementation” stage, but they expect clearer documentation/specification before moving to the “completed” stage. The next step is to provide a way to ship it to the hands of the users so that they can start experimenting with it and provide feedback.

bjornregnell:
This is good news now that scala-cli as the new scala is accepted as an offial "thing" in the experimental stage. See my comment/question here on proposed ways forward: https://github.com/VirtusLab/scala-cli/discussions/266#discussioncomment-3951382

PR #47, opened 2022-08-17
SIP-47 - Clause Interleaving

Sporarum:
We propose to generalize method signatures to allow any number of type parameter lists, interleaved with term parameter lists and using parameter lists. As a simple example, it would allow to define
~~~ scala
def pair[A](a: A)[B](b: B): (A, B) = (a, b)
~~~
Here is also a more complicated and contrived example that highlights all the possible interactions:
~~~ scala
def foo[A][B <: A](using a: A)(b: B)[C <: a.type, D](cd: (C, D))[E]: Foo[A, B, C, D, E]
~~~
(taken from the summary section)

### DIFF
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -72,6 +72,7 @@ class Store:
 This definition provides the expected source API at call site, but it has two issues:
 * It is more complex than expected, forcing a user looking at the API to navigate to the definition of `StoreGetOrElse` to make sense of it.
 * It is inefficient, as an intermediate instance of `StoreGetOrElse` must be created for each call to `getOrElse`.
+* Overloading resolution looks at clauses after the first one, but only in methods, the above is ambiguous with any `def getOrElse(k:Key): ...`, whereas the proposed signature is not ambiguous with for example `def getOrElse(k:Key)[A,B](x: A, y: B)`
 
 Another workaround is to return a polymorphic function, for example:
 ~~~scala
@@ -84,6 +85,7 @@ While again, this provides the expected API at call site, it also has issues:
 * The definition is cumbersome to write, especially if there are a lot of term parameters
 * Methods containing curried type clauses like `def foo[A][B](x: B)` cannot be represented in this way, as polymorphic methods always have to have a term parameter right after.
 * It is inefficient, as many closures must be created for each call to `getOrElse` (one per term clause to the right of the first non-initial type clause).
+* Same problem as above with overloading
 
 ### Statically sized collection
 Let us suppose we have defined a statically sized collection `Vec[N <: Int, +A]`, which contains exactly `N` elements of type `A`. Its API could look like the following:
@@ -165,6 +167,8 @@ Param             ::=  id ‘:’ ParamType [‘=’ Expr]
 ~~~
 The main rules of interest are `DefParamClauses` and `DefParamClause`, which now allow any number of type parameter clauses, term parameter clauses and using parameter clauses, in any order.
 
+Note that these are also used for the right-hand side of extension methods, clause interleaving thus also applies to them.
+
 It is worth pointing out that there can still only be at most one implicit parameter clause, which, if present, must be at the end.
 
 The type system and semantics naturally generalize to these new method signatures.
@@ -179,6 +183,10 @@ class Pair[+A](val a: A)[+B](val b: B)
 Class signatures already have limitations compared to def signatures. For example, they must have at least one term parameter list. There is therefore precedent for limiting their expressiveness compared to def parameter lists.
 
 The rationale for this restriction is that classes also define associated types. It is unclear what the type of an instance of `Pair` with `A` and `B` should be. It could be defined as `Foo[A][B]`. That still leaves holes in terms of path-dependent types, as `B`'s definition could not depend on the path `a`. Allowing interleaved type parameters for class definitions is therefore restricted for now. It could be allowed with a follow-up proposal.
+
+Note: As `apply` is a normal method, it is totally possible to define a method `def apply[A](a: A)[B](b: B)` on `Pair`'s companion object, allowing to create instances with `Pair[Int](4)[Char]('c')`.
+
+The left hand side of extension methods remains unchanged, since they only have one explicit term clause, and since the type parameters are very rarely passed explicitly, it is not as necessary to have multiple type clauses there.
 
 ### Compatibility
 The proposal is expected to be backward source compatible. New signatures currently do not parse, and typing rules are unchanged for existing signatures.
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -73,6 +73,18 @@ This definition provides the expected source API at call site, but it has two is
 * It is more complex than expected, forcing a user looking at the API to navigate to the definition of `StoreGetOrElse` to make sense of it.
 * It is inefficient, as an intermediate instance of `StoreGetOrElse` must be created for each call to `getOrElse`.
 
+Another workaround is to return a polymorphic function, for example:
+~~~scala
+def getOrElse(k:Key): [V >: k.Value] => (default: V) => V = 
+    [V] => (default: V) => ???
+~~~
+While again, this provides the expected API at call site, it also has issues:
+* The behavior is not the same, as `default` has to be a by-value parameter
+* The definition is hard to visually parse, as users are more used to methods (and it is our opinion this should remain so)
+* The definition is cumbersome to write, especially if there are a lot of term parameters
+* Methods containing curried type clauses like `def foo[A][B](x: B)` cannot be represented in this way, as polymorphic methods always have to have a term parameter right after.
+* It is inefficient, as many closures must be created for each call to `getOrElse` (one per term clause to the right of the first non-initial type clause).
+
 ### Statically sized collection
 Let us suppose we have defined a statically sized collection `Vec[N <: Int, +A]`, which contains exactly `N` elements of type `A`. Its API could look like the following:
 ~~~ scala
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -188,6 +188,8 @@ Note: As `apply` is a normal method, it is totally possible to define a method `
 
 The left hand side of extension methods remains unchanged, since they only have one explicit term clause, and since the type parameters are very rarely passed explicitly, it is not as necessary to have multiple type clauses there.
 
+Currently, Scala 2 can only call/override methods with at most one leading type parameter clause, which already forbids calling extension methods like `extension (x: Int) def bar[A](y: A)`, which desugars to `def bar(x: Int)[A](y: A)`. This proposal does not change this, so methods like `def foo[A](x: A)[B]` will not be callable from Scala 2.
+
 ### Compatibility
 The proposal is expected to be backward source compatible. New signatures currently do not parse, and typing rules are unchanged for existing signatures.
 
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -133,9 +133,11 @@ The type system and semantics naturally generalize to these new method signature
 ### Restrictions
 
 #### Type Currying
-Type parameters cannot be curried (having two type clauses next to each other), as this would allow partial type inference, there is a big concern it would become a recommended norm to _always_ curry type parameters.
+Currying type clauses enables partial type inference, as the left clause can be specified while the right one is not.
+As this is a very useful feature, we expect people would use it liberally, and recommending the curried form.
+We are uncertain about the readability of the resulting methods, we have therefore decided to not include type currying as part of this proposal.
 
-Note however that, if absolutely necessary, it is still possible to curry type parameters as such: `def foo[A](using A =:= A)[B]`, since the implicit search for `A =:= A` should always succeed.
+Note however that, if absolutely necessary, it is still possible to curry type parameters as such: `def foo[A](using DummyImplicit)[B]`, since the implicit search for `DummyImplicit` will always succeed.
 This is sufficiently unwieldy that it is unlikely the above becomes the norm.
 
 #### Class Signatures
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -1,0 +1,192 @@
+---
+layout: sip
+title: SIP-NN Clause Interleaving
+stage: design
+permalink: /sips/:title.html
+---
+
+**By: Quentin Bernet and Guillaume Martres and Sébastien Doeraene**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| May 5th 2022  | Initial Draft      |
+| Aug 17th 2022 | Formatting         |
+
+## Summary
+
+We propose to generalize method signatures to allow any number of type parameter lists, interleaved with term parameter lists and using parameter lists. As a simple example, it would allow to define
+~~~ scala
+def pair[A](a: A)[B](b: B): (A, B) = (a, b)
+~~~
+Here is also a more complicated and contrived example that highlights all the possible interactions:
+~~~ scala
+def foo[A][B <: A](using a: A)(b: B)[C <: a.type, D](cd: (C, D))[E]: Foo[A, B, C, D, E]
+~~~
+
+
+## Motivation
+
+We motivate the feature with two use cases:
+
+* a `getOrElse` method for a heterogeneous key-value store, which is an occurrence of wanting a type parameter whose bounds are path-dependent on a term parameter, and
+
+* a `fill` method for a statically sized collection, where we want to explicitly pass one type parameter, but leave the other one to be inferred.
+
+### Heterogeneous key-value store
+Consider an API for a heterogenous key-value store, where keys know what type of value they must be associated to:
+~~~ scala
+trait Key:
+  type Value
+
+class Store:
+  def get(key: Key): key.Value = …
+  def put(key: Key)(value: => key.Value): Unit = …
+~~~
+We want to provide a method `getOrElse`, taking a default value to be used if the key is not present. Such a method could look like
+~~~ scala
+def getOrElse(key: Key)(default: => key.Value): key.Value = …
+~~~
+However, at call site, it would prevent from using as default value a value that is not a valid `key.Value`. This is a limitation compared to other `getOrElse`-style methods such as that of `Option`, which allow passing any supertype of the element type.
+
+In current Scala, there is no way to define `Store.getOrElse` in a way that supports this use case. We may try to define it as
+~~~ scala
+def getOrElse[V >: key.Value](key: Key)(default: => V): V = …
+~~~
+but that is not valid because the declaration of `V` needs to refer to the path-dependent type `key.Value`, which is defined in a later parameter list.
+
+We might also try to move the type parameter list after `key` to avoid that problem, as
+~~~ scala
+def getOrElse(key: Key)[V >: key.Value](default: => V): V = …
+~~~
+but that is also invalid because type parameter lists must always come first.
+
+A workaround is to return an intermediate object with an `apply` method, as follows:
+~~~ scala
+class Store:
+  final class StoreGetOrElse[K <: Key](val key: K):
+    def apply[V >: key.Value](default: => V): V = …
+  def getOrElse(key: Key): StoreGetOrElse[key.type] = StoreGetOrElse(key)
+~~~
+This definition provides the expected source API at call site, but it has two issues:
+* It is more complex than expected, forcing a user looking at the API to navigate to the definition of `StoreGetOrElse` to make sense of it.
+* It is inefficient, as an intermediate instance of `StoreGetOrElse` must be created for each call to `getOrElse`.
+
+### Statically sized collection
+Let us suppose we have defined a statically sized collection `Vec[N <: Int, +A]`, which contains exactly `N` elements of type `A`. Its API could look like the following:
+~~~ scala
+final class Vec[N <: Int, +A] private (...):
+  def dot(that: Vec[N, A])(using Numeric[A]): Vec[N, A] = …
+~~~
+We would like to define a `fill` method for its companion object, as follows:
+~~~ scala
+object Vec:
+  def fill[N <: Int, A](value: A)(using ValueOf[N]): Vec[N, A] = …
+~~~
+The implicit `ValueOf[N]` parameter would allow the implementation of `fill` to get hold of the run-time representation of `N` as an `Int` value, which is required to actually create a `Vec` of the right length.
+
+We would like to use `fill` as follows:
+~~~ scala
+val v = Vec.fill[3]("hello")
+~~~
+We want to explicitly specify the `N` parameter as `3`, since there is no way for the compiler to infer it for us. However, we would like it to infer `A` as `String`. This is unfortunately not possible in current Scala, as either all or none of the type parameters for any given method must be provided.
+
+The typical workaround would be to explicitly provide both type parameters:
+~~~ scala
+val v = Vec.fill[3, String]("hello")
+~~~
+If we feel strongly about inferring `A`, we can use a similar indirection with an intermediate class, as follows:
+~~~ scala
+object Vec:
+  final class Fill[N <: Int]()(using ValueOf[N]):
+    def apply[A](value: A): Vec[N, A] = …
+  def fill[N <: Int](using ValueOf[N]): Fill[N] = Fill[N]()
+~~~
+which enables the desired call site API, with the same drawbacks that we already saw with the heterogeneous key-value store.
+
+## Proposed solution
+### High-level overview
+
+To solve the above problems, we propose to generalize method signatures so that they can have multiple type parameter lists, interleaved with term parameter lists and using parameter lists. There can be multiple consecutive type parameter lists.
+
+For the heterogeneous key-value store example, this allows to define `getOrElse` as follows:
+~~~ scala
+def getOrElse(key: Key)[V >: key.Value](default: => V): V = …
+~~~
+It provides the best of all worlds:
+* A convenient API at call site
+* A single point of documentation
+* Efficiency, since the method erases to a single JVM method with signature `getOrElse(Object,Object)Object`
+
+The `Vec.fill` method can be defined as follows:
+~~~ scala
+def fill[N <: Int][A](value: A)(using ValueOf[N]): Vec[N, A] = …
+~~~
+Since `N` and `A` are in two separate type parameter lists, it is possible to call it as
+~~~ scala
+val v = Vec.fill[3][String]("hello")
+~~~
+or, letting the second type parameter list be inferred by the compiler, as
+~~~ scala
+val c = Vec.fill[3]("hello")
+~~~
+
+### Specification
+We amend the syntax of def parameter clauses as follows:
+~~~
+DefDcl            ::=  DefSig ‘:’ Type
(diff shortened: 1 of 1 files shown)


kitbellew:
> We propose to generalize method signatures to allow any number of type parameter lists, interleaved with term parameter lists and using parameter lists. As a simple example, it would allow to define
> 
> ```scala
> def pair[A](a: A)[B](b: B): (A, B) = (a, b)
> ```
> 
> Here is also a more complicated and contrived example that highlights all the possible interactions:
> 
> ```scala
> def foo[A][B <: A](using a: A)(b: B)[C <: a.type, D](cd: (C, D))[E]: Foo[A, B, C, D, E]
> ```
> 
> (taken from the summary section)

what about
```scala
def pair[A1, A2](a1: A1, a2: A2)[B1, B2, B3](b1: B1, b2: B2, b3: B3): (A1, B3) = (a1, b3)
```

Sporarum:
Yes of course !
If you haven't already, do read the "changed file" (even though it's really an added file), it should give you a full picture of what this feature allows

And if anything's not clear, feel free to ask
(if possible as a PR comment, so I can reply more easily)

Kordyjan:
While I agree that what is proposed in this SIP would be a valuable addition to Scala. I remember myself having to write intermediate classes with `apply` methods to circumvent the single type param clause limitation. 

However, right now in Scala 3 there are some ways to achieve that. For example, if I'm not mistaken:
``` scala
def getOrElse(key: Key)[V >: key.Value](default: => V): V = ???
```
from the proposal, can be right now defined as 
```scala
def getOrElse: (k:Key) => [V >: k.Value] => (default: V) => V = 
    (k: Key) => [V] => (default: V) => ???
```

Of course, the current approach is much less visually pleasing and has some (quite arbitrary) limitations. For example, it is  not legal to declare
```scala
def makePair: [A] => [B] => (A, B) = ???
```
while I assume your proposal would allow
```scala
def makePair[A][B]: (A, B) = ???
```

I think the proposal itself should address that. It should make a statement about why we should add a new thing to the language that will have only very niche use cases, instead of improving already existing mechanisms.

Sporarum:
@Kordyjan, thank you for your feedback !
I added a section on that, let me know what you think

Sporarum:
@Kordyjan I added a section addressing what you asked, what do you think ?

Kordyjan:
Sorry for not answering earlier. The section indeed addresses what I thought was missing in the proposal.

About the proposal itself: I think it is well thought-out. It feels much more like lifting an arbitrary limitation than adding a new feature to the language. As I said, it would have made much code that I wrote in the past clearer and easier to follow. My only fear was that it would make it possible to write unnecessarily complicated signatures for a small benefit, making libraries relying heavily on this feature scary for the newcomers. However, I haven't found anything worse than what the current language features allow library authors to create. 

Therefore I'm provisionally **voting to accept this proposal**. However, I'm curious about the opinions of other reviewers.

Sporarum:
@soronpo, @chrisandrews-ms, I have not heard from you, what do you think of the proposal ?

soronpo:
Apologies, I remembered I was not assigned to it at the beginning, and didn't notice I was assigned after. 
For motivation, you should mention that it resolves an extension method feature limitation (need to find the relevant issue).
I'm **in favor of accepting** this proposal, but I should mention that to accept it after the implementation stage is a much greater challenge. The feature interaction and possible combinations here are massive! To be able to properly test all that and report good error messages may be very difficult.

Sporarum:
> extension method feature limitation

Do you mean `extension [T](...) def foo[U](...)` ?
If so this has already been added, and at the time, the codebase was refactored to deal with arbitrarily many type clauses, so your point:
> accept it after the implementation stage is a much greater challenge

Might not actually be that challenging, as the implementation boils down to "modify the parser", since the generated tree is already valid in the rest of the compiler

Sporarum:
I will add two things to the proposal later:
The fact that
```scala
def getOrElse(k:Key): [V >: k.Value] => (default: V) => V = 
    (k: Key) => [V] => (default: V) => ???
```
Poses problem with overloading resolution, if there is another `def getOrElse(k:Key)...`

Add a section clarifying that Clause Interleaving also applies to the right-hand side of extension methods

soronpo:
> Do you mean `extension [T](...) def foo[U](...)` ?
> If so this has already been added, and at the time, the codebase was refactored to deal with arbitrarily many type clauses

I remembered https://github.com/lampepfl/dotty/issues/14451 but now I see it will not be resolved.

> Might not actually be that challenging, as the implementation boils down to "modify the parser", since the generated tree is already valid in the rest of the compiler

My guess is that type inference and overloading checks are not that generic. Hoping to be proven wrong.

chrisandrews-ms:
I'm in **favor of accepting** too. I can see the utility of this proposal, especially for partial inference of type parameters, a problem that has come up a few times in our Scala 2 codebase. The SIP seems well thought out and uncontroversial.

A couple minor questions:

1. Am I right in assuming that methods with interleaved clauses _wouldn't_ be callable from Scala 2.13 (using the TASTy reader)? I don't see that as a problem (since it also applies to other Scala 3 features) but it would be good to document that limitation.

1. I note that there's already an experimental implementation PR. Is that ready to merge if this SIP passes the first vote? It would be good to be able to experiment and look for edge cases.

Sporarum:
>     1. Am I right in assuming that methods with interleaved clauses _wouldn't_ be callable from Scala 2.13 (using the TASTy reader)? I don't see that as a problem (since it also applies to other Scala 3 features) but it would be good to document that limitation.

That is correct, I'll update the proposal

>     2. I note that there's already an experimental implementation PR. Is that ready to merge if this SIP passes the first vote? It would be good to be able to experiment and look for edge cases.
The implementation is pretty much ready, yes, but it adds it under `experimental`, so it will only be available in nightly and with an import, which would give us some time to experiment

raulraja:
@Sporarum Thanks for the proposal. The feature looks great and is in the spirit of scala's existing support for multiple argument groups. I'm also in **favor of accepting**.

Do you know what integrations or code changes are required in Metals, IDEA, scalafmt, and other tools from the ecosystem to support this feature?

gabro:
> Do you know what integrations or code changes are required in Metals, IDEA, scalafmt, and other tools from the ecosystem to support this feature?

@Sporarum knows best, but I would guess:

- Scalameta parser (which affects Metals and Scalafmt)
- IJ parser
- code highlighting tools / editor grammars
  - https://github.com/scala/vscode-scala-syntax
  - and possibly
	- linguist
	- highlight.js
    - treesitter

Sporarum:
> Do you know what integrations or code changes are required in Metals, IDEA, scalafmt, and other tools from the ecosystem to support this feature?

The changes to parsing are minimal, so the real question is how much do we have to change in the back end ?
Since extension methods already desugar to methods like `def foo[A](x: A)[B](y: B)`, some infrastructure has to already exist, but it might not be general enough for clause interleaving
The amount of work is difficult for me to estimate, as I have not contributed to most of these, so I do not know how adapted that infrastructure is.

However, I have started working on [scalameta](https://github.com/scalameta/scalameta/pull/2820); Metals and Scalafmt will follow, as for the other tools, I have not started looking into it yet (and am not sure whether I will have the time to work on them)

soronpo:
> (and am not sure whether I will have the time to work on them)

Nobody is requesting you to do so, at all.

gabro:
> > (and am not sure whether I will have the time to work on them)
> 
> 
> 
> Nobody is requesting you to do so, at all. 

Yep, just to clarify, I'm purely stating the impact of the change. It's already great that you're working on the Scalameta parser!

Sporarum:
> > (and am not sure whether I will have the time to work on them)
> 
> Nobody is requesting you to do so, at all.

While that is very kind, it is incorrect, as it has been requested of me as part of my internship !

soronpo:
> While that is very kind, it is incorrect, as it has been requested of me as part of my internship !

OK, I meant not in the scope of submitting a SIP proposal 😄

Kordyjan:
@soronpo @chrisandrews-ms I'll be presenting this proposal at today's Committee meeting.

content/clause-interleaving.md
```diff
@@ -1,8 +1,9 @@
 ---
 layout: sip
-title: SIP-NN Clause Interleaving
-stage: design
-permalink: /sips/:title.html
+title: SIP-47 Clause Interleaving
+stage: implementation
+status: waiting-for-implementation
```
Kordyjan, line 5: Shouldn't we change that AFTER the vote? 😅

content/clause-interleaving.md
```diff
@@ -1,8 +1,9 @@
 ---
 layout: sip
-title: SIP-NN Clause Interleaving
-stage: design
-permalink: /sips/:title.html
+title: SIP-47 Clause Interleaving
```
kitbellew, line 3: I'd like to mention once again that it should be "Parameter Clause Interleaving" as we're not doing that with if-clauses here.

soronpo:
@Sporarum Thank you for the work so far. The committee decided to return this proposal for a revision. 
We are inclined to accept, but request that there will be an added restriction:
There cannot be two (or more) adjacent type parameter clauses. 
The committee believes that Named Typed Parameter Application (a future SIP) is the better approach to answer use-cases where adjacent type parameter clauses are useful.

content/clause-interleaving.md
Sporarum: I appreciate your concern, however this has not caused confusion for anyone else, and there is precedent for omitting "parameter" from the name:
https://docs.scala-lang.org/scala3/reference/contextual/using-clauses.html
I will therefore not change the name for now, but I encourage affected others to comment on this potential problem

morgen-peschke:
> The committee believes that Named Typed Parameter Application (a future SIP) is the better approach to answer use-cases where adjacent type parameter clauses are useful.

That's disappointing, having to remember what name a particular method chose for it's un-inferred parameter is going to be really annoying.

content/clause-interleaving.md
kitbellew: > I appreciate your concern, however this has not caused confusion for anyone else, and there is precedent for omitting "parameter" from the name: https://docs.scala-lang.org/scala3/reference/contextual/using-clauses.html I will therefore not change the name for now, but I encourage affected others to comment on this potential problem

it's not about omitting the word "parameter", it's about naming the type of clause. the word "using" names the clause in your other example.

soronpo:
> having to remember what name a particular method chose for it's un-inferred parameter is going to be really annoying

That's where tooling comes in, and if your API warrants a named application you better give it a proper name instead of just a letter.

content/clause-interleaving.md
anatoliykmetyuk: /cc @julienrf, WDYT?

morgen-peschke:
> > having to remember what name a particular method chose for it's un-inferred parameter is going to be really annoying
> 
> That's where tooling comes in, and if your API warrants a named application you better give it a proper name instead of just a letter.

That's actually worse. 

Something like `opt.liftTo[F = EitherNel[String,*]](error)` vs `opt.liftTo[FE = EitherNel[String,*]](error)` is annoying to remember, but something like `opt.liftTo[EffectError = EitherNel[String,*]](error)` is both annoying to remember and tedious to type.

Meanwhile, because the kind of the arguments are different, it's very difficult to screw up something like `opt.liftTo[EitherNel[String,*]](error)`.

Sporarum:
To address these concerns, maybe the best is to split this PR into two: strictly interleaved clauses, and type currying
The goal would be:
Treat strictly interleaved clauses as the new version of this SIP, so things can keep moving forward, as this part is uncontroversial
Have named type parameters and type currying in experimental at the same time, to gather feedback and see what people do with them

To this end, I will edit this SIP to make it into strictly interleaved clauses (but will keep the current name), and if requested, I'll make a separate SIP for type currying

lrytz:
I was very surprised to see how small this implementation is (https://github.com/lampepfl/dotty/pull/14019), which is of course a good thing! Some of the changes I expected to see are actually already in the compiler, e.g. https://github.com/lampepfl/dotty/pull/10940/commits/61185949b13969296b9e4d2f3f91427c8914e297.

Sporarum:
Yes !
I suspect most bugs I will find with clause interleaving would have already been reachable through extension methods, similarly to lampepfl/dotty#14451

The only non reachable bugs would be the ones related to:
* type currying, which is dropped from this proposal anyways
* exclusively 3 or more type clauses, but that would be strange

content/clause-interleaving.md
julienrf: It makes sense to change before the vote: the Committee votes for moving the proposal to the “implementation” stage.

Sporarum:
> @Sporarum Thank you for the work so far. The committee decided to return this proposal for a revision. We are inclined to accept, but request that there will be an added restriction: There cannot be two (or more) adjacent type parameter clauses. The committee believes that Named Typed Parameter Application (a future SIP) is the better approach to answer use-cases where adjacent type parameter clauses are useful.

What was the committee's opinion on `foo[T](using G)[U]...` ?
This poses problems, as:
* If we allow it, people can call it with `foo[A][B]`, which goes against the spirit of that restriction
* If we disallow it, then there are two different (yet very similar) rules around polymorphism: 
* * Polymorphic methods must have an explicit term parameter between types
* * Polymorphic functions must have a term parameter, but it doesn't have to be explicit

sjrd:
We did not discuss that particular scenario. My opinion is that we can allow it, especially if it makes things easier. It *could* be abused by library designers who really really want their curried type params, but the definition would be obscured enough that it won't be viable as a "recommendation".

Sporarum:
@Kordyjan, @soronpo, @chrisandrews-ms
You are free to review the updated proposal

As I was not present at the SIP meeting, please advise me if I did not transcribe accurately the perceived downsides of type currying

content/clause-interleaving.md
```diff
@@ -134,49 +101,44 @@ It provides the best of all worlds:
...
 The type system and semantics naturally generalize to these new method signatures.
 
-When a method poly type consists of _n_ consecutive type parameter lists, and _m < n_ lists are specified at call site, they always correspond to the left-most _m_ type parameter lists of the signature. The right-most _(n-m)_ type parameter lists are then inferred.
-
 ### Restrictions
+
+#### Type Currying
+Type parameters cannot be curried (having two type clauses next to each other), as this would allow partial type inference, there is a big concern it would become a recommended norm to _always_ curry type parameters.
+
+Note however that, if absolutely necessary, it is still possible to curry type parameters as such: `def foo[A](using A =:= A)[B]`, since the implicit search for `A =:= A` should always succeed.
```
soronpo, line 138: You can just use `DummyImplicit` instead of `A =:= A`.

Sporarum: I didn't know about that, thank you !
```suggestion
Note however that, if absolutely necessary, it is still possible to curry type parameters as such: `def foo[A](using DummyImplicit)[B]`, since the implicit search for `DummyImplicit` will always succeed.
```

bjornregnell: I guess the worry is that over-usage of type currying would make the code unnecessarily cryptic and more difficult to read for humans. If I'm right about that then perhaps include something similar in the text - "always" is a bold word and the universe will outlive the Scala programming language, presumably :)

morgen-peschke: Considering the odd constructs that have come out of `fewerBraces`, this seems like a weird hill to die on. `def foo[A][B](...)` is way less cryptic than `def foo[A]: FooPartiallyApplied[A] ...`

bjornregnell: Well, I guess the point is that it is viewed as less cryptic than `def foo[A, B](...)`
(Anyways, I was just trying to help sharpening the text and describe in more detail the voiced concerns.)

Sporarum: @morgen-peschke I appreciate your point of view
While type currying is not included in this SIP, this doesn't mean it will be banned from ever becoming a part of the language!

There are three mechanisms for partial type inference that I could come up with so far:
* type currying
* named type parameters
* allowing `?` in type application: `Vec.fill[4, ?]("Hello")` (`?` would get replaced by `String`)

I believe all of them should be carefully considered, to try to find the best way forward
(In particular, the best way forward might be to allow more than one)

To this end, a new SIP for type currying will need to be written, would you be interested in being the one to do it ?
(Of course you can take as much as you like from this SIP, in particular before commit [a492268](https://github.com/scala/improvement-proposals/pull/47/commits/a4922683dd61904aa2b0ac06fc8b8ff2ab766c20))

If I do write it instead, you will of course be welcome to comment, to try and make the best argument possible for type currying

Sporarum: > I guess the worry is that over-usage of type currying would make the code unnecessarily cryptic and more difficult to read for humans. If I'm right about that then perhaps include something similar in the text - "always" is a bold word and the universe will outlive the Scala programming language, presumably :)

Thank you, I will incorporate changes to make it more clear and nuanced

> the universe will outlive the Scala programming language

That's true ^^
However, the "always" in the text is implicitly constrained by the usage of Scala (as type currying can only happen at definition), so those two statements can coexist without contradiction !

bjornregnell: > those two statements can coexist without contradiction

Lol. Yes different types of currying can co-exist in this ambiguous real world. (I had some extra garam masala in my spicy stew yesterday, for a change and I gave my cooking :100:  )

content/clause-interleaving.md
```diff
@@ -0,0 +1,178 @@
...
+Note that these are also used for the right-hand side of extension methods, clause interleaving thus also applies to them.
+
+It is worth pointing out that there can still only be at most one implicit parameter clause, which, if present, must be at the end.
+
+The type system and semantics naturally generalize to these new method signatures.
+
+### Restrictions
+
+#### Type Currying
+Type parameters cannot be curried (having two type clauses next to each other), as this would allow partial type inference, there is a big concern it would become a recommended norm to _always_ curry type parameters.
```
Sporarum, line 136: > I guess the worry is that over-usage of type currying would make the code unnecessarily cryptic and more difficult to read for humans.
```suggestion
Currying type clauses enables partial type inference, as the left clause can be specified while the right one is not.
As this is a very useful feature, we expect people would use it liberally, and recommending the curried form.
We are uncertain about the readability of the resulting methods, we have therefore decided to not include type currying as part of this proposal.
```

Sporarum: @bjornregnell What do you think of the above ?

bjornregnell: Looks very Good To Me!

content/clause-interleaving.md
morgen-peschke: @Sporarum 
> To this end, a new SIP for type currying will need to be written, would you be interested in being the one to do it ?

I'd be interested, but due to offline stuff, I wouldn't be able to pick it up until around December.

Sporarum: @morgen-peschke 
Okay, if it needs to be done before, I'll do it instead, otherwise feel free to contact me when you have the time

julienrf:
Please @Kordyjan, @soronpo, and @chrisandrews-ms let us know if the author addressed your feedback, and confirm that you agree the Committee should vote on this proposal during the next meeting to accept it, or tell us if that changed.

Kordyjan:
I maintain my recommendation that the Committee should **accept** the proposal.

chrisandrews-ms:
I also recommend to accept

raulraja:
> I maintain my recommendation that the Committee should vote during the next meeting and accept the proposal.

Also, **recommend accepting**

adampauls:
> The committee believes that Named Typed Parameter Application (a future SIP) is the better approach to answer use-cases where adjacent type parameter clauses are useful. 

Out of curiosity, why does the committee believe these two answers should be mutually exclusive? They are both available for non-type parameters.

soronpo:
> > The committee believes that Named Typed Parameter Application (a future SIP) is the better approach to answer use-cases where adjacent type parameter clauses are useful.
> 
> Out of curiosity, why does the committee believe these two answers should be mutually exclusive? They are both available for non-type parameters.

To prevent API entropy and abuse by "creative" developers. In short, it was easier to get approval for the strict version than the general version. It's always possible to submit the following SIP in the future, if indeed there are good use-cases. Small well-calculated and steady increments are good for languages.

odersky (approved):
I also recommend accept.

Kordyjan:
The proposal was accepted during our last SIP meeting!

This means that the Committee agrees that the proposal can be implemented as an experimental feature.

Sporarum:
Thank you, I'm happy to hear that !
The PR for the feature (already in experimental) can be found here:
https://github.com/lampepfl/dotty/pull/14019
I believe it is ready or almost ready, and therefore requires only review from the compiler team
(And I will address the reviews personally)

content/clause-interleaving.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Specification
+We amend the syntax of def parameter clauses as follows:
+
+~~~
+DefDcl                 ::=  DefSig ‘:’ Type
+DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
+DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
+DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
+DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
```
kitbellew, line 112: does this mean that this is not valid anymore:
```
  def foo[A](implicit a: A) = ...
```

Sporarum: Oh, nicely spotted, my intent was for it to be allowed,
but it appears I made a mistake with the grammar

Since the implementation team (which is also myself) can make changes to the proposal, I believe this can fall under this umbrella
I will think about what's the clearest way to fix the grammar

Sporarum: This should fix it:
```diff
  DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
- DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
+ DefSig                 ::=  id [DefParamClauses] [DefTypeParamClause] [DefImplicitClause]
  DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
  DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
```
There is another quirk of this definition however, it allows to parse `def f(a: Int)(b: Int)` to be parsed in two different ways (one vs two `DefParamClauseChunk`), is this an issue ?

julienrf: I didn’t have a look at the complete grammar but what you propose does not look correct, what about the following?

~~~ text
DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
DefSig                 ::=  id [DefParamClauses]
DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
DefParamClauseChunk    ::=  DefTypeParamClause | TermOrUsingParamClause
~~~

kitbellew: Alas, this allows two consecutive `type` clauses, or multiple `implicit` clauses.

here's another option:
```
DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
DefParamClauses        ::=  DefParamClauseChunkTypeOpt {DefParamClauseChunkTypeReq}
DefParamClauseChunkTypeOpt    ::=  [DefTypeParamClause] {TermOrUsingParamClause}
DefParamClauseChunkTypeReq    ::=  DefTypeParamClause {TermOrUsingParamClause}
```

content/clause-interleaving.md
```diff
@@ -0,0 +1,180 @@
...
+TermOrUsingParamClause ::=  DefTermParamClause
+                         |  UsingParamClause
+DefTypeParamClause     ::=  [nl] ‘[’ DefTypeParam {‘,’ DefTypeParam} ‘]’
+DefTypeParam           ::=  {Annotation} id [HkTypeParamClause] TypeParamBounds
+DefTermParamClause     ::=  [nl] ‘(’ [DefTermParams] ‘)’
+UsingParamClause       ::=  [nl] ‘(’ ‘using’ (DefTermParams | FunArgTypes) ‘)’
+DefImplicitClause      ::=  [nl] ‘(’ ‘implicit’ DefTermParams ‘)’
```
kitbellew, line 119: I'd also rename
```
TermOrUsingParamClause ::=  DefTermParamClause
                         |  UsingParamClause
UsingParamClause       ::=  [nl] ‘(’ ‘using’ (DefTermParams | FunArgTypes) ‘)’
DefImplicitClause      ::=  [nl] ‘(’ ‘implicit’ DefTermParams ‘)’
```
to
```
DefNonTypeParamClause  ::=  DefTermParamClause |  DefUsingParamClause
DefUsingParamClause    ::=  [nl] ‘(’ ‘using’ (DefTermParams | FunArgTypes) ‘)’
DefImplicitParamClause ::=  [nl] ‘(’ ‘implicit’ DefTermParams ‘)’
```

content/clause-interleaving.md
kitbellew: ah, that doesn't work either... perhaps this:
```
DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
DefParamClauses        ::=  [DefTypeParamClause] {DefParamClauseChunk} [TermOrUsingParamClause]
DefParamClauseChunk    ::=  TermOrUsingParamClause {TermOrUsingParamClause} DefTypeParamClause
```

Sporarum: > This should fix it:
> 
> ```diff
>   DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
> - DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
> + DefSig                 ::=  id [DefParamClauses] [DefTypeParamClause] [DefImplicitClause]
>   DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
>   DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
> ```
> 
> There is another quirk of this definition however, it allows to parse `def f(a: Int)(b: Int)` in two different ways (one vs two `DefParamClauseChunk`), is this an issue ?

> I didn’t have a look at the complete grammar but what you propose does not look correct

Could you expand on this, I can't manage to find a failing example with the above syntax

Sporarum: I talked about it some more with @julienrf and we came to a very easy and clean solution:
Use the pre-"no type currying" syntax, and add an end of line comment that says "and type clauses cannot be adjacent"

Sporarum:
```diff
- DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
+ DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]    -- and type clauses cannot be adjacent
- DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
- DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
- TermOrUsingParamClause ::=  DefTermParamClause
-                          |  UsingParamClause
+ DefParamClauses        ::= DefParamClause { DefParamClause }
+ DefParamClause         ::= DefTypeParamClause
+                          | DefTermParamClause
+                          |  UsingParamClause
```
I think this syntax is easier to understand than the formal one

@julienrf How should I go about updating the proposal ?

julienrf: Please open a PR in this repository with your changes.

PR #49, opened 2022-09-23
SIP-49 - Polymorphic Eta-Expansion

Sporarum:
We propose to extend eta-expansion to polymorphic methods. 
This means automatically transforming polymorphic methods into corresponding polymorphic functions, for example:

~~~ scala
def f1[A](x: A): A = ???
val v1_1: [B] => B => B = f1 // f1 becomes [B'] => (y: B') => f1[B'](y)
val v1_2                = f1 // f1 becomes [A'] => (x': A') => f1[A'](x')
~~~

(taken from Summary)

### DIFF
--- a/content/polymorphic-eta-expansion.md
+++ b/content/polymorphic-eta-expansion.md
@@ -1,0 +1,472 @@
+---
+layout: sip
+title: SIP-NN Polymorphic Eta Expansion
+stage: implementation
+permalink: /sips/:title.html
+---
+
+**By: Quentin Bernet and Guillaume Martres** <!-- and Martin and Seb ?-->
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Sep 23th 2022 | Initial Draft      |
+
+## Summary
+
+<!-- A summary of the proposed changes. This should be no longer than 3 paragraphs. It is intended to serve in two ways:
+
+- For a first-time reader, a high-level overview of what they should expect to see in the proposal.
+- For returning readers, a quick reminder of what the proposal is about. -->
+
+We propose to extend eta-expansion to polymorphic methods. 
+This means automatically transforming polymorphic methods into corresponding polymorphic functions, for example:
+
+~~~ scala
+def f1[A](x: A): A = ???
+val v1_1: [B] => B => B = f1 // f1 becomes [B'] => (y: B') => f1[B'](y)
+val v1_2                = f1 // f1 becomes [A'] => (x': A') => f1[A'](x')
+~~~
+
+Returning readers, for a quick glance at a wide array of examples illustrated like the above, go to [High-level overview](#high-level-overview).
+
+In the following, "Note" never introduces new concepts, it points out a non-obvious consequence, and/or reminds the reader of a pertinent fact about Scala.
+
+## Motivation
+
+<!-- A high-level overview of the proposal with:
+
+- An explanation of the problems or limitations that it aims to solve,
+- A presentation of one or more use cases as running examples, with code showing how they would be addressed *using the status quo* (without the feature), and why that is not good enough.
+
+This section should clearly express the scope of the proposal. It should make it clear what are the goals of the proposal, and what is out of the scope of the proposal. -->
+
+Regular eta-expansion is so ubiquitous that most users are not aware of it, for them it is intuitive and obvious that methods can be passed where functions are expected.
+
+When manipulating polymorphic methods, we wager that most users find it confusing not to be able to do the same. 
+This is the main motivation of this proposal.
+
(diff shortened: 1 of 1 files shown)


julienrf:
Thank you for your submision @Sporarum, I’ve assigned a team of reviewers to the proposal.

soronpo:
Yes, and I'm inclined to support its acceptance, but will need to re-read it over the weekend.

lrytz:
I support this proposal. I think it would be good if source/binary/tasty compatibility concerns could be better explained and illustrated.

Sporarum:
@soronpo Did you have a chance to re-read the proposal ?

@dragos I would be curious to have your opinion on this SIP

@lrytz I reworked the compatibility section with help from @sjrd, is the new one clearer ?

lrytz:
> @lrytz I reworked the compatibility section with help from @sjrd, is the new one clearer ?

Yes, this is much more clear now, thank you!

content/polymorphic-eta-expansion.md
```diff
@@ -0,0 +1,495 @@
...
+extension [A](x: A)
+  def foo[B](y: B) = (x, y)
+
+val voo: [T] => T => [U] => U => (T, U) = foo
+// foo expands to: 
+// [T'] => (t: T') => ( foo[T'](t) with expected type [U] => U => (T', U) )
+// [T'] => (t: T') => [U'] => (u: U') => foo[T'](t)[U'](u)
+~~~
+
+#### Without expected type
```
odersky, line 330: To be safe I would strongly suggest that there should be no eta expansion without an expected type. The reason is that missing type parameters are always inferred and we do not want to make an exception now. Type arguments are in that sense analogous to implicits.

Here's an example:
```scala
def f(x: Int)(using C)
```
Here, `f` expands to
```scala
(x: Int) => f(x)(summon[C])
```
It does _not_ expand to
```scala 
(x: Int) => (c: C) ?=> m(x)(using c)
```
Analogously,  when faced with
```scala
class C { type T }
def g(x: C)[U >: x.T]
```
`g` should expand to 
```scala
(x: C) => g(x)[x.T]
```
and not to
```scala
(x: C) => [U] => g(x)[U]
```
Granted, the second version in the polymorphic eta expansion might seem more useful, (and maybe the second version for the implicit eta expansion is deemed more useful, too). But that's not how things work. I believe the analogy between polymorphic and implicit eta expansion is too strong to throw overboard.

Generally, I have a bit mixed feelings about this but am overall positive. The downside is that it is an "exotic" feature with not a lot of use cases, and the worked out definition is very complicated. This will have a significant impact on spec complexity. Do we really need it? On the positive side, the details are worked out very well. If we can restrict eta expansion to kick in only with a polymorphic expected type, I think it's fine. Otherwise the feature interactions with general inference principles of Scala are serious enough to hold back. 

Also note that if we would decide at some point that we need eta expansion without an expected type because we have lots of convincing use cases, we can still add it later.

odersky (changes requested):

content/polymorphic-eta-expansion.md
Sporarum: I ammended the proposal as requested, I will leave this conversation as unresolved so the reasons are apparent

odersky (approved):
With the changes I think this is now good to go in.

soronpo:
I accept the amendment.
I think the proposal should also mention how this feature interacts with the interleaving clauses feature #47.

julienrf:
Hey @Sporarum, the proposal was discussed during the SIP meeting today. The Committee voted to accept the proposal, but they would like to see an example with a method that has interleaved parameter clauses in [this section](https://github.com/scala/improvement-proposals/pull/49/files#diff-4a7cbc7483385994a2174d36a8619c617053905e3ddf2c1d078c832e42834fc8R215-R223) in addition to the extension method.

julienrf:
I went ahead and added the example myself to unblock the situation. Please adjust it if you feel it needs to be adjusted.

julienrf:
I have created an issue on the IntelliJ issue tracker to follow its progress: https://youtrack.jetbrains.com/issue/SCL-20759/Support-polymorphic-eta-expansion-SIP-49

Are there other tools that may be impacted by this SIP?

Sporarum:
@julienrf thank you for taking care of this while I was away

I do not believe any adjustments need to be made

The IntelliJ issue is incorrect however, the eta-expansion of `v1_2` should not be changed by this SIP
This was a later change following this feedback:
> To be safe I would strongly suggest that there should be no eta expansion without an expected type.

julienrf:
OK, thank you for proofreading, I’ve fixed the issue description on the IntelliJ tracker.

PR #50, opened 2022-10-20
SIP-50 - Struct Classes

julienrf:
Based on the [pre-SIP discussion](https://contributors.scala-lang.org/t/pre-sip-structural-data-structures-that-can-evolve-in-a-binary-compatible-way/5684).

### DIFF
--- a/content/struct-classes.md
+++ b/content/struct-classes.md
@@ -16,7 +16,7 @@ title: SIP-50 - Struct Classes
 
 ## Summary
 
-Introduce “struct classes”, a new flavour of class definitions that is a middle
+We introduce “struct classes”, a new flavour of class definitions that is a middle
 ground between regular classes and case classes. Like case classes, struct classes
 structurally implement `equals`, `hashCode`, and `toString`. However, unlike with
 case classes, the companions of struct classes don't have `apply` and `unapply`
@@ -135,7 +135,7 @@ With this definition of `User`, our example program would look as follows:
 
 ~~~ scala
 val alice = User("Alice", 42)
-val updatedAlice = alice.withAge(age = alice.age + 1)
+val updatedAlice = alice.withAge(alice.age + 1)
 assert(updatedAlice == User("Alice", 43))
 updatedAlice match
   case User(_, age) => println(s"Alice is $age years old")
@@ -238,7 +238,7 @@ This definition can be used by our example program as follows:
 
 ~~~ scala
 val alice = User("Alice", 42)
-val updatedAlice = alice.withAge(age = alice.age + 1)
+val updatedAlice = alice.withAge(alice.age + 1)
 assert(updatedAlice == User("Alice", 43))
 updatedAlice match
   case user: User => println(s"Alice is ${user.age} years old")
@@ -333,7 +333,7 @@ Here is how the type `User` could be used:
 // public constructor
 val alice = User("Alice", 42)
 // constructor parameters are public members
-val updatedAlice = alice.withAge(age = alice.age + 1)
+val updatedAlice = alice.withAge(alice.age + 1)
 // structural equality
 assert(updatedAlice == User("Alice", 43))
 // only typed patterns are supported in match expressions
--- a/content/struct-classes.md
+++ b/content/struct-classes.md
@@ -1,0 +1,502 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: implementation
+status: waiting-for-implementation
+title: SIP-50 - Struct Classes
+---
+
+**By: Julien Richard-Foy**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Oct 20th 2022 | Initial Version    |
+
+## Summary
+
+Introduce “struct classes”, a new flavour of class definitions that is a middle
+ground between regular classes and case classes. Like case classes, struct classes
+structurally implement `equals`, `hashCode`, and `toString`. However, unlike with
+case classes, the companions of struct classes don't have `apply` and `unapply`
+methods. This allows developers to benefit from structural equality while still
+having the possibility of making evolutions (e.g., adding a new field) in a
+backward binary compatible way.
+
+Here is a quick example that illustrates the definition and usage of a struct class:
+
+~~~ scala
+struct class User(name: String, age: Int):
+  def withName(newName: String): User =
+    copy(name = newName) // "copy" is a private method synthesized by the compiler
+  def withAge(newAge: Int): User =
+    copy(age = newAge)
+
+val alice = User("Alice", 42)
+val bob   = User("Bob", 18)
+println(bob) // prints "User(Bob, 18)"
+assert(alice != bob)
+val updatedBob = bob.withAge(bob.age + 1) // constructor parameters are public members
+assert(updatedBob == User("Bob", 19)) // structural equality
+~~~
+
+Then, it is possible to define a new version of the class definition `User`, with
+a new field, and without breaking the backward binary compatibility:
+
+~~~ scala
+// new field "email"
+struct class User(name: String, age: Int, email: Option[String]):
+  // old constructor added for compatibility
+  def this(name: String, age: Int): User = this(name, age, email = None)
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  // new convenient transformation methods
+  def withEmail(newEmail: String): User = copy(email = Some(newEmail))
+  def withoutEmail: User = copy(email = None)
+~~~
+
+## Motivation
+
+### Introductory Example Based on Case Classes
+
+Currently, case classes are often preferred to simple classes to model data
+structures because they provide convenient features:
+
+- concise syntax at the definition and usage sites (direct field access, `copy`
+  method)
+- structural implementation of `equals`, `hashCode`, and `toString`
+- support for "constructor patterns" in match expressions.
+
+However, case classes have a major drawback: they can hardly evolve in a binary
+compatible way. Adding or removing fields to a case class almost always breaks the
+binary compatibility.
+
+Consider for example the following case class definition `User` and its usage:
+
+~~~ scala
+case class User(name: String, age: Int)
+
+val alice = User("Alice", 42)
+val updatedAlice = alice.copy(age = alice.age + 1)
+assert(updatedAlice == User("Alice", 43))
+updatedAlice match
+  case User(_, age) => println(s"Alice is $age years old")
+~~~
+
+We would like to add a new field, say `email`, without breaking existing code that
+relies on the above definition of `User`.
+
+A first attempt is to provide a default value, so that existing code would always
+use that default value:
+
+~~~ scala
+case class User(name: String, age: Int, email: Option[String] = None)
+~~~
+
+Unfortunately, this does not work because the new definition of `User` has a
+different constructor type signature from the former definition of `User`. As a
+consequence, existing code constructing instances of `User` will have to be
+recompiled with the new definition of `User`.
+
+One way to fix this problem would be to re-introduce the old constructor signature
+as a secondary constructor:
+
+~~~ scala
+case class User(name: String, age: Int, email: Option[String]):
+  def this(name: String, age: Int): User = this(name, age, email = None)
+~~~
+
+Now, the previous constructor is still in the ABI of `User`, however this is still not
+enough because we also have a problem with the signature of the compiler-generated
+method `copy`. Indeed, the new version of `copy` is not binary compatible with the
+former one.
+
+One way to fix this problem would be to make the primary constructor `private`.
+If we do that, the compiler propagates the `private` visibility to the generated
+`copy` method as well. However, if we make it private, we have to provide public
+transformation methods allowing users of the class to transform instances.
+
+This means that the case class should have been designed from the beginning with
+a `private` constructor:
+
+~~~ scala
+case class User private (name: String, age: Int):
+  // public transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  
+object User:
+  // public constructor
+  def apply(name: String, age: Int): User = new User(name, age)
+~~~
+
+With this definition of `User`, our example program would look as follows:
+
+~~~ scala
+val alice = User("Alice", 42)
+val updatedAlice = alice.withAge(age = alice.age + 1)
+assert(updatedAlice == User("Alice", 43))
+updatedAlice match
+  case User(_, age) => println(s"Alice is $age years old")
+~~~
+
+Then, adding the `email` field would be achieved as follows:
+
+~~~ scala
+case class User private (name: String, age: Int, email: Option[String]):
+  // public constructor
+  def this(name: String, age: Int): User = this(name, age, email = None)
+  // public transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  def withEmail(newEmail: Option[String]): User = copy(email = newEmail)
+  
+object User:
+  // public constructor (only needed for binary compatibility)
+  def apply(name: String, age: Int): User = new User(name, age)
+~~~
+
+Unfortunately, the new definition of `User` is still not compatible with the previous
+one. Indeed, existing code that was using pattern matching does not compile anymore
+because the type `User` now has three fields instead of two, so the compiler would
+fail with a message like "Wrong number of argument patterns for User; expected
+(String, Int, Option[String])".
+
+Note that the type signature of the method `unapply` has not changed, meaning that
+it would be possible to run existing code (without recompiling it) with the new
+classfile of `User` with no issues. However, this works only when we add new
+fields, not when we remove them.
+
+There is a way to disable pattern matching on the case class by explicitly defining
+the `unapply` method to be `private`:
+
+~~~ scala
+// (previous code has been omitted for brevity)
+object User:
+  private def unapply(user: User): User = user
+~~~
+
+When we do this, the compiler does not generate the `unapply` method, which makes
+it impossible for users of the class `User` to use it in match expression in
+"constructor patterns". Users can still use "typed patterns", though:
+
+~~~ scala
+updatedAlice match
+  case user: User => println(s"Alice is ${user.age} years old")
+~~~
+
+For reference, here is the complete definition of `User` so that we can add or
+remove fields without breaking the binary compatibility:
+
+~~~ scala
+case class User private (name: String, age: Int, email: Option[String]):
+  // public constructor
+  def this(name: String, age: Int): User = this(name, age, email = None)
+  // public transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  def withEmail(newEmail: Option[String]): User = copy(email = newEmail)
+  
+object User:
+  // public constructor (only needed for binary compatibility)
+  def apply(name: String, age: Int): User = new User(name, age)
+  // explicit extractor
+  private def unapply(user: User): User = user
+~~~
+
+So, it _is_ possible to define case classes and make them evolve without breaking
+the backward binary compatibility, but that somewhat requires "undoing" some of
+the things the compiler does when we define case classes.
+
+Would it be simpler to go the other way around? Namely, not use a case class in the
+first place, and manually "re-do" some of the things the compiler does for case
+classes.
+
+### Using Regular Classes
+
+The first version of the type `User`, with support for structural equality,
+and JVM-based serialization, can be defined as follows:
+
+~~~ scala
+class User(val name: String, val age: Int) extends Serializable:
+  private def copy(name: String = name, age: Int = age): User = new User(name, age)
+  // structural implementation of "toString", "equals", and "hashCode"
+  override def toString(): String = s"User($name, $age)"
+  override def equals(that: Any): Boolean =
+    that match
+      case user: User => user.name == name && user.age == age
+      case _ => false
+  override def hashCode(): Int =
+    37 * (37 * (17 + name.##) + age.##)
+  // transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+~~~
+
+This definition can be used by our example program as follows:
+
+~~~ scala
+val alice = User("Alice", 42)
+val updatedAlice = alice.withAge(age = alice.age + 1)
+assert(updatedAlice == User("Alice", 43))
+updatedAlice match
+  case user: User => println(s"Alice is ${user.age} years old")
+~~~
+
+To add the new field `email`, we would make the primary constructor `private`
+and introduce a secondary constructor with the same signature as the former one.
+We would also update the implementations of `toString`, `equals`, and `hashCode`.
+
+~~~ scala
+class User private (val name: String, val age: Int, val email: Option[String]) extends Serializable:
+  // public constructor that matches the signature of the previous primary constructor
+  def this(name: String, age: Int): User = this(name, age, None)
+  private def copy(name: String = name, age: Int = age, email: Option[String] = email): User = new User(name, age, email)
+  // structural implementation of "toString", "equals", and "hashCode"
+  override def toString(): String = s"User($name, $age, $email)"
+  override def equals(that: Any): Boolean =
+    that match
+      case user: User => user.name == name && user.age == age && user.email == email
+      case _ => false
+  override def hashCode(): Int =
+    37 * (37 * (37 * (17 + name.##) + age.##) + email.##)
+  // transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  def withEmail(newEmail: Option[String]): User = copy(email = newEmail)
+~~~
+
+Adding one field requires adding it to the `private` constructor, to the `private`
+method `copy`, to the `toString` implementation, to the `equals` implementation,
+to the `hashCode` implementation, and to add a transformation method (here,
+`withEmail`). This arguably a lot of effort.
+
+Are there any other solutions?
+
+### Using Code Generation or Meta-Programming
+
+Other solutions used by the community are based on code generation or meta-programming.
+
+[Contraband](https://eed3si9n.com/contraband-an-alternative-to-case-class/) is an
+example of tool that generates "case class like" definitions from an _ad-hoc_
+interface definition language. Approaches based on code generation have the following
+drawbacks:
+- the definitions of the data types use another language than Scala that you need
+  to learn and get familiar with the surrounding tooling
+- they require customizing the build definition to correctly generate the sources
+- they are often not well supported by IDEs
+- they make the code harder to navigate through.
+
+The last type of solutions are based on meta-programming facilities.
+[data-class](https://github.com/alexarchambault/data-class) and
+[scalameta](https://github.com/scalameta/scalameta/blob/01cb1137cac89d1453846ef1e7acb8f4a8833e6c/scalameta/common/shared/src/main/scala/org/scalameta/data/data.scala)
+are two examples of such solutions. They seem to be hard to maintain(see e.g.
+[data-class#120](https://github.com/alexarchambault/data-class/issues/120) and
+[scalameta#2485](https://github.com/scalameta/scalameta/issues/2485)). Also, IDE
+support of meta-programming-based approaches is not always good.
+
+### Scope of the Proposal
+
+We would like to support, at the language level, the definition of data types
+that would support the following features:
+- can evolve in a binary compatible way (ie, developers can define a new version
+  of the data type, with added or removed fields, in a backward binary compatible
+  way)
+- structural implementation of `equals`, `hashCode`, `toString`
+- support of JVM-based serialization
+- primary constructor parameters promoted to public fields
+
+## Proposed solution
+
+A new "flavour" of class definition: `struct class`.
+
+### High-level overview
+
+Here is how to define the first version of the type `User` used in the Motivation
+section, with a `struct class`:
+
+~~~ scala
+struct class User(name: String, age: Int):
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+~~~
+
+This defines a type `User` with a public constructor, and two public fields `name`
+and `age`. The type `User` automatically extends `Serializable`, and automatically
+defines a private method `copy` (similar to what a case class would have). It also
+structurally override the implementation of `equals`, `hashCode`, and `toString`.
+
+Here is how the type `User` could be used:
+
+~~~ scala
+// public constructor
+val alice = User("Alice", 42)
+// constructor parameters are public members
+val updatedAlice = alice.withAge(age = alice.age + 1)
+// structural equality
+assert(updatedAlice == User("Alice", 43))
+// only typed patterns are supported in match expressions
+updatedAlice match
+  case user: User => println(s"Alice is ${user.age} years old")
+~~~
+
+Then, developers can define the following new version of `User` with an added
+field `email`:
+
+~~~ scala
+struct class User(name: String, age: Int, email: Option[String]):
+  // manually add former constructor
+  def this(name: String, age: Int): User = this(name, age, email = None)
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  // transformation method to change the email
+  def withEmail(newEmail: Option[String]): User = copy(email = newEmail)
+~~~
+
+And the existing program that uses the former definition of `User` is still binary
+compatible with the new version of `User`.
+
+Similarly, developers can define yet another version of `User` where the `email`
+field would have been removed, and still keep the backward binary compatibility:
+
+~~~ scala
+struct class User(name: String, age: Int):
+  // manually add former constructor
+  def this(name: String, age: Int, email: Option[String]): User = this(name, age)
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+  // "email" methods, for compatibility
+  def email: Option[String] = None
+  def withEmail(newEmail: Option[String]): User = this
+~~~
+
+Note that the fact that we explicitly define the transformation methods (here,
+`withName`, `withAge`, etc.) may be seen as a lack of "language support". On the
+contrary, this gives more flexibility to the developers to define custom transformation
+methods. For instance, the following snippet shows two possibilities of transformation
+methods to set the value of the `email` field:
+
+~~~ scala
+// 1st solution, as above
+def withEmail(newEmail: Option[String]): User
+// 2nd solution
+def withEmail(newEmail: String): User
+def withoutEmail: User
+~~~
+
+Here, in the 2nd solution, calling `withEmail` would always set the email to
+`Some(newEmail)`, and calling `withoutEmail` would set it to `None`. We can imagine
+similar convenience methods for fields containing collections of values.
+
+Internally, a `struct class` is implemented in a similar way as a `case class`: the
+compiler synthesizes the `private` method `copy`, and overrides the default implementations
+of `toString`, `hashCode`, and `equals`. For reference, here is what the definition
+of the first version of `User` "desugars" to:
+
+~~~ scala
+import scala.util.hashing.MurmurHash3
+
+class User(val name: String, val age: Int) extends Serializable:
+  private def copy(name: String = name, age: Int = age): User = User(name, age)
+  override def toString() = s"User($name, $age)"
+  override def hashCode() =
+    MurmurHash3.finalizeHash(
+      MurmurHash3.mix(MurmurHash3.mix(MurmurHash3.productSeed, name), age),
+      2
+    )
+  override def equals(that: Any): Boolean =
+    that match
+      case user: User => user.name == name && user.age == age
+      case _ => false
+
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+end User
+~~~
+
+### Specification
+
+A struct class is a `class` definition prefixed with `struct`:
+
+~~~ text
+TmplDef ::= 'struct' 'class' ClassDef
+~~~
+
+A struct class is required to have a parameter list that is not contextual. The
+formal parameters in the first parameter list are called _elements_ and are treated
+specially. A `val` prefix is implicitly added to such a parameter, unless the parameter
+already carries a `val` or `var` modifier. Hence, an accessor definition for the
+parameter is generated.
+
+A struct class definition `c[tps](ps_1)...(ps_n)` with type parameters `tps` and
+value parameters `ps` is handled as follows. A `private` method named `copy` is
+implicitly added to the class definition unless the class already has a member
+(directly defined or inherited) with that name, or the class has a repeated parameter.
+The method is defined as follows:
+
+~~~ scala
+private def copy[Ts](ps_1')...(ps_n'): c[Ts] = c[Ts](xs_1)...(xs_n)
+~~~
+
+Where `Ts` is the vector of types defined in the type parameter section of the class
+definition. Each `xs_i` denotes the parameter names of the parameter section `ps_i`.
+The value parameters `ps_1j'` of the first parameter list have the form
+`x_1j: T_1j = this.x_1j`. The other parameters `ps_ij'` of the `copy` method are
+defined as `x_ij: T_ij`. In all cases, `x_ij` and `T_ij` refer to the name and type
+of the corresponding class parameter `ps_ij`.
+
+Every struct class implicitly overrides some method definitions of class `scala.AnyRef`
+unless a definition of the same method is already given in the struct class itself or
+a concrete definition of the same method is given in some base class of the struct
+class different from `AnyRef`. In particular:
+
+
+- Method `def equals(that: Any): Boolean` is structural equality, where two instances
+  are equal if they both belong to the struct class in question and they have equal
+  (with respect to `equals`) constructor arguments (restricted to the class’s elements,
+  i.e., the first parameter list)
+- Method `def hashCode(): Int` computes a hash-code. If the `hashCode` method of the
+  data structure members map equal (with respect to `equals`) values to equal hash-codes,
+  then the struct class `hashCode` method does too
+- Method `def toString(): String` returns a string representation which contains the
+  name of the class and its elements.
+
+Finally, mirrors (instances of `scala.deriving.Mirror`) are not synthesized by the
+compiler for struct classes.
+
+### Compatibility
+
+The changes described in this proposal are backward binary compatible because they do
+not affect the bytecode produced by the existing language features.
+
+The changes described in this proposal are backward TASTy compatible because they
+don’t require any changes at the TASTy level.
+
+Source compatibility is preserved, and the semantics of existing valid programs is not
+changed because the proposal does not change the existing language features, it only
+adds a new keyword, `struct`.
+
+## Alternatives
+
+Besides the two solutions shown in the Motivation section (based on regular classes
+or case classes), we also considered a more powerful variant of `struct class` that
+would also automatically generate the transformation methods (`withName`, `withAge`, 
+etc. in the example). However, this solution is more complex to specify, and it also
+has some annoying drawbacks. In particular, it does not work well with all the names,
+especially with symbolic names (e.g., `+`, `%`, etc.), or names that contain
+acronyms (e.g. consider a field `httpHeaders`, should the transformation method
+be named `withHttpHeaders` or `withHTTPHeaders`?). Furthermore, it would be extremely
+complicated to specify rules to handle "smarter" transformation methods, such as
+`withoutEmail` in our example, to specifically handle fields whose type is `Option`,
+or a collection.
+
+## Related work
+
+- [Pre-SIP discussion](https://contributors.scala-lang.org/t/pre-sip-structural-data-structures-that-can-evolve-in-a-binary-compatible-way/5684)
+- [SIP-43 - Pattern matching with named fields](https://github.com/scala/improvement-proposals/pull/44) introduces a new type of patterns that
+  would be worth supporting out of the box in struct classes.
+
+## FAQ
+
+N/A.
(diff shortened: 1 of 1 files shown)


Ichoran:
I am not convinced that such modifications are wise instead of an anti-pattern, so I'm not convinced of the wisdom of including it as a core language feature.

The reason is that adding optional fields is not a neutral operation when it comes to very many likely operations.  It's _good_ to make people recompile and rethink.

For instance, suppose you add an optional `version` field, and suddenly there are a profusion of versions...and...you store the thing in a `TreeMap`.  Now what?  If the `Ordering` doesn't respect the new field, the version will get randomly lost--it's not _just_ that you default to the old behavior, but rather that you'll get a _random_ selection from the new things available.  But there's no mechanism by which to automatically respect the new field (especially if it's a custom `Ordering`).

Suppose you have a `Permissions` class with two fields that represent superuser level for two different areas, and you decide to add a third area.  You want to have the idea of a maximal superuser, to test the full capacity you've given users, which you implement like so:

```scala
val superuser = Permissions(
  allUsers.map(u => u.level1).max,
  allUsers.map(u => u.level2).max
)
```

Except you're now silently broken, with no clue _even by the requirement to recompile_ suggesting that your logic is broken.

Or what about custom serialization?

Or what about transformation of a data field where the code doesn't use `copy`?

```scala
user match
  case User(first, last) => User(capitalize(first), capitalize(last))
```

Oh boy, I just lost all my addresses!

It seems that to get this right, you require a whole list of best practices, and leaning on best practices is the least desirable route to correctness.  And you have to expect people will have done them in anticipation of possible changes, without a recompile (which is at least slightly hinted at by having it be a `struct class`, but maybe not, since there might be other reasons to like the more limited set of synthetic methods).

I grant that all these issues already exist, and that the boilerplate-heavy manual method allows you to get around them (as do the other mechanisms), with the same risks.  But just like `.asInstanceOf[X]` discourages a danger-prone pattern, to me, it seems like the difficulty of doing this is entirely appropriate given the danger of silently adding fields without even so much as a warning of binary incompatibility to alert people to re-check their logic.

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+object User:
+  // public constructor
+  def apply(name: String, age: Int): User = new User(name, age)
+~~~
+
+With this definition of `User`, our example program would look as follows:
+
+~~~ scala
+val alice = User("Alice", 42)
+val updatedAlice = alice.withAge(age = alice.age + 1)
```
rjolly, line 138: Should be alice.withAge(alice.age + 1)

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+  // transformation methods
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+~~~
+
+This definition can be used by our example program as follows:
+
+~~~ scala
+val alice = User("Alice", 42)
+val updatedAlice = alice.withAge(age = alice.age + 1)
```
rjolly, line 241: Should be alice.withAge(alice.age + 1)

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+defines a private method `copy` (similar to what a case class would have). It also
+structurally override the implementation of `equals`, `hashCode`, and `toString`.
+
+Here is how the type `User` could be used:
+
+~~~ scala
+// public constructor
+val alice = User("Alice", 42)
+// constructor parameters are public members
+val updatedAlice = alice.withAge(age = alice.age + 1)
```
rjolly, line 336: Should be `alice.withAge(alice.age + 1)`

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+### Scope of the Proposal
+
+We would like to support, at the language level, the definition of data types
+that would support the following features:
+- can evolve in a binary compatible way (ie, developers can define a new version
+  of the data type, with added or removed fields, in a backward binary compatible
+  way)
+- structural implementation of `equals`, `hashCode`, `toString`
+- support of JVM-based serialization
+- primary constructor parameters promoted to public fields
```
raulraja, line 308: Thanks, Julien, for looking for a solution to the bin compact issues we face with case classes.
Would `struct` classes implement `Product` as case classes do?

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+A struct class is required to have a parameter list that is not contextual. The
+formal parameters in the first parameter list are called _elements_ and are treated
+specially. A `val` prefix is implicitly added to such a parameter, unless the parameter
+already carries a `val` or `var` modifier. Hence, an accessor definition for the
+parameter is generated.
+
+A struct class definition `c[tps](ps_1)...(ps_n)` with type parameters `tps` and
+value parameters `ps` is handled as follows. A `private` method named `copy` is
+implicitly added to the class definition unless the class already has a member
+(directly defined or inherited) with that name, or the class has a repeated parameter.
```
raulraja, line 435: Is it just the name `copy` used to decide generation? Or would other parts of an existing function signature, such as the return type used to determine if it needs to generate a copy method?

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+  are equal if they both belong to the struct class in question and they have equal
+  (with respect to `equals`) constructor arguments (restricted to the class’s elements,
+  i.e., the first parameter list)
+- Method `def hashCode(): Int` computes a hash-code. If the `hashCode` method of the
+  data structure members map equal (with respect to `equals`) values to equal hash-codes,
+  then the struct class `hashCode` method does too
+- Method `def toString(): String` returns a string representation which contains the
+  name of the class and its elements.
+
+Finally, mirrors (instances of `scala.deriving.Mirror`) are not synthesized by the
```
raulraja, line 465: Could you elaborate on why `struct` classes won't get synth Mirror instances? 
If there are bin compact reasons or another rationale why `struct` classes don't get these may be good to have them explained here.

Currently, Scala 3 docs at https://docs.scala-lang.org/scala3/reference/contextual/derivation.html#mirror say

> Instances of the Mirror type class are generated automatically by the compiler unconditionally for:
> enums and enum cases,
> case objects.

> Instances for Mirror are also generated conditionally for:
> case classes where the constructor is visible at the callsite (always true if the companion is not a case object)
> sealed classes and sealed traits where:
> there exists at least one child case,
> each child case is reachable from the parent's definition,
> if the sealed trait/class has no companion, then each child case is reachable from the callsite through the prefix of the type being mirrored,
> and where the compiler can generate a Mirror type class instance for each child case.

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+
+Source compatibility is preserved, and the semantics of existing valid programs is not
+changed because the proposal does not change the existing language features, it only
+adds a new keyword, `struct`.
+
+## Alternatives
+
+Besides the two solutions shown in the Motivation section (based on regular classes
+or case classes), we also considered a more powerful variant of `struct class` that
+would also automatically generate the transformation methods (`withName`, `withAge`,
```
raulraja, line 484: I feel not having this kind of feature will be frustrating to users. Most will reach out to struct classes as a way to deal with lighter-weight case classes or solve the bin compact issues, but the problem of updating and dealing with immutable structures does not go away.

We have a similar problem in Kotlin in the context of value classes where they may include special syntax for immutable copy updates https://github.com/Kotlin/KEEP/blob/master/notes/value-classes.md#updating-immutable-classes. I don't propose we follow a similar approach; just pointing out that not having a feature like this and forcing users into considerable boilerplate to get it working may introduce more issues than `struct` classes solve. In contrast, case classes that at least give you `copy` are not great to work with and with bin compact issues, but they are still an alternative to writing a bunch of boilerplate.

I think `struct` classes will be even better if they improve on `case classes` not just from a bin compact standpoint but also the state of what working with immutable data looks like in Scala.

content/struct-classes.md
sjrd: If we synthesize a `Mirror`, we get static access to its `MirroredElemTypes` type member. When you then add a field, the type behind that type alias changes in a binary-incompatible way.

bjornregnell:
I'd like to consider this from a beginner perspective: is yet-another kind of class really worth it? 

We already need to explain the difference between plain classes and case classes, which digs into tricky things such as equality. 

But does a beginner need to bother about such `struct` classes? Well, if it is in the language learners will sooner or later encounter the strange word `struct` and need to have at least a superficial understanding of what they are and why the are in the language. But trying to understand binary compatibility is probably not the best spending of a learner's grit...

Perhaps an annotation similar to `@struct case class` would be less "intrusive" in terms of generating less "conceptual load" when learning the language? The rationale for having "only" an annotation, instead of a whole new kind of class, would be that this is a specialized thing for those that care about implementation details such as binary compatibility. This has after all something to with code generation similar to `@tailrec`.

sjrd:
It changes the public API of the class. As such, it falls way beyond what an annotation is supposed to be able to do.

odersky:
I share Bjorn's sentiment. Adding another kind of class with all its special rules adds considerable complexity to the language. 

But I am also dubious about "hiding" the complexity under an annotation `@struct`. Without could open doors to all sorts of complicating additions that seem easy since it's just an annotation that is added.

In my mind, an annotation can

 - influence how code is generated
 - add some restrictions how a feature is used

`@targetName` is a good example, which does both. But it should not cause API changes. 

Which leads me to ask: Do we really need this? What do other languages have to offer in this respect? Is this an area where Scala falls clearly short, or is it an area where we think that Scala should be leading the pack? If the latter, why?

Otherwise, it looks like a major complication for a relatively minor use case. One could argue that we could just keep using normal classes and implement structural operations by hand.  Or, use case classes with the tweaks described in the proposal. Adding a third kind of class is a very large complication. Is it really worth it?

nafg:
I think it's mostly useful for library authors who want to provide case classes and evolve them. Most languages don't have something comparable.

Maybe it should be a keyword modifier of case class.

Or maybe we should have a more general mechanism to limit what case class normally generates.

Regarding generating withXXX methods it would be nice to have a way to do that but maybe orthogonal to this. But I'd actually rather modXXX which take a function instead of a constant. Or even better, lenses, which doesn't currently seem possible in Scala 3. But it's not specific to this new kind of class.

I also don't think struct is the best name. It's not more of a value class than case classes. I guess without an extractor "case" may not be the best name. I guess it's just a simple product type, or data class.

Coming back to the issue of learning a new keyword, I think `data class` would be much less of an issue than `struct class`. It's basically self explanatory. And besides, Kotlin has it, although I'm not sure how it compares.

Anyway it's not hard to teach that case classes are data class plus pattern matching and a public copy method.

Ichoran:
If it is _really_ important to support this use case with less boilerplate, can we extend `derives` to also derive parts of the case class functionality?

So if you want just `copy` and element-by-element equality, but nothing else, you

```
class Foo(i: Int) {
  derives def copy(_): Foo
  derives override def equals(_): Boolean
}
```

where `_` in this case means "let the deriving fill in the argument list".  (Arguably, it could just be `derives def copy` and the return type would be filled in too.)

If it's really really important to have even less boilerplate, then a predictable assemblage of such things could be created, maybe something like

```
transparent trait DataClass derives {
  override def equals
  override def hashCode
}
```

which you would then deploy with

```
class Foo(val i: Int, val j: Int) derives DataClass {}
```

as if it were a typeclass, except it would work for this quasi-structural-type instead.

Then we support more use-cases, we don't have to argue about exactly what does and does not go into a struct class, we punt on anything difficult and make the user implement it themselves, and what is entailed by "data class" becomes somewhat more explicit.

I'm still not sure this is a very good idea, but I think it has substantially fewer dangers and greater benefits than a `struct class` feature.  (One big downside--it's a bigger change to the language and more work to implement.)

content/struct-classes.md
```diff
@@ -0,0 +1,502 @@
...
+~~~
+
+Where `Ts` is the vector of types defined in the type parameter section of the class
+definition. Each `xs_i` denotes the parameter names of the parameter section `ps_i`.
+The value parameters `ps_1j'` of the first parameter list have the form
+`x_1j: T_1j = this.x_1j`. The other parameters `ps_ij'` of the `copy` method are
+defined as `x_ij: T_ij`. In all cases, `x_ij` and `T_ij` refer to the name and type
+of the corresponding class parameter `ps_ij`.
+
+Every struct class implicitly overrides some method definitions of class `scala.AnyRef`
```
kubukoz, line 449: Shall struct classes be implicitly `final`?

content/struct-classes.md
julienrf: I don’t have a strong opinion about that. What would be the motivation for implementing `Product`?

content/struct-classes.md
julienrf: I copied and adapted the specification of case classes from [here](https://scala-lang.org/files/archive/spec/2.13/05-classes-and-objects.html#case-classes). The reason is that I want the same schema to be applied both to struct classes and case classes.

In that specification, only the name is used, but not the signature. I’ve noticed that the Scala 2 compiler does not synthesize a `copy` method if I already define one (even if it’s signature does not match the case class fields). However, I’ve just noticed that the Scala 3 compiler does not behave the same here: if you add a method `copy`, the compiler adds another one as an overload. I’ve created https://github.com/lampepfl/dotty/discussions/16309 to clarify whether this is expected or not.

sjrd:
> Which leads me to ask: Do we really need this? What do other languages have to offer in this respect? Is this an area where Scala falls clearly short, or is it an area where we think that Scala should be leading the pack? If the latter, why?

As mentioned in the Motivation section (under "Using Code Generation or Meta-Programming"), the need is clearly there. Several libraries have reinvented the pattern using a variety of ways. And that's not even counting libraries who simply write everything by hand (e.g., the Scala.js linker library).

This is a very well-known pattern, reinvented many times, but requiring a lot of boilerplate if done in user space, as well as a lot of care not to mess up. Isn't that the best motivation we could possibly have for putting something into the language? Isn't it the same kind of motivation that led to `derives`? Or even `object` in the first place, as opposed to the singleton pattern?

content/struct-classes.md
julienrf: That’s a good question. I am tempted to stay conservative and follow the same behavior as case classes (ie, forbid struct-to-struct inheritance, struct-to-case inheritance, and case-to-struct inheritance, but not make struct classes `final`). 

That being said, if we allow a class to extend a struct class, we should probably make struct classes extend [Equals](https://www.scala-lang.org/api/current/scala/Equals.html) and use its `canEqual` method to implement `equals` to make sure we don’t generate an implementation of `equals` that is not an equivalence relation.

content/struct-classes.md
julienrf: I have clarified the motivation for not having mirrors in 6a9fdee

content/struct-classes.md
kubukoz: My take would be that these should be final. What would be the usecase for subclassing these?

kubukoz:
I wonder if we should have a `structEnum` in Scala 3 as well 🤔 - allow library authors to benefit from `struct class` but also the conciseness of enums.

kubukoz:
Riiight. Forget that one. :)

bjornregnell:
I guess the SIP should mention its relation to Java's `record` https://openjdk.org/jeps/359 with similarities and differences?

bjornregnell:
If we think binary compatible case classes without copy etc is important enough to include in the language, I'd like to challenge the keyword `struct` which I think is a bit cryptic. The terms `record` or `data` are less cryptic IMHO - perhaps there are others even better. I think the SIP should discuss pros and cons of different soft keyword proposals.

morgen-peschke:
> The terms `record` or `data` are less cryptic IMHO - perhaps there are others even better.

It's probably better to avoid "record", depending on how sure we are that Java is going to use it. There's enough similarity in what https://openjdk.org/jeps/359 is proposing that using the same keyword will likely be more confusing than helpful.

julienrf:
> We already need to explain the difference between plain classes and case classes, which digs into tricky things such as equality.
> 
> But does a beginner need to bother about such `struct` classes? Well, if it is in the language learners will sooner or later encounter the strange word `struct` and need to have at least a superficial understanding of what they are and why the are in the language. But trying to understand binary compatibility is probably not the best spending of a learner's grit...

That’s a good point. I agree that the weakness of this proposal is that `struct classes` are very specific. However, I would argue that the situation was a bit similar for opaque types. Opaque types are useful only in the case of performance-critical programs. Otherwise, encapsulation can be achieved in a simpler way with plain classes. Do we need to teach beginners opaque types? I would expect they would come quite late in the curriculum. Several other features don’t target beginners and yet are part of the language (multiversal equality, transparent traits, erased terms, matchable, `AnyKind`, `TypeTest`, …).

My strategy to reduce the overhead for explaining `struct classes` is to make them very similar to `case classes`. The same mechanisms are used to desugar them, the difference is that `case classes` are “more specific” than `struct classes` (in the sense that they synthesize more things such as the `unapply` method in the companion).

julienrf:
> If we think binary compatible case classes without copy etc is important enough to include in the language, I'd like to challenge the keyword `struct` which I think is a bit cryptic. The terms `record` or `data` are less cryptic IMHO - perhaps there are others even better. I think the SIP should discuss pros and cons of different soft keyword proposals. Or even if it should be a single hard keyword. like just `record` (remember Pascal...)

The keyword `struct` stands for “structural”, which relates directly to the semantics of `struct class` definitions: class definitions that automatically implement structural equality (as opposed to the default object reference equality).

However, it is true that `struct` has historically also been used in C and C++ to define “data structures”, with no relation with structural equality. The keyword `struct` is also used in Swift and Rust with a similar meaning (“data structure”).

I agree that this can bring some confusion.

Other languages implementing a similar feature use the following keywords:
- `record` (Java), a restricted variant of `struct class` (Java records can’t
  extend parent classes, can’t be extended, and can’t have mutable fields --- even
  private ones),
- `data class` (Kotlin), which are equivalent to `case classes` in Scala and have
  the same issues regarding binary compatibility.

I am not sure we should reuse those keywords just because Java and Kotlin use them. Especially `data` would be tricky because we would have to explain that a `data class` in Kotlin is not like a `data class` in Scala but like a `case class`. About `record`, the name does not really say what it means, so I wonder if people may get confused to remember the difference between a `record class` and a `case class` (whereas I hope people can remember the difference between `struct class` and `case class` because of the reference to “structural” in `struct`).

Maybe we need to come up with a different word then. What do you think?

sjrd:
> The keyword `struct` stands for “structural”

What about `structural`, then?

content/struct-classes.md
raulraja: The motivation would be that struct classes work with derived Mirror instances, which include `fromProduct`.

julienrf:
> What about `structural`, then?

Yes, I am fine with that.

julienrf:
@smarter Could you please review the proposal?

bjornregnell:
I like `structural class` better than `struct class` as it makes it less cryptic and less similar to the C/C++ keyword. I guess i t is more likely that people will spell it out when speaking about them and the connection to structural equality becomes more evident. And we don't say "struct equality"..,

bjornregnell:
@julienrf I think it would be good if you include your analysis above of the similarity to other languages in the proposal, explaining the relation between Scala's structural classes to Java's record, Kotlin's data class, C/C++'s struct etc

bjornregnell:
And maybe also include in the proposal text an analysis of @Ichoran 's alternative based on `derives` with pros and cons of using derivation instead of a new keyword `structural class`.

smarter:
Just to recap some of the discussion:

From @bjornregnell:
> I'd like to consider this from a beginner perspective: is yet-another kind of class really worth it?
> [...]
> Perhaps an annotation similar to `@struct` case class would be less "intrusive" in terms of generating less "conceptual load" when learning the language?

Reply from @sjrd:
> It changes the public API of the class. As such, it falls way beyond what an annotation is supposed to be able to do.

Reply from @odersky :
> In my mind, an annotation can
>
> -  influence how code is generated
> -   add some restrictions how a feature is used

So, could we in fact design an `@struct` annotation that respects those constraints and let us evolve classes in an API- and ABI-compatible way? (let's set aside any discussion on what the exact name of the annotation should be for now.)

To make our job easier, let's add a few caveats that do not contradict the constraints given above:
1. `@struct` can only be added on a case class.
2. Adding or removing `@struct` is **not** ABI-compatible.
3. We will only support _adding_ and not _removing_ fields, the latter seems much harder to support given our constraints, and I think is less important.

Let's consider the **API** of a case class which our annotation is not allowed to change (adapted from `-Xprint:typer`):
```scala
// Source code: case class User(name: String, age: Int) 
/*case*/ class User(val name: String, val age: Int) extends Product, Serializable:
  def copy(name: String = this.name, age: Int = this.age): User
  def _1: String
  def _2: Int
object User:
  def apply(name: String, age: Int): User
  def unapply(x$1: User): User
```

For each method, we need to define what the impact of our annotation would be:
- `unapply`: since we cannot remove it, but we don't want users to call it, let's just emit an error when someone tries to call it. You might think this is cheating but consider that in the future we may have [pattern-matching with named field](https://github.com/scala/improvement-proposals/pull/44), in which case we'll be able to relax this restriction and allow pattern-matching on `@struct` case classes where all fields are named.
- `apply`: Scala 3 has [universal apply methods](https://docs.scala-lang.org/scala3/reference/other-new-features/creator-applications.html) where the compiler synthesizes apply methods on the fly which get compiled into constructor calls, so I think we can just drop the apply method from the generated code and rely on this feature. This is API but not ABI compatible, which is fine as per our second caveat.
- `_1`, `_2`: We can just leave them unchanged, this is fine as per our third caveat.
- `copy`: This is where things get interesting, can we keep the current API of `copy` while making it possible to add fields in an ABI-compatible way? Suppose that during typechecking we see:
  ```scala
  @struct /*case*/ class User(val name: String, val age: Int) extends …:
     def copy(name: String = this.name, age: Int = this.age): User
     // …
  def foo(user: User) = user.copy(name = "foo", age = 12)
  ```
  Then we could rely on the presence of `@struct` to perform the following rewriting in a phase after typechecking so as to not affect the user-visible API:
  ```scala
  @struct /*case*/ class User(val name: String, val age: Int) extends …:
     def withName$(name: String): User = new User(name, this.age)
     def withAge$(age: Int): User = new User(this.name, age)
     // …
  def foo(user: User) = user.withName("foo").withAge(12)
  ```
  In other words, the rewriting would **remove** the `copy` method, so it wouldn't be part of the ABI, replace it by `withX$` methods (that are not visible to the user) and rewrite calls to `copy` into chained calls to `withX$`. I believe this is enough to allow for adding an extra field in an API- and ABI-compatible way (you'd still need to add the secondary constructor by hand as in this SIP).

WDYT, did I miss anything crucial?

nafg:
I still don't think struct or structural is a good name.

It's true that, like case classes, equality is determined not by reference but by the value's structure. But that is usually referred to as _value equality_. To me, "structural" only evokes structural type systems. So I don't like the argument:

> The keyword `struct` stands for “structural”, which relates directly to the semantics of `struct class` definitions: class definitions that automatically implement structural equality (as opposed to the default object reference equality).

In fact, if the important semantics is the generated equality then I would use a name that evokes "generates an equals method." Because a generated `equals` method is obviously going to be based on the structure. For reference equality you just don't override it. So if the point is to evoke "class definitions that automatically implement structural equality (as opposed to the default object reference equality" then `struct` is missing the most important part of the sentence: that the thing that's being done structurally is the **equality**.
 
> However, it is true that `struct` has historically also been used in C and C++ to define “data structures”, with no relation with structural equality. The keyword `struct` is also used in Swift and Rust with a similar meaning (“data structure”).

In fact I would classify that as a point in _favor_ of `struct`, because it conveys a similar message to `data class`, which is the broader point, that it's a data class.


> Especially `data` would be tricky because we would have to explain that a `data class` in Kotlin is not like a `data class` in Scala but like a `case class`.

That really doesn't bother me. Plenty of things are different between Kotlin and Scala. It's not our job to play defense and make sure someone used to Kotlin is allowed to assume that any keyword in Kotlin means the same thing in Scala.

The more important thing is that when someone reads code containing this new thing, if they've never seen it before can they basically figure out what's going on? And I would argue that if it says `data class` then they go "this is a data class, I don't know its exact semantics but I get what this code is trying to do." But if it says `struct class` then they are like "I have no idea what this code is doing, I better start searching the docs or googling."

At the point where you're actually reading the docs, assuming the docs are not too dense it really doesn't matter what it's called. Learning the semantics is learning the semantics, at that point the choice of keyword is irrelevant. Except that either way the docs should explain what the name is there to convey, and hopefully it's an explanation that sits well with the reader.

odersky:
I am still very skeptical about the syntax extension. I am not looking forward to having to answer questions like "What is the difference between struct classes and case classes and when should I use which?". If I talk to a newcomer the answer will feel like splitting hairs and unnecessary complexity. 

There are alternatives we should consider. Here is one: 

The purpose of `struct` classes is to give convenient implementations of `equals`, `hashCode` and `toString` without any of the other niceties added by case classes. All three operations are implemented in terms of `Product`. So, alternatively, we could simply inherit a base trait 
```scala
    trait Structural extends Product:
      override def hashCode: Int = ...
      override def equals(other: Any): Boolean = ...
      override def toString: String = ....
```
`Structural` can be defined in the standard library. The implementations of its members use the `productArity` and `productElement` methods defined in `Product`. The only downside here is that we'd have to manually implement 
`productArity` and `productElement` in inheriting classes. So a "struct" class would look like this:
```scala
class Person(val name: String, val age: Int) extends Structural:
  def productArity = 2
  def productElement(n: Int) = n match
    case 0 => name
    case 1 => age
```
All this can be done today, with no language extension. 

A possible objection is that defining `productArity` and `productElement` manually feels boilerplatey. This could be addressed by adding the following rule to the language:

 - if a concrete class inherits `Product` and does not define or inherit `productArity` and `productElement` methods other than the abstract methods defined in `Product`, then such methods are automatically synthesized from the parameters of the class constructor. 

This rule is very much "do what I mean". It requires no syntax, and does not invalidate existing code.

bjornregnell:
I think your proposal to extend a sub-type of Product from the standard library seems much more appealing than adding another type of class, and I'm fine with the productX synthesis "magic", as this is for library authors to consider and not beginners. (Not sure if Structural is the best name but CanValueBasedEquals or similar is perhaps to long.)
 
Could we also have a rule that synthesize withX methods for all val X if not implemented manually, to scrap even more boilerplate? @odersky

bjornregnell:
I think the proposal by @smarter to use an annotation is less "intrusive" than adding a new type of class to the language, but it is still reads more like a cryptic special-case than simply extending a marker trait as in the last proposal by @odersky so if the latter trick can do the job, I think that seems most promising, esp. if the withX methods can be generated.

odersky:
For me, either annotation or marker trait would work. Since we have work on annotation macros under way, we should also check whether this could simply be an annotation macro on a class.

But it looks like neither of these approaches can support the `withXXX` methods, since that's an API change. So I am afraid you have to write these by hand if they are needed. On the other hand: For how many case classes do you invoke the copy method, ever? For me that's maybe somewhere between 5% and 10%. So is the whole thing really worth the complication?

smarter:
> But it looks like neither of these approaches can support the withXXX methods, since that's an API change. 

The point of my annotation proposal is that we don't need the withXXX methods in the API if we can have an ABI-compatible copy method.

odersky:
@smarter And, could this possibly be done with a macro annotation? That would be interesting.

smarter:
It's possible the logic for copy I outlined could be implemented with a macro annotation if we can generate an inline def for copy, but the special-casing for unapply would require something else to emit an error, especially if we further complicate it to allow named pattern matching in the future.

smarter:
> I agree that the weakness of this proposal is that struct classes are very specific. However, I would argue that the situation was a bit similar for opaque types. Opaque types are useful only in the case of performance-critical programs. 

In my mind, opaque type are a tool for abstraction: they let a class author hide implementation details of their type members from users in a convenient way (previously, this could only be done with much more boilerplate with an abstract type member in a parent class and a concrete type member in a private subclass). If they were purely useful for performance reasons I agree they wouldn't be a great feature to have in the language.

sjrd:
Perhaps it will help the discussion to see real-world use cases for this feature. They abound in the Scala.js linker-interface API. Here are some choice excerpts:
* [`Semantics.scala`](https://github.com/scala-js/scala-js/blob/v1.11.0/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Semantics.scala)
* [`ESFeatures.scala`](https://github.com/scala-js/scala-js/blob/v1.11.0/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/ESFeatures.scala)
* [`StandardConfig.scala`](https://github.com/scala-js/scala-js/blob/v1.11.0/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala)

And here are some real-world *evolution* situations that happened for these classes:

* Standard addition of a field: https://github.com/scala-js/scala-js/commit/b1137d5bc6aa94b6449c02839dd9d35ed5f5de56#diff-756d2bbf166adf5c33788585e4c3f69dd20f99f8a3581ac0121f4b3c45e1c368 (here in the middle, but it *could* have been at the end)
* Deprecation for the "setter" of a field: https://github.com/scala-js/scala-js/commit/a1f52c73a1e961caf3f140359287ba8347e30f99#diff-756d2bbf166adf5c33788585e4c3f69dd20f99f8a3581ac0121f4b3c45e1c368
* Deprecate one field altogether, and *replace* it with a field of a more general type: https://github.com/scala-js/scala-js/commit/c4aa18becd16ea9e9b94afe42d1819928cf3cd28#diff-daf708dbf113c60b6bda05022ba4098fce40d470cdc15a44f17bad24f393c8d0 (this cannot be done if we only allow to *add* fields)

In addition, when we are really careful, we don't even call the private constructor with all the fields from the companion object. Instead, we declare another 0-arg constructor that calls the main constructor, and we call the 0-arg constructor from the companion object. That's because the constructor called from the companion object becomes *public* in the bytecode, so any modification requires manual MiMa filters. `ESFeatures` is a well-designed example.

So where's the true boilerplate here? It's `equals`, `hashCode`, `toString`, and the `private copy`. If I got that, I would be happy.

---

About beginner-friendliness:

* No, it's unlikely that a beginner will need this. It is a feature for libraries.
* About when to use `structural` vs `case`: my answer remains straightforward:
  * Do you want/need to pattern-match on it, i.e, write `case Foo(...) =>`? Then it's a `case class`. But then maybe you should be using `enum` in the first place?
  * Otherwise, it's a `structural class`.

In addition, doing these things by hand is very hard and error-prone, and requires foresight that you can only gain by being bitten before. That means very few people in the world can do it correctly for new APIs. And that, in turn, means that most APIs have to be broken, weakening our ecosystem for everyone, including beginners. If the good, correct way of doing it was in the language, library maintainers would use it, and it would strengthen the overall compatibility of our ecosystem, and therefore its quality.

odersky:
I believe if this is not a feature for beginners we will not want to make an analogy to case classes like `struct class` vs `case class`. That sends the wrong message. We need something considerably more subtle than that.

sjrd:
Eventually I believe the correct message would be not to talk about `case class`es at all. If you need pattern matching, it's because you're part of an `enum`. And otherwise, if you want a bag of fields with structural equality, use a `structural class`.

The `case class`/`case object` will become the advanced feature (for complicated pattern-matchable hierarchies that don't fit the restrictions of `enum`).

content/struct-classes.md
julienrf: Mirrors of struct classes could create compatibility issues: consider code compiled with a mirror synthesized for a version x of a data type, and code compiled with a mirror synthesized for a version y of the same data type. Both program may refer to the “same” type from those mirrors, but those types may be different in practice, leading to runtime issues.

bjornregnell:
> The `case class`/`case object` will become the advanced feature (for complicated pattern-matchable hierarchies that don't fit the restrictions of `enum`).

I have been teaching case classes and enum fpr two years now and I find the simple enums `enum Color { case Black, Red }` is easy to grok for a beginner but ADT-stuff with enums I have reverted to take *after* case-classes and traits, as I have found that a beginner often wants to connect to a mental model of types and classes when understanding how and why more advanced enums work.

I also find case classes for data modelling to be a fine starter (*before* ADT etc) for example things like `case class Person(name: String, age: Int)` together with simple constructor patterns in `match`, before a more advanced discussion about the motivation behind `sealed` etc.

In short: I don't agree with @sjrd that case-classes are a more advanced concept than general enums...

julienrf:
> @julienrf I think it would be good if you include your analysis above of the similarity to other languages in the proposal, explaining the relation between Scala's structural classes to Java's record, Kotlin's data class, C/C++'s struct etc

I’ve added that in [00a9276](https://github.com/scala/improvement-proposals/pull/50/commits/00a927679d4eae4aa7b8fd32bac45a502bd9d22e).

nafg:
Can we PLEASE use a name that is a lot more self-explanatory and meaningful than struct(ural)?

nafg:
> The only downside here is that we'd have to manually implement `productArity` and `productElement` in inheriting classes. So a "struct" class would look like this:
> 
> ```scala
> class Person(val name: String, val age: Int) extends Structural:
>   def productArity = 2
>   def productElement(n: Int) = n match
>     case 0 => name
>     case 1 => age
> ```
> 
> All this can be done today, with no language extension.

IINM this can easily be implemented in a trait via runtime reflection or macros

julienrf:
I’ve pushed a commit, e5084c5, with the following changes:
- add a note about annotations and why they can’t be used to implement a solution in Scala 3
- add an analysis of the alternative solution proposed by Martin Odersky, based on `Product` methods

content/struct-classes.md
```diff
@@ -0,0 +1,627 @@
...
+especially with symbolic names (e.g., `+`, `%`, etc.), or names that contain
+acronyms (e.g. consider a field `httpHeaders`, should the transformation method
+be named `withHttpHeaders` or `withHTTPHeaders`?). Furthermore, it would be extremely
+complicated to specify rules to handle "smarter" transformation methods, such as
+`withoutEmail` in our example, to specifically handle fields whose type is `Option`,
+or a collection.
+
+### Leverage `Product`
+
+We could introduce a new trait, say `Structural`, to the standard library,
```
sideeffffect, line 534: Would it be a good idea from then on to have all case classes extending specifically `Structural` instead of the more general `Product`?
Aren't _case classes_ specific cases of `Structural`?

content/struct-classes.md
```diff
@@ -0,0 +1,627 @@
...
+which would implement the methods `toString`, `hashCode`, and `equals` in 
+terms of methods provided by the trait `Product`:
+
+~~~ scala
+trait Structural extends Product:
+  override def hashCode: Int =
+    productIterator.foldLeft(17)((hash, field) => 37 * (hash + field.##))
+  override def equals(other: Any): Boolean =
+    other match
+      case that: Structural if that.canEqual(this) =>
```
sideeffffect, line 544: Wouldn't it be better to turn it around and have
```scala
... if this.canEqual(that) =>
```
So that I can potentially override `canEqual` in the `Structural` sub-class in those rare cases where I want to such thing?

content/struct-classes.md
```diff
@@ -0,0 +1,627 @@
...
+
+~~~ scala
+class User(val name: String, val age: Int) extends Structural:
+  private def copy(name: String = this.name, age: Int = this.age): User =
+    User(name, age)
+  def withName(newName: String): User = copy(name = newName)
+  def withAge(newAge: Int): User = copy(age = newAge)
+~~~
+
+Developers would still have to manually define and maintain the `copy` method.
```
sideeffffect, line 613: Why can't the private `copy` method also be automatically synthesize?

julienrf: You are right that technically, the compiler can synthesize it. To be honest, I find it weird that the compiler does a special treatment to this trait. If it also synthesizes new methods (instead of “just” magically implementing the inherited methods), it crosses the line “that’s too many features for just extending a trait, it deserves a proper keyword in the language”.

Maybe that’s very subjective… We do have language features associated with marker traits: `Selectable` and `Dynamic`.

content/struct-classes.md
julienrf: I’ve followed the pattern described here: https://www.artima.com/pins1ed/object-equality.html

content/struct-classes.md
julienrf: That makes sense, yes. This is something we could discuss further in case the marker trait approach is selected.

julienrf:
Thank you @bjornregnell for the review. Please @smarter and @raulraja could you also share your conclusion regarding this proposal?

content/struct-classes.md
sideeffffect: I see. But still. Don't you see utility in a sub-class of `Structural` to decide to be `equals`able with other classes, for example case classes?
```scala
class UserDto(name: String, age: Int)

class User(val name: String, val age: Int, val email: Option[String]) extends Structural:
  private def copy(name: String = this.name, age: Int = this.age, email: Option[String] = this.email): User =
    User(name, age, email)
  def with...
  def canEqual(other: Any): Boolean = other match:
    case _: User => true
    case _: UserDto => true
    case _ => false

User("John", 21, None) == UserDto("John", 21)  // returns `true`
User("John", 21, Some("john@email.com")) == UserDto("John", 21)  // also returns `true`
```

content/struct-classes.md
```diff
@@ -0,0 +1,627 @@
...
+or a collection.
+
+### Leverage `Product`
+
+We could introduce a new trait, say `Structural`, to the standard library,
+which would implement the methods `toString`, `hashCode`, and `equals` in 
+terms of methods provided by the trait `Product`:
+
+~~~ scala
+trait Structural extends Product:
```
sideeffffect, line 539: Another idea: Is it thinkable to "upstream" these definitions to `Product`? That way we wouldn't need `Structural` to begin with. The definitions seem reasonable and whoever doesn't like them can always override them.

smarter (commented):
Overall I don't really know how to proceed. Adding a keyword defining a new kind of class would be a step back for usability (people are already confused about class vs case class vs trait), and everything else has flaws/unknowns.

smarter:
Here's a possible workaround to our self-imposed limitations on macro annotations: if we don't want to allow them to add new definitions visible during typechecking, perhaps instead we could have the annotation expansion check if a definition with the correct body is defined, and emit an error if this isn't the case. The compiler is already capable of doing rewrite using `-rewrite` and we'd like to expose this to IDEs so you can just click on the error and have it fixed automatically. This could be a way to get "built-in codegen" that doesn't involve other tooling.

raulraja:
@julienrf After reviewing the proposals and discussion, I'm inclined to recommend we accept the proposed feature.
Thank you for all revisions and all the time you put into this, and for trying to improve Scala.
I agree with many comments, including the last one from @smarter, that introducing a new type of class or new keywords would be yet another thing to explain about classes in Scala, and it's a step back. I agree with @bjornregnell that having a `extends WithStructuralEq` or similar extension mechanism would be better.

smarter:
The workflow would be as follow:
1. Update your class (e.g., by adding an extra class parameter)
2. Get a compiler error
3. Click on the "Fix it" button on the compiler error (or run the compiler with `-rewrite`) to fix the code

Ichoran:
@raulraja - I'm not sure what it means to recommend accepting the feature when the details say the feature is triggered by a keyword, but you seem to agree that `extends SomeTrait` is better?  Doesn't that mean that the proposal should be reworked to a sufficient extent to figure out any wrinkles that might come up when using a trait implementation?

(For instance, if there is copy, does the trait have a copy method in it, and if so, with what type signature?  Is it a single trait that pulls in exactly what this proposal suggests, or, since it's trait-based anyway, do we allow a more fine-grained approach?  For instance, what if you just want equality and printing, and don't care about any of the other stuff?  I am constantly using case classes when I don't need the other stuff, or avoiding case classes because I don't want, e.g., copy (because I need input validation, for instance, and the signature needs to be `Either[Oops, MyClass]`) and then writing that stuff by hand in the obvious way.  Seems like fine-graining could solve additional problems while also solving this one.)

Or do you mean, "even though `extends SomeTrait` would be better, I am inclined to recommend supporting the keyword version"?

raulraja:
@Ichoran I meant that I think we should provide the feature proposed, regardless of keyword or trait-based approach, in Scala. I think the feature is desired and we should find the best encoding. I agree with @bjornregnell on
> In summary: My recommendation is to vote for extends SomeGoodMarkerTraitName + synthesis of Product and Equality stuff, rather than a new kind of class.

Extending a trait and providing synthesized or extended members is easier to grasp than adding a new type of class or keyword modifier. Extending a trait seems transparent to the user as if the available and new members are brought into scope by the trait, whether these are autogenerated by the presence of the trait or implemented on the trait is something as a user I'd not be concerned about.

 I think the keyword-based approach is more rigid and does not scale as well as what @bjornregnell proposes in https://github.com/scala/improvement-proposals/pull/50#issuecomment-1341384085
 Selecting what you want to be generated would be great to have.

odersky:
Have we evaluated to what degree this could be done with macros? Something like

```scala
class C(name: String, age: Int):
  def toString = Structural.toString(this)
  def equals(other: C) = Structural.equals(this, other)
  def hashCode = Structural.hashCode(this)
  def withName(x: String) = Structural.copy(this, 0, x)
  def withAge(x: Int) = Structural.copy(this, 1, x)
```
`Structural` would be an object providing `toString`, `hashCode, `equals`, and `copy` as macros.

If that's considered too boilerplatey, one could also envision a magic extends of `Product` that auto-implements `toString`, `hashCode, `equals`.

Somehow, I feel the solution to all this should be more along these lines than creating a new kind of class.

julienrf:
I would rather use the workaround mentioned in the motivation section:

~~~ scala
case class Person private (name: String, age: Int):
  def withName(newName: String): Person = copy(name = newName)
  def withAge(newAge: Int): Person = copy(age = newAge)

object Person:
  private def unapply(person: Person): Person = person
~~~

bjornregnell:
@odersky Does that mean that no change is needed for this or is there some changed behavior of the compiler assumed by your last proposal to make the generated code of the updated with "delta" binary compatible?

odersky:
@bjornregnell That works out of the box. No compiler change is needed.

bjornregnell:
@odersky Ok great! 

I guess there is still the problem that you might start off with a "normal" case class and then realize that it should be evolve-able in a bincompat fashion, so maybe there is still some value in introducing some kind of marker trait or something that can signal this intent? And perhaps then it would still be valuable to scrap some boilerplate by including some compiler magic? If it's worth it, the SIP proposal could then perhaps be transformed to something like "less boilerplate for binary compatible evolution of case classes" or similar.

sideeffffect:
@odersky 
> Here's the setup of an extensible `Person` class.

In Scala, we have the `@tailRec` annotation. It ensures, that the annotated method is compilable using tail call optimization. In an analogous manner, can we have `@compatible` which verifies that an annotated class doesn't contain members which pose issues for backward compatibility when further evolving the class (adding more fields)?

smarter:
I've written up more on how I envision a "built-in codegen" approach working in https://contributors.scala-lang.org/t/scala-3-macro-annotations-and-code-generation/6035

julienrf:
Thank you everyone for your feedback on this proposal!

I do understand the concern of introducing yet another type of type definition that would be complicated to explain to beginners. I think the workaround mentioned in [this comment](https://github.com/scala/improvement-proposals/pull/50#issuecomment-1346210753) is acceptable, although not ideal. Therefore, I withdraw the proposal.

If someone else wants to revive the proposal, or add anything to the discussion, feel free to comment here.

julienrf:
I agree that the pattern should be well documented. Thank you @sideeffffect for going ahead with a PR!

sideeffffect:
So it turns out we have to do 3 things now:
 * add a new constructor, private to the companion object, with the original fields
 * add a custom `withAddress` method and
 * add an `apply` factory method to the companion.

This pattern doesn't seem as attractive as it did in the beginning... :confused: 
https://github.com/scala/docs.scala-lang/pull/2662#discussion_r1055563207

keynmol:
I was asked to post here an example of me trying to use a cross-compilable builder pattern (can't dataclass because I need Scala 3 as well).

The end result is very verbose, requires discipline, and from what I can tell isn't even going far enough to maintain compatibility (I didn't apply all the suggestions from the referenced docs PR).

I would ask the SIP committee to reconsider this SIP or a version thereof.

image
image

bjornregnell:
Thanks @keynmol for this example "from the wild" illustrating the current boilerplate situation well. I will bring this up at the SIP meeting tomorrow (if time permits @julienrf )

julienrf:
Indeed, we don’t have a good solution when you need to cross-compile your code with Scala 2 and 3 (which is common when you write a library…).

But I am not sure even if we had a good proposal that it would be implemented in both Scala 2 and Scala 3?

SethTisue:
> Indeed, we don’t have a good solution when you need to cross-compile your code with Scala 2 and 3

This was discussed at the meeting today — apparently `-Xsource:3` might help here...?

julienrf:
Yes, I was checking that and indeed with `-Xsource:3` it is possible to write a solution that cross-compiles. However, I have seen here https://github.com/scala/bug/issues/11661 that `-Xsource` should not be used in production, so I am not sure we should recommend using it.

keynmol:
IMO there are two problems here:

1. The Scala 3 specific solution is verbose in its own right and absolutely not scalable beyond trivial number of fields
2. Even if possible to just use that solution, seems like it doesn't cross-compile without a flag that carries its own dangers?

That seems like a lose-lose situation, IMO.

julienrf:
> 1. The Scala 3 specific solution is verbose in its own right and absolutely not scalable beyond trivial number of fields

Could you please elaborate on that point?

smarter:
> However, I have read here https://github.com/scala/bug/issues/11661 that -Xsource should not be used in production

-Xsource:3 has improved quite a bit since this issue was posted, so this could be revisited.

bjornregnell:
@smarter How far can the coming annotation macros in Scala 3 take us with this (also when cross-compiling to Scala 2)?

smarter:
- With macro definitions as they exist in both Scala 2 and Scala 3, one can avoid having to write down the body of methods to reduce the boilerplate:
 ```scala
class Foo(n: Int, s: String) {
  def withN(n: Int) = data.generated()
  def withS(s: String) = data.generated()

  override def equals(x: Any): Boolean = data.generated()
  override def hashCode: Int = data.generated()
}
  ```
- With macro annotations as they already exist in Scala 2 and will exist in Scala 3, one can *check* that the boilerplate we write is correct:
  ```scala
@data class Foo(n: Int, s: String) { // error: missing method `def withS(s: String) = data.generated()`
  def withN(n: String) = data.generated() // error: method parameter should have type `Int`
  ...
}
  ```
- In https://contributors.scala-lang.org/t/scala-3-macro-annotations-and-code-generation/6035 I further proposed reusing the rewrite mechanism present in Scala 3 to semi-automate the boilerplate generation.

I've prototyped this concept in https://github.com/lampepfl/dotty/pull/16545

SethTisue:
> -Xsource:3 has improved quite a bit since this issue was posted, so this could be revisited.

Agree, and I've added  a note to https://github.com/scala/bug/issues/11661 to say so.

PR #53, opened 2022-11-29
Fix and simplify grammar in SIP - 47

Sporarum:
As noted in https://github.com/scala/improvement-proposals/pull/47, the grammar did not correspond to the rest of the proposal, this PR solves that

### DIFF
--- a/content/clause-interleaving.md
+++ b/content/clause-interleaving.md
@@ -103,14 +103,13 @@ It provides the best of all worlds:
 
 ### Specification
 We amend the syntax of def parameter clauses as follows:
-
 ~~~
 DefDcl                 ::=  DefSig ‘:’ Type
 DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
-DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
-DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
-DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
-TermOrUsingParamClause ::=  DefTermParamClause
+DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]    -- and two DefTypeParamClause cannot be adjacent
+DefParamClauses        ::=  DefParamClause { DefParamClause }
+DefParamClause         ::=  DefTypeParamClause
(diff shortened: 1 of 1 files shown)


Sporarum:
@julienrf What is the process for addendum PRs ?
In particular, who should review this PR, and when can it be merged ?

julienrf:
Sorry @Sporarum, I somehow missed your PR. I’ve just asked the original team of reviewers to review it.

content/clause-interleaving.md
```diff
@@ -103,14 +103,13 @@ It provides the best of all worlds:
...
-
 ~~~
 DefDcl                 ::=  DefSig ‘:’ Type
 DefDef                 ::=  DefSig [‘:’ Type] ‘=’ Expr
-DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
-DefParamClauses        ::=  DefParamClauseChunk {DefParamClauseChunk}
-DefParamClauseChunk    ::=  [DefTypeParamClause] TermOrUsingParamClause {TermOrUsingParamClause}
-TermOrUsingParamClause ::=  DefTermParamClause
+DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]    -- and two DefTypeParamClause cannot be adjacent
+DefParamClauses        ::=  DefParamClause { DefParamClause }
```
chrisandrews-ms, line 110: I think it would be clearer if the comment was on this line rather than the line above, but I'm really nitpicking here 😄

Sporarum: I see what you mean, I guess it mostly depends on how we want to use `DefParamClauses` in other places

I also assumed people would be looking for it  first at the `DefSig` level, so it might be faster if the information is already here

Ultimately, I do not have a strong opinion either way, so maybe the other reviewers can comment on this ?

julienrf: I agree with @chrisandrews-ms. If @Kordyjan or @soronpo have a different opinion, please let us know!

Sporarum, line 110:
```suggestion
DefSig                 ::=  id [DefParamClauses] [DefImplicitClause]
DefParamClauses        ::=  DefParamClause { DefParamClause }    -- and two DefTypeParamClause cannot be adjacent
```
Suggestion from @chrisandrews-ms
(I couldn't put it in the relevant conversation because of the way `suggestion` works)

PR #55, opened 2022-12-30
SIP-52 - Wildcard context bounds

adampauls:
This SIP proposes new syntactic sugar for context bounds in the case where type context-bounded type argument need not be named. For example, 

```scala
// new sugar from this SIP
def showAll(xs: Set[? : Showable]): Unit
// existing sugar
def showAll[A: Showable](xs: Set[A]): Unit
// unsugared
def showAll[A](xs: Set[A])(using Showable[A]): Unit
```


julienrf:
Hello @bjornregnell, @soronpo and @dragos, could you please review this proposal?

soronpo:
Why doesn't this SIP have a Pre-SIP discussion thread on contributors?

sjrd:
Here is the Pre-SIP discussion: https://contributors.scala-lang.org/t/pre-sip-additional-syntactic-sugar-for-typeclasses/5675

soronpo:
Oh, my mistake, I thought this was a different proposal from the title that just related. They align well enough to no warrant a different thread.

soronpo:
OK, here are my thoughts on the proposal:
1. I think the `? : Showable` syntax is a deal breaker for me. I support @sjrd's suggestion of `using Showable` (I could be convinced that `impl` or `given` are better fit). My reasons are that it's less convoluted, easier to search for, and still fits what view as the motivation; it's not about saving characters, it's about saving the mental burden of naming things.
2. I think the proposal should address overloading such a definition in its desugared form.
3. I think the desugaring should be use a different type parameter block to resolve the problem mentioned in the open question.
```
def foo[T](x: using Showable): T // desugars to def foo[T][E$1: Showable](x: E$1): T 
foo[Int]("x") // compiles
def foo2[T](x: using Showable, y: using Showable): T // desugars to def foo2[T][E$1: Showable, E$2: Showable](x: E$1, y: E$2): T 
foo2[Int]("x", "y") // compiles
```
This, of course, contradicts the limitation we applied on interleaving clauses #56, preventing adjacent type parameter blocks, but we can still enable the compiler to do this disugaring internally. However if we do this, than overloading is forced to be applied in disugared form only because the compiler would reject the adjacent alternative block.

bjornregnell:
I think this reads well: `def showAll[A](xs: Set[A])(using Showable[A]): Unit`
So I'm not really convinced that new syntax is needed. What is the argument against that the above is enough, where you already can skip the name?  @adampauls

smarter:
> Do you similarly believe that ? <: Showable and ? >: Showable do not pull their weight?

The main difference is that `Foo[? <: Showable]` is an actual type, and while it is annoying to support, it turns out to be important for expressiveness (as briefly discussed in http://guillaume.martres.me/thesis.pdf, see reference to "use-site variance" on p. 90 and "wildcards" on p. 107).

By contrast, `Foo[? : Showable]` is syntactic sugar that can be used instead of a type for parameters of methods only. This raises various issues, for example, while I can write:
```scala
def showAll(xs: Set[? : Showable]): Unit =
  val y = xs
```
I cannot write down the type of `y` explicitly since it involves a compiler-generated type parameter whose name I don't know (and shouldn't be able to write in source code). This compiler-generated type parameter is also likely to leak in error messages, IDE hover information, etc.

This isn't completely unprecedented since repeated parameters (`xs: Int*`) are similarly awkward not-real-types, but this can be justified by the different use-site syntax they enable. Repeated parameters are also a concept that shows up in many programming languages, so the special syntax for them is less likely to confuse users. On the other hand, I often see people use the existing context bound syntactic sugar as an example of Scala having too many ways to do the same thing, so I'd be wary of introducing another syntax (things would be different if we didn't have the context bound syntax, but it's too late to change that now).

adampauls:
> The main difference is that Foo[? <: Showable] is an actual type

I think this is true in a technical sense, but not in an intuitive, practical sense. Consider:

```scala
trait Showable

class Box[T](val t: T)

def showAll(xs: Box[? <: Showable], ys: Box[? <: Showable]): Unit =
  var xst = xs.t
  val t1: Int = xst // Error: Found:    (xst : xs.T) Required: Int
  // But xs.T is lie of course
  val t2: xs.T = xst // Error: type T cannot be accessed as a member of Playground.Box[?1.CAP] from module class Playground$.

                             // where:    ?1 is an unknown value of type scala.runtime.TypeBox[Nothing, Playground.Showable]
  // I can't simply use `? <: Showable` as a type either
  val t2: ? <: Showable = xst // Error: Unbound wildcard type
  // Two occurences of ? <: Showable are not the same type
  xst = ys.t // Error: Found:    (ys.t : ys.T) Required: xs.T
```

As a Scala programmer, I'm used to thinking of each `?` instance as indicating an entirely new type variable, and unpredictable type widening aside, I don't think of `? <: Showable` as representing a "type", I think of it as representing a type variable with a bound. I think the analogy holds the same for the new proposed syntax.

That said, I do agree the fact that `? : Showable` is a leakier syntactic abstraction than `? <: Showable` is a knock against this proposal.

smarter:
> But xs.T is lie of course

Right, it's a type that cannot be written down by the user, it's here because it's useful for inference in some situations but it's true that it's awkward when it shows up, e.g. in error messages. I've been wondering if we should widen it when possible to make that less likely to happen (so in `var xst = xs.t`, `xst` would be inferred to have type `Showable` rather than `xs.T`).

adampauls:
> Right, it's a type that cannot be written down by the user,

Indeed, just pointing out that so is `? : Showable`, so this is problem is not new. As I said above, I agree it is a little bit leakier. 

>  I've been wondering if we should widen it when possible to make that less likely to happen (so in var xst = xs.t, xst would be inferred to have type Showable rather than xs.T)

A dangerous game of course. Suppose `Box` had a `set(t: T): Unit` method. If `var xst: xs.t` is widened to `Showable`, then `xs.set(xst)` would not compile right? I do think the mental model of `?` as an invisible type parameter whose name you don't know is pretty accurate of Scala already, and this proposal needn't change that?

smarter:
> I do think the mental model of ? as an invisible type parameter whose name you don't know is pretty accurate of Scala already, and this proposal needn't change that? 

If I have an `Array[Array[?]]` then every inner Array can be of a different type, on the other hand if I understand this proposal correctly, then `Array[Array[? : Showable]]` would have every inner Array share the same type parameter, so that's a potential source of confusion.
This is more likely to show up with something like 
```scala
def foo(xs: Box[? : Showable]*)
```
where it's natural to expect every argument to get its own Showable. As an aside, the fact that typeclasses and repeated parameters don't compose was cited as a reason for keeping implicit conversions in the language in https://contributors.scala-lang.org/t/can-we-wean-scala-off-implicit-conversions/4388).

adampauls:
> If I have an Array[Array[?]] then every inner Array can be of a different type

Checkmate, that's an unacceptable confusion. Thanks, I should I have seen that. 

`x: using Showable` seems to confusable with `using x: Showable` to me, but if other folks aren't concerned, I can modify the proposal.

adampauls:
I have decided that I think the `x: using Showable` syntax is not something I want to spend time championing, so I will close this for now.

bjornregnell:
Thanks anyway for your efforts on this SIP @adampauls 
Perhaps we find new ideas and ways forward later.

odersky:
I thought a bit more about it. I believe a more powerful extension would be to emulate something like Rust's "impl trait". The difference is that in `Array[Array[? : T]] ` indeed every inner array element can have a different implementation of `T`. This gives us a clear expressiveness win, which cannot easily be encoded now. 

There's a link with the discussion about implicit conversions. One advantage cited for conversions over typeclasses was that only conversions allow multiple instances for vararg arguments. With an impl trait like feature, typeclasses could support that as well.

But it's also much harder to implement. We can come back to it at some point in the future.

bjornregnell:
@smarter @odersky Just to settle my mind on this: Would it possible | not too problematic to allow that:
```scala
$ scala
Welcome to Scala 3.2.1 (17.0.4.1, Java OpenJDK 64-Bit Server VM).
                                                                                
scala> def add(x: Int, xs: Set) = xs + x
-- [E056] Syntax Error: --------------------------------------------------------
1 |def add(x: Int, xs: Set) = xs + x
  |                    ^^^
  |                    Missing type parameter for Set
1 error found
```
...would be expanded by the compiler to:
```scala
scala> def add(x: Int, xs: Set[Int]) = xs + x
def add(x: Int, xs: Set[Int]): Set[Int]
```
That would make simple things more concise, and also make it easier to talk about collections to a beginner without going too much into all the generic stuff.
(I see beginner students often write similar things as above when taking their first steps so there is something intuitive about `def add(x: Int, xs: Set) = xs + x`

sjrd:
@bjornregnell That's not possible. It would require inferring parameter types out of the method body. Scala does not do that. Its local type inference premise is such that it cannot ever do that.

bjornregnell:
Aha thanks. Sad though, as my human brain thinks this looks easy enough. Now with ChatGPT and all I just want to tell the compiler "do what I mean" and I get the most reasonable thing. 

But maybe the fancy future AI embedded in the compiler of Scala 4 should start working on hints first to see where that AI will take us :)

bjornregnell:
with some limited fantasy:
```scala
Welcome to Scala 4.0.0-RC1 (21.0.4.1, Java OpenJDK 64-Bit Server VM).
                                                                                
scala> def add(x: Int, xs: Set) = xs + x
-- [E056] Syntax Error: --------------------------------------------------------
1 |def add(x: Int, xs: Set) = xs + x
  |                    ^^^
  |                    Missing type parameter for Set
  |-- Compiler suggestion -- Did you mean
  |     def add(x: Int, xs: Set[Int]): Set[Int] = ...
1 error found
```

smarter:
> I believe a more powerful extension would be to emulate something like Rust's "impl trait". The difference is that in Array[? : T] every array element can have a different implementation of T. This gives us a clear expressiveness win, which cannot easily be encoded now.

For the record, I don't think that's how `impl Trait` works in Rust, the [spec](https://doc.rust-lang.org/reference/types/impl-trait.html) says:
> impl Trait in argument position is syntactic sugar for a generic type parameter like <T: Trait>

I believe heterogeneousness can still be achieved using [trait objects](https://doc.rust-lang.org/reference/types/trait-object.html) which are comparable to [existential types](https://wiki.haskell.org/Existential_type#Examples) in Haskell.

odersky:
@smarter Ah, thanks for clarifying. I must have confused the two.

PR #58, opened 2023-02-28
Add Binary APIs SIP

nicolasstucki:


content/binary-api.md
```diff
@@ -0,0 +1,211 @@
...
+When an inline method accesses a private/protected definition that is defined outside of its class, we generate an inline in the class of the inline method. This implies that accessors might be duplicated if a private/protected definition is accessed from different classes.
+
+### Removing deprecated APIs
+
+There is no precise mechanism to remove a deprecated method from a library without causing binary incompatibilities. We should have a straightforward way to indicate that a method is no longer publicly available but still available in the generated code for binary compatibility.
+
+```diff
+- @deprecated(...) def myOldAPI: T = ...
++ private[C] def myOldAPI: T = ...
+```
```
armanbilge, line 45: I believe this ties-in well with the discussion in https://github.com/lightbend/mima/discussions/724. To summarize:
1. MiMa currently ignores changes to `private[package]` APIs, because the assumption is that these are internal to the library. This improves the signal-to-noise when making such internal changes.
2. As noted here, `package[private]` is often used for removing a public API without breaking bincompat. But, due to (1), MiMa will no longer check these now-package-private APIs for binary-compatibility, which is dangerous. The workaround is to check MiMa against _every_ previous artifact, but this has its own costs/risks.
3. By adding an explicit annotation to denote public APIs, MiMa would be able to distinguish between internal APIs and once-public-now-removed APIs that must otherwise maintain binary-compatibility.

So, I hope that integration with MiMa will be considered as part of this proposal :)

Big 👍

content/binary-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+The introduction of the `@binaryAPI` and `@binaryAPIAccessor` do not introduce any binary incompatibility.
+
+Using references to `@binaryAPI` and `@binaryAPIAccessor` in inline code can cause binary incompatibilities. These incompatibilities are equivalent to the ones that can occur due to the unsoundness we want to fix. When migrating to binary APIs, the compiler will show the implementation of accessors that the users need to add to keep binary compatibility with pre-binaryAPI code.
+
+A definition can be both `@binaryAPI` and `@binaryAPIAccessor`. This would be used to indicate that the definition used to be private, but now we want to publish it as public. The definition would become public, and the accessor would be generated for binary compatibility.
+
+### Other concerns
+
+* Tools that analyze inlined TASTy code might need to know about `@binaryAPI`. For example TASTy MiMa.
```
armanbilge, line 178: And MiMa "classic" :)

https://github.com/lightbend/mima/

content/binary-api.md
sjrd: Yes, this proposal was partially designed to address that MiMa problem.

content/binary-api.md
nicolasstucki: I added it to the list of examples.

content/binary-api.md
nicolasstucki: I added the link https://github.com/lightbend/mima/discussions/724 to the SIP document.

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+Ideally, private definitions should have a maximum of one inline accessor, which is not the case now.
+When an inline method accesses a private/protected definition that is defined outside of its class, we generate an inline in the class of the inline method. This implies that accessors might be duplicated if a private/protected definition is accessed from different classes.
+
+### Removing deprecated APIs
+
+There is no precise mechanism to remove a deprecated method from a library without causing binary incompatibilities. We should have a straightforward way to indicate that a method is no longer publicly available but still available in the generated code for binary compatibility.
+
+```diff
+- @deprecated(...) def myOldAPI: T = ...
++ private[C] def myOldAPI: T = ...
```
smarter, line 44: Should this be `private[pkg]` ? The behavior of `private[C]` is to generate a name-mangled definition `C$$myOldAPI` to avoid clashes in case a subclass defines its own `myOldAPI`.

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+
+This proposal introduces 2 the `@binaryAPI` and `@binaryAPIAccessor` annotations, and changes adds a migration path to inline methods.
+
+#### `@binaryAPI` annotation
+
+A binary API is a definition that is annotated with `@binaryAPI` or overrides a definition annotated with `@binaryAPI`.
+This annotation can be placed on `def`, `val`, `lazy val`, `var`, `object`, and `given` definitions.
+A binary API will be publicly available in the bytecode.
+
+This annotation cannot be used on `private`/`private[this]` definitions.
```
smarter, line 61: If this is because those get mangled, then it should also be disallowed on `private[C]` definitions.

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+Example:
+
+~~~ scala
+class C {
+  @binaryAPI private[C] def packagePrivateAPI: Int = ...
+  @binaryAPI protected def protectedAPI: Int = ...
+  @binaryAPI def publicAPI: Int = ... // warn: `@binaryAPI` has no effect on public definitions
+}
+~~~
+will generate the following bytecode signatures
```
smarter, line 74: Since we're talking about bytecode it would be good to actually specify what we're generating at the bytecode level exactly. I assume that `public` in the code example below maps to `ACC_PUBLIC` in https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html, but we might also want to use `ACC_SYNTHETIC` so that these definitions stay hidden from javac and java IDEs.

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+}
+~~~
+
+#### `@binaryAPIAccessor` annotation
+
+A binary API with accessor is a definition that is annotated with `@binaryAPIAccessor`.
+This annotation can be placed on `def`, `val`, `lazy val`, `var`, `object`, and `given` definitions.
+The annotated definition will get a public accessor.
+
+This can be used to access `private`/`private[this]` definitions within inline definitions.
```
smarter, line 90: What about accesses to inner private classes?

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+~~~ scala
+class C {
+  @binaryAPI protected def a: Int = ...
+  protected def b: Int = ...
+  final def C$inline$b: Int = ...
+  inline def foo: Int = a + C$inline$b
+}
+~~~
+
+Note that if the inlined member is `a` would be private, we would generate the accessor `C$inline$a`, which happens to be binary compatible with the automatically generated one.
```
smarter, line 135: I don't understand this sentence (also I think there is a typo: "is `a`" should be "`a`"?), could an example be provided?

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+This is only a tiny mitigation of binary compatibility issues compared with all the different ways accessors can be generated.
```
smarter, line 136: Could you elaborate on this?

content/binary-api.md
```diff
@@ -0,0 +1,212 @@
...
+
+* Only valid on `def`, `val`, `lazy val`, `var`, `object`, and `given`.
+* An public accessor will be generated for the annotated definition. This accessor will be named `<fullClassName>$$inline$<definitionName>`.
+
+#### Inline
+
+* Inlining will not require the generation of an inline accessor for binary APIs.
+* Inlining will not require the generation of a new inline accessor, it will use the binary API accessors.
+* The user will be warned if a new inline accessor is automatically generated.
+  The message will suggest `@binaryAPI` or `@binaryAPIAccessor` and how to fix potential incompatibilities.
```
smarter, line 166: Will the message suggest using either of them? It seems we should always suggest using the latter, because it would break the ABI of existing code to add the former.

content/binary-api.md
smarter: Same question for primary and secondary private constructors.

content/binary-api.md
nicolasstucki: Not sure in which cases `private[C]`  generates a mangled name. I tried the following and non of the names get mangled.
```scala
class C:
  private[C] def f = 1
object C:
  private[C] def f = 1
trait B:
  private[B] def f = 1
class A extends B
```

nicolasstucki: I do get the mangling for `private[this] val` in traits.

content/binary-api.md
nicolasstucki: Not sure if adding `ACC_SYNTHETIC` is always a good idea.

* `@binaryAPI def publicAPI: Int = ... ` should not be marked as `ACC_SYNTHETIC` because it is truly public.
* Would `@binaryAPI protected  def publicAPI: Int = ... ` overridable in JAVA if it has `ACC_SYNTHETIC`?

content/binary-api.md
nicolasstucki: > What about access to inner private classes?

We currently do not allow them in inline methods. Not sure if we could use `@binaryAPI` to make the inner class public in the bytecode. I will investigate this possibility.

> Same question for primary and secondary private constructors.

Same for the constructors. We would require `@binaryAPI` and not `@binaryAPIAccessor` in this case as well.

content/binary-api.md
nicolasstucki: Both can break the binary API. 

* `@binaryAPI` migration will always break binary compatibility.
* Only if the is `private`/`private[this]` need `@binaryAPIAccessor`. These can break binary compatibility if the class is final.

nicolasstucki: Code like this
```scala
class A:
  private[foo] val foo: Int = 1
  inline def inlined = foo // error
```
would show an error message like this
```scala
-- Error: inline-unstable-accessors.scala:3:4 ------------------------------------
10 |    valBinaryAPI + // error
   |    ^^^^^^^^^^^^^
   |    Generated unstable inline accessor for value foo defined in class A.
   |
   |    Annotate foo with `@binaryAPI` to make it accessible.
   |
   |    Adding @binaryAPI may break binary compatibility if a previous version of this
   |    library was compiled with Scala 3.0-3.3, Binary compatibility should be checked
   |    using MiMa. To keep binary you can add the following accessor to class A:
   |      @binaryAPI private[A] def foo$A$$inline$foo: Int = this.foo
```

More examples can be found in https://github.com/lampepfl/dotty/pull/16992/files#diff-d12e949663e52759a532bf54a827260b0513390bfcdebddf5ff1c1bd04919aaa

content/binary-api.md
smarter: > I tried the following and non of the names get mangled.

Ah, it looks like the mangling only happens in Scala 2, not Scala 3. So that's still an issue for cross-compiling codebases.

content/binary-api.md
smarter: > Would @binaryAPI protected  def publicAPI: Int = ... overridable in JAVA if it has ACC_SYNTHETIC?

Yeah if you want the method to be overridable then it can't be ACC_SYNTHETIC.

content/binary-api.md
nicolasstucki: [19165b1](https://github.com/scala/improvement-proposals/pull/58/commits/19165b1b113f01368eaac6a236d8c95773423e3a) adresses the question on constructors. https://github.com/lampepfl/dotty/pull/16992 Has been updated to reflect this addition.

content/binary-api.md
nicolasstucki: Updated in the doc

nicolasstucki: Updated the example

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+  In a future version, these will become an error.
+
+### Compatibility
+
+The introduction of the `@binaryAPI` and `@binaryAPIAccessor` do not introduce any binary incompatibility.
+
+Using references to `@binaryAPI` and `@binaryAPIAccessor` in inline code can cause binary incompatibilities. These incompatibilities are equivalent to the ones that can occur due to the unsoundness we want to fix. When migrating to binary APIs, the compiler will show the implementation of accessors that the users need to add to keep binary compatibility with pre-binaryAPI code.
+
+A definition can be both `@binaryAPI` and `@binaryAPIAccessor`. This would be used to indicate that the definition used to be private, but now we want to publish it as public. The definition would become public, and the accessor would be generated for binary compatibility.
```
lrytz, line 190: > the compiler will show the implementation of accessors that the users need to add to keep binary compatibility

...

> the accessor would be generated for binary compatibility


IIUC, currently accessors are generated next to the inline method, whereas `@binaryAPIAccessor` wil generate an accessor next to the annotated private method, so that migration cannot be made in a binary compatbile manner - rigtht?

nicolasstucki: > currently accessors are generated next to the inline method

This is the case if the inline method is in a different class from the accessed member.

On the other hand, if the inline method is in the class the accessor will be placed in the class that defines the accessed members, similar to `@binaryAPIAccessor`. In some cases, these accessors are the same and we avoid binary incompatibilities in the migration from auto-generated to `@binaryAPIAccessor` (example 1).

> @binaryAPIAccessor wil generate an accessor next to the annotated private method, so that migration cannot be made in a binary compatbile manner?

In general, the migration can't be made binary-compatible. We chose to reuse the accessor name `<fullClassName>$$inline$<definitionName>` to be binary compatible with some auto-generated accessors. The accessors that are binary compatible are the ones in a non-final class that are generated by an inline method in that class (example 1).

#### Example 1 (compatible):
```scala
class A(x: Int):
  inline def y = x + 1 // inline def y = A$$inline$x + 1
  // def A$$inline$x = x // auto-generated
```
migrates to
```scala
class A(@binaryAPIAccessor x: Int):
  inline def y = x + 1 // inline def y = A$$inline$x + 1
  // def A$$inline$x = x // generated by @binaryAPIAccessor
```

#### Example 2 (incompatible):
```scala
final class A(x: Int):
  inline def y = x + 1 // inline def y = inline$x + 1
  // def inline$x = x // auto-generated
```
migrates to
```scala
class A(@binaryAPIAccessor x: Int):
  inline def y = x + 1 // inline def y = A$$inline$x + 1
  def inline$x = x // user needs to add this. Migration warning shows tells the users shows this code
  // def A$$inline$x = x // generated by @binaryAPIAccessor
```

#### Example 3 (incompatible):
```scala
package p
final class A(private[p] x: Int)
```
```scala
package p
inline def f(a: A): Int = a.x // inline def f(a: A): Int = A$$inline$x(a) 
// def A$$inline$x(a: A) = a.x // auto-generated
```
migrates to
```scala
package p
final class A(@binaryAPI private[p] x: Int)
```
```scala
package p
inline def f(a: A): Int = a.x // inline def f(a: A): Int = a.x
def A$$inline$x(a: A) = a.x // user needs to add this. Migration warning shows tells the users shows this code
```

lrytz: > In general, the migration can't be made binary-compatible

In that case, it's maybe not worth introducing both `@binaryAPI` and `@binaryAPIAccessor`? Is it always possible to start using `@binaryAPI` and maintain binary compatibility by manually adding the accessors that were previously generated by the compiler?

nicolasstucki: The issue with that approach is that users would need to change all their `private/private[this]` members that are accessed from inline methods `protected/private[Cls]/pribate[pack]`. This can cause source incompatibilities.

lrytz: I don't understand, I'm sorry. Where am I wrong?
  - Currently, if a `private` method is accessed in an `inline def`, the compiler generates an accessor to the private method
  - The user can annotate the method `@binaryAPI` (but keep it private in source), which makes it public in bytecode, and the accessor is no longer generated
  - To maintain binary compatibility, the user can write the method that was previously generated into the source code

nicolasstucki: > The user can annotate the method @binaryAPI (but keep it private in source), which makes it public in bytecode, and the accessor is no longer generated

[`@binaryAPI` cannot be placed on private/private[this] APIs](https://github.com/scala/improvement-proposals/pull/58/files#diff-fe3d0b5bf9913b4901dab94b4dc8046f1abb7a37eb3884c783125bda9f3f868dR65). This is because that change could be a source or binary incompatible change. For example:
```scala
class A:
  @binaryAPI private val x: Int = 1 // should not be allowed
  def f: Int = x
class B extends A:
  val x: Int = 2
```
If we make `x` public in the bytecode, we would accidentally make `B.x` override `A.x` which would change the semantics of a private definition. Or we would have to flag them as overrides of private members, which implies that we would have to rename anything that would clash with a `@binaryAPI private`. In this case we would not be able to use the name `x` in `B` anymore.

lrytz: Ah, right, thanks. Was name-mangling considered?

"Only add `@binaryAPIAccessor`" could also be an option. The concern

> The drawback is that we would add code size and runtime overhead to all uses of this feature

is maybe minor? Accessors are very common in Scala, why would these ones be problematic?

nicolasstucki: > "Only add @binaryAPIAccessor" could also be an option. 

If we do that we would not cover 2 of the motivations: _Removing deprecated APIs_ and _No way to inline reference to private constructors_. Those use cases could be supported later with the addition of `@binaryAPI`.

If we add `@binaryAPI` later there is also a migration concern. In a first step users would add `@binaryAPIAccessor` to all the definitions, but then when `@binaryAPI` they might also want to add that one to access the member directly. This would force them to have both `@binaryAPIAccessor` and `@binaryAPI`. If we add both at the same time, then they would be able to only use `@binaryAPI`.

> Ah, right, thanks. Was name-mangling considered?

I assume with `@binaryAPIAccessor` of private members. The issue is that we would need to mangle the names of private members but not public/protected members. This implies that if the user changes a private member into a public one the accessor would not be generated anymore. This is the exact same issue we have right now but triggered in a different way.

lrytz: Thanks for the details. I don't want to be annoying and keep discussing details here, the discussion should move elsewhere?

I think introducing two annotations is bad. Users will have to pick one or the other, they have to know how they both work and how using them affects binary compatibility.

nicolasstucki: I do not see a scenario where a user would need to think about which annotation to choose. In the scenario of _Removing deprecated APIs_ they can only use `@binaryAPI`. In all other scenarios with inlining, the users would be told which annotation they need to use.

nicolasstucki:
## Migration analysis

In the following table we can see how many `@binaryAPI`, `@binaryAPIAccessor`, and explicit accessors must be added to migrate. This list contains all project from the dotty community build and VirtusLab/community-build3.

| project | `@binaryAPI` | `@binaryAPIAccessor` | explicit accessor |
| ------- | ------------ | -------------------- | ----------------- |
| endpoints4s/endpoints4s ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8803805431)) | 1 | 0 | 1 |
| fingo/spata ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8807306452)) | 1 | 4 | 5 |
| getkyo/kyo ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8811045966)) | 1 | 2 | 2 |
| giiita/refuel ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8799581487)) | 5 | 0 | 5 |
| katrix/perspective ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8808774376)) | 0 | 11 | 11 |
| plokhotnyuk/jsoniter-scala ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8808776623)) | 0 | 1 | 1 |
| rssh/dotty-cps-async ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8799594630)) | 0 | 2 | 0 |
| scala-ts/scala-ts ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8799596565)) | 0 | 1 | 1 |
| scodec/scodec ([diff](https://github.com/dotty-staging/scodec/compare/96a77ecaaf913f195bb4079966a2e9fb41ce214e...706de4608c474b292907495f1896c4cb9c377bcc#diff-56e287498ade7b36062fe85610ae019cb3772cf8149d917da4017d8561cde1bc)) | 0 | 3 | 1 |
| vigoo/desert ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8812635815)) | 0 | 1 | 1 |
| virtuslab/avocado ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8805401190)) | 1 | 0 | 1 |
| yakivy/jam ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8803361407)) | 1 | 0 | 1 |
| zio/caliban-deriving ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8812309824)) | 1 | 0 | 1 |
| zio/prelude ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452/jobs/8804355619)) | 0 | 2 | 0 |
| | | | |
| total | 11 | 27 | 31 |

### Full output
* VirtusLab/community-build3 ([CI output](https://github.com/VirtusLab/community-build3/actions/runs/4925312452))

### Conclusion
Migration of most projects is relatively simple. katrix/perspective it the only project with a medium migration complexity. The complexity in all these projects comes from explicit accessors that users have to add in source (copied from migration message).

### Possible improvement
Some of the explicit accessors come from accessors in objects (not counted the exact number). These used an accessor named `inline$<definitionName>` (without the `<fullClassName>$$` prefix). It might be possible to special case `@binaryAPIAccessor` in object to generate the accessor with the same name as before. I will try to add this to the prototype and check how it affects the migration.

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+### No way to inline reference to private constructors
+
+It is currently impossible to refer to private constructors in inline methods. If users want to access one of those, they must write an accessor explicitly. This extra indirection is undesirable.
+
+## Proposed solution
+
+### High-level overview
+
+This proposal introduces 2 the `@binaryAPI` and `@binaryAPIAccessor` annotations, and changes adds a migration path to inline methods.
```
julienrf, line 57:
```suggestion
This proposal introduces 2 the `@binaryAPI` and `@binaryAPIAccessor` annotations, and adds a migration path to inline methods.
```

julienrf, line 57:
```suggestion
This proposal introduces 2 the `@binaryAPI` and `@binaryAPIAccessor` annotations, and adds a migration path to inline methods.
```

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+```diff
+- @deprecated(...) def myOldAPI: T = ...
++ private[C] def myOldAPI: T = ...
+```
+
+Related to discussion in https://github.com/lightbend/mima/discussions/724
+
+### No way to inline reference to private constructors
+
+It is currently impossible to refer to private constructors in inline methods. If users want to access one of those, they must write an accessor explicitly. This extra indirection is undesirable.
```
julienrf, line 51: Would it be possible to have a short example that illustrates the problem?

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+### Removing deprecated APIs
+
+There is no precise mechanism to remove a deprecated method from a library without causing binary incompatibilities. We should have a straightforward way to indicate that a method is no longer publicly available but still available in the generated code for binary compatibility.
+
+```diff
+- @deprecated(...) def myOldAPI: T = ...
++ private[C] def myOldAPI: T = ...
+```
+
+Related to discussion in https://github.com/lightbend/mima/discussions/724
```
julienrf, line 47:
```suggestion
Related to discussion in [https://github.com/lightbend/mima/discussions/724](https://github.com/lightbend/mima/discussions/724).
```

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+## Motivation
+
+### Provide a sound way to refer to private members in inline definitions
+
+Currently, the compiler automatically generates accessors for references to private members in inline definitions. This scheme interacts poorly with binary compatibility. It causes the following three unsoundness in the system:
+* Changing any definition from private to public is a binary incompatible change
+* Changing the implementation of an inline definition can be a binary incompatible change
+* Removing final from a class is a binary incompatible change
+
+You can find more details in https://github.com/lampepfl/dotty/issues/16983
```
julienrf, line 31:
```suggestion
You can find more details in [https://github.com/lampepfl/dotty/issues/16983](https://github.com/lampepfl/dotty/issues/16983)
```

julienrf, line 31:
```suggestion
You can find more details in [https://github.com/lampepfl/dotty/issues/16983](https://github.com/lampepfl/dotty/issues/16983)
```

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+public class C {
+  public C();
+  public int packagePrivateAPI();
+  public int protectedAPI();
+  public int publicAPI();
+}
+~~~
+
+In the bytecode, `@binaryAPI` definitions will have the [ACC_PUBLIC](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.1-200-E.1) flag.
+<!-- We can also set the [ACC_SYNTHETIC](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.1-200-E.1) to hide these definitions from javac and java IDEs. -->
```
julienrf, line 89: Has this been studied further? Is there a conclusion?

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+final class binaryAPI extends scala.annotation.StaticAnnotation
+final class binaryAPIAccessor extends scala.annotation.StaticAnnotation
+```
+
+#### `@binaryAPI` annotation
+
+* Only valid on `def`, `val`, `lazy val`, `var`, `object`, and `given`.
+* TASTy will contain references to non-public definitions that are out of scope but `@binaryAPI`. TASTy already allows those references.
+* Annotated definition will be public in the generated bytecode. Definitions should be made public as early as possible in the compiler phases, as this can remove the need to create other accessors. It should be done after we check the accessibility of references.
```
julienrf, line 168:
```suggestion
* The annotated definitions will be public in the generated bytecode. Definitions should be made public as early as possible in the compiler phases, as this can remove the need to create other accessors. It should be done after we check the accessibility of references.
```

julienrf: About the second part “Definitions should be made public as early as possible in the compiler phases, as this can remove the need to create other accessors”: is that an implementation detail? Or is it part of the specification? If that’s part of the specification, could you please explain which accessor will avoid being generated?

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+#### `@binaryAPIAccessor` annotation
+
+* Only valid on `def`, `val`, `lazy val`, `var`, `object`, and `given`.
+* An public accessor will be generated for the annotated definition. This accessor will be named `<fullClassName>$$inline$<definitionName>`.
```
julienrf, line 174:
```suggestion
* A public accessor will be generated for the annotated definition. This accessor will be named `<fullClassName>$$inline$<definitionName>`.
```

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Only add `@binaryAPI`
+This would simplify the system and the user interaction with this feature. The drawback is that we could not access `private[this]` definitions in inline code. Users would need to use `private[C]` instead, which could cause name clashes.
+
+### Only add `@binaryAPIAccessor`
+This would simplify the system and the user interaction with this feature. The drawback is that we would add code size and runtime overhead to all uses of this feature. It would not solve the [Removing deprecated APIs](#removing-deprecated-apis) motivation.
+
+## Related work
+
+* Proof of concept: https://github.com/lampepfl/dotty/pull/16992
```
julienrf, line 224:
```suggestion
* Proof of concept: [https://github.com/lampepfl/dotty/pull/16992](https://github.com/lampepfl/dotty/pull/16992)
```

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+* Initial discussions: https://github.com/lampepfl/dotty/issues/16983
```
julienrf, line 225:
```suggestion
* Initial discussions: [https://github.com/lampepfl/dotty/issues/16983](https://github.com/lampepfl/dotty/issues/16983)
```

julienrf (commented):
I reviewed the format and style. Please keep in mind that the proposals are ultimately published to https://docs.scala-lang.org/sips so the markup has to be valid, and the content should be as clear as possible.

content/binary-api.md
nicolasstucki: I found that we currently need to do that for other accessors. I opened a discussion in https://github.com/lampepfl/dotty/issues/17461. `@binaryAPI` should have `ACC_SYNTHETIC` if and only if other accessors.

content/binary-api.md
```diff
@@ -0,0 +1,227 @@
...
+You can find more details in [https://github.com/lampepfl/dotty/issues/16983](https://github.com/lampepfl/dotty/issues/16983)
+
+### Avoid duplication of inline accessors
+
+Ideally, private definitions should have a maximum of one inline accessor, which is not the case now.
+When an inline method accesses a private/protected definition that is defined outside of its class, we generate an inline in the class of the inline method. This implies that accessors might be duplicated if a private/protected definition is accessed from different classes.
+
+### Removing deprecated APIs
+
+There is no precise mechanism to remove a deprecated method from a library without causing binary incompatibilities. We should have a straightforward way to indicate that a method is no longer publicly available but still available in the generated code for binary compatibility.
```
JD557, line 40: Is this an actual problem?

Based on the [Semantic Versioning spec](https://semver.org/#how-should-i-handle-deprecating-functionality) and a [discussion with clarifications](https://github.com/semver/semver/issues/712), my understanding is that the semver way would require a MAJOR version bump to remove a deprecated method.

Binary incompatibility of different major versions seems like a reasonable expectation.

There are obviously other version schemes other than semver, but I imagine the handling of deprecated APIs is similar (e.g. I think `EPOCH.MAJOR.MINOR` would still require a MAJOR bump)?

sjrd: Yes, this is an actual problem because the Scala ecosystem, and therefore use of SemVer, is rooted in binary and TASTy compatibility. You want to be able to make *source* breaking changes while retaining *binary* and *TASTy* compatibility, and that should stay a MINOR version bump. Perhaps you are mitigating the source breakage with extension methods or new overloads instead.

(If breaking source compat required a MAJOR version bump in Scala, every new version of a library with a new public class or method would be a MAJOR version, and we don't want that for obvious reasons.)

content/binary-api.md
nicolasstucki: I added an example.

julienrf:
Dear @smarter, @lrytz, and @raulraja, would you have time to review this PR this week? It would be great to converge to a recommendation before the next SIP meeting.

lrytz:
[IIUC - please correct me].

Accessing non-public members from inline methods currently causes syntehtic accessors. The goal of this proposal is to require user input (an annotation) in this case to ensure the binary API is stable.

I think introducing two annotations is not great, ideally there should be one way for users to address the problem. If `@binaryAPIAccessor` cannot solve all the issues / cases, maybe `@binaryAPI` is the way to go. It seems to me that any accessor that is currently compiler-generated (or that would be generated by `@binaryAPIAccessor`) can be written manually.

nicolasstucki:
> Accessing non-public members from inline methods currently causes syntehtic accessors. The goal of this proposal is to require user input (an annotation) in this case to ensure the binary API is stable.

Yes, the main goal is to ensure the binary API is stable with secondary goals such as supporting private constructors and Removing deprecated APIs (both covered by `@binaryAPI`). We also need to make sure that every project can migrate to the new scheme without breaking binary compatibility (covered by both `@binaryAPI` and `@binaryAPIAccessor`).

> ... It seems to me that any accessor that is currently compiler-generated (or that would be generated by `@binaryAPIAccessor`) can be written manually.

Using `@binaryAPI` we can add any of the accessors from the previous scheme using the correct mangled names (similar to removing deprecated APIs). Unfortunately, it is impossible to add explicit accessors for private fields using `@binaryAPI` without breaking source compatibility or leveraging undefined behavior of names with a `$` in them. Adding both annotations is the only option we found that does not introduce new issues (or push the same issues a bit further). 

It is also what will make the [migration](https://gist.github.com/nicolasstucki/003f7293941836b08a0d53dbcb913e3c) relatively simple for most projects. With both annotations, 33% of inline methods that access private members would be complex to port (in 5 projects of the open community build). With only `@binaryAPI`, we would only have a 80% of complex methods to port (in 11 projects, only one trivially ported). It would also be unclear what new projects would use when accessing private members (binary breaking change or unsound names).

lrytz:
What is a "complex migration"? Can the scala compiler emit instructions with a method template in a warning message, or directly add the method to the source code via `-rewrite`?

If adding both annotations, could `@binaryAPIAccessor` be marked `@deprecated` from the beginning?

lrytz:
According to [your survey](https://gist.github.com/nicolasstucki/003f7293941836b08a0d53dbcb913e3c), the number of cases where `@binaryAPIAccessor` can be used is low (34). IMO addressing these with explicit accessors is fine, especially if the compiler emits them (in the warning / as a quickfix / ...). It also makes things explicit, maintainers coming across a `@binaryAPIAccessor` don't need to go figure out what it means.

My suggestion to deprecate the annotation was to make it clear that the annotation is only intended to address backwards compatibility, the warning can be silenced with `@nowarn`.

bjornregnell:
I have very limited knowledge in this area and maybe my questions/concerns below are not relevant, but here they come anyway (I'd be happy if someone explains to me in simple terms if/why my questions/concerns might not be relevant):

1. Does this mean that library authors need to sprinkle annotations on all or many members of an API if they are to be sure that they can fulfill their goal of keep evolving the library in a binary compatible way (if combined with optimizations based on inline)? My concern is that this will give cluttered code with much less readability. Will this also clutter the scaladoc user-facing view of the api?
3. Would it be possible to have an annotation on class or package level once and for all for the whole api, so that the binary compatibility nitty gritty can be applied transitively instead so that the code of all the members can look less cluttered. Or maybe a language import or something less intrusive? I mean: could a smarter compiler fix this for me if I tell it to do so on a higher level?
4. Is it possible that the effect and inner workings of this proposal for binary compatibility could be automatically applied on all relevant members always in a smart way, so that the annotations are not needed? Even if there is a cost to general application in byte code size or call indirection, maybe its worth it? Maybe it can be opt-out instead of opt-in and call it reversely like @noBinCompat or something? I assume many library authors would like better binary compatibility out-of-the-box if it improves readability, depending on how they prioritize maintainability over performance or other goals. My concern is that the proposed annotations looks like a leaking compiler back-end implementation detail...

bjornregnell:
@nicolasstucki Perhaps you want to include parts of the clarifications made by @sjrd in some form in the introductory sections of the SIP text, to help readers better understand the (simplified) pragmatics of why and how to use these annotations and why they are really needed?

nicolasstucki:
> My suggestion to deprecate the annotation was to make it clear that the annotation is only intended to address backwards compatibility, the warning can be silenced with @nowarn.

It is not only intended for backward compatibility. New code that refers to private members in inline methods also needs `@binaryAPIAccessor`.

nicolasstucki:
@lrytz, taking a step back I figured out why our arguments are opposed. There are two distinct use cases and we are each arguing based on the requirements of one of them. The two use cases can be labeled as **Sound/stable private member accessibility** and **Migration from the old scheme**. You are mostly looking at the requirement of the second one while I am looking at the first one. The first use case happens to require `@binaryAPIAccessors` but benefits from `@binaryAPI`, while the second requires `@binaryAPI` but benefits from `@binaryAPIAccessors`. I will try to disentangle all these requirements below. Sorry for not being able to be clearer before.


### Sound/stable private member accessibility

In this use case, we could use `@binaryAPIAccessor` to make all private definitions accessible. 
```scala
class A:
  private def x = ??? // requires an accessor
  private[p] def y = ??? // requires an accessor or to be public in the Bytecode
  def z = ???
  inline def f = x + y + z
```
We can also benefit from `@binaryAPI` on `y` to avoid the generation of an extra accessor. We cannot use `@binaryAPI` to make `x` accessible without breaking compatibility.

The addition of `@binaryAPI` also solves the limitation we have on calling private constructors. This use case cannot be encoded with `@binaryAPIAccessor`.

Therefore we require `@binaryAPIAccrssor` and can benefit from the addition of `@binaryAPI`.

### Migration from the old scheme

In this use case, on top of the requirements of the previous section, we also need to provide a sound migration path libraries compiled with automatically generated accessors.
```scala
class A:
  private def x = ???
  // def A$inline$x = x // compiler generated
  private[p] def y = ???
  // def A$inline$y = y // compiler generated
  def z = ???
  inline def f = x + y + z
```
To keep binary compatibility once the accessors are not generated anymore, the libraries will need to define them explicitly (`A$inline$x` and `A$inline$y` in the example). These methods should be public in the bytecode, but should not be public in the source/pickles to avoid users calling those directly and effectively leaking private members. This is where `@binaryAPI` can be used to define the accessor (package) private in the scope of the class to maintain its privateness in the public API. Therefore to avoid breaking the libraries we require `@binaryAPI`.

Note that it is safe to use `@binaryAPI` in this use case only because we are adding a method that has a name that is guaranteed not to clash with user-defined methods in the bytecode. This is because we are using the reserved names `...$inline$...`. This is only acceptable in the case of a (dirty but sound) patch for migration.

We are choosing the names of the accessor of `@binaryAPIAccessor` to match as many of the ones that were automatically generated. In the cases where those names match we can benefit from the `@binaryAPIAccessor` in this use case to make the migration simpler (as shown in the [survey](https://gist.github.com/nicolasstucki/003f7293941836b08a0d53dbcb913e3c)).

### Conclusion

We need `@binaryAPIAccessor` to support all the use-cases inline method currently supports. We need `@binaryAPI` to be able to migrate to the new scheme.

The addition of `@binaryAPI` also enables new use cases such as accessing private constructors and avoiding unnecessary accessors. The addition of `@binaryAPIAccessor` allows for a smoother migration.

lrytz:
Thank you for clarifying everything. I'm only going to reply to points where I have something to add.

> We cannot use `@binaryAPI` to make `x` accessible without breaking compatibility.

IMO that's not a necessary use case. When writing a new inline method which accesses a private method `x`, there needs to be something public in bytecode. So we can tell the user to change `x` to be `private[p]` (and rename it if desired) and use the `@binaryAPI` annotation.

With that, the only remaining use case for `@binaryAPIAccessor` would be to make migration simpler in a few cases, which has very limited value.

nicolasstucki:
> When writing a new inline method which accesses a private method x, there needs to be something public in bytecode. So we can tell the user to change x to be private[p] (and rename it if desired) and use the @binaryAPI annotation.

There are a couple of issues with that approach. Changing a private definition to a private[p] definition is a binary-breaking change. Consider the following example:

```diff
- class A(x: T)
+ class A(@binaryAPI private[A] val x: T)
```
```scala
class B extends A:
  val x: T = 3
```
Changing `x` from private to public (protected, or package private) in the Bytecode breaks or changes the semantics of `B`. Accessing `A.x` could now erroneously return the value of `B.x`. Note that `B` may be in another project and the author of `A` would not notice the change. 

Renaming `x` to something else is also not possible because we do not know all the members that the children of `A` may define.

From the ergonomics point of view, we do not want to degrade the usability of inline methods. Accessing private members from inline methods is relatively common and we do not when the users to have to need to add unnecessary syntactic overhead to these private fields.

```diff
- class A(x: T)
+ class A(@binaryAPIAccessor x: T) // simpler and semantically meaningful 
```

lrytz:
Thanks @nicolasstucki. I think we both understand the pros and cons / each others points, so the next step is to discuss in the committee.

lrytz:
For those who want to catch up on the long discussion:
  - (Re-)read the SIP (skip any comments)
  - Read the last few comments on this PR, starting at https://github.com/scala/improvement-proposals/pull/58#issuecomment-1600657922
 
I think this sums everything up.

smarter:
The spec mentions  "Avoid duplication of inline accessors" as one justification for this change.
If I understand correctly, this refers to duplication in situations like this:
```scala
class A:
  protected def foo = 1

class B extends A:
  inline def bar1 = foo
  // def B$$inline$foo = this.foo

class C extends A:
  inline def bar2 = foo
  // def C$$inline$foo = this.foo
```
This proposal disallows this and instead forces the user to annotate the protected definition to make it inlinable in subclasses, but this does mean that if the author of `B` and `C` cannot change `A` (because it is defined by an external library), then he must add accessors manually (something which might be hard to figure out for users):
```scala
class B extends A:
  @binaryAPIAccessor
  private def myFoo = foo
  inline def bar1 = myFoo

class C extends A:
  @binaryAPIAccessor
  private def myFoo = foo
  inline def bar2 = myFoo
```
So effectively duplication can still happen.

nicolasstucki:
In this situation, the authors of `B` and `C` are accessing a non-private member of `A`. This implies that they can use `@binaryAPI` and avoid the accessors. 

```scala
class B extends A:
  @binaryAPI
  protected override def foo = super.foo
  inline def bar1 = foo

class C extends A:
  @binaryAPI
  protected override def foo = super.foo
  inline def bar2 = foo
```

In this scenario, it is `B` and `C` that are guaranteeing the existence of a publically accessible. These could also become just `override def foo = super.foo`.

We could try to detect those cases and suggest the alternative of overriding the definition instead of adding the annotation to the original definition.

Obviously, if `B` and `C` are adding `@binaryAPI` to `foo` they would consider asking the authors of `A` if it is possible to add `@binaryAPI` directly to their definition.

smarter:
I briefly presented this proposal at the SIP meeting last week, some points that were mentioned:
- People would generally prefer one over two annotations
- @odersky suggested having the warnings be hidden under a linting option since ABI compatibility it is not something that application developers need to care about, only library developers in a binary ecosystem. The annotation would then only be used to suppress the linting warning, not to influence code generation.

nicolasstucki:
I updated the SIP to have a single annotation and warn under a linter option.

Note that I moved the prototype implementation to https://github.com/lampepfl/dotty/pull/18402 and kept the old one it the old PR.

nicolasstucki:
Thumbs up here if you prefer `@binaryAPI`

nicolasstucki:
Thumbs up here if you prefer `@binaryPublic`

nicolasstucki:
Thumbs up here if you prefer `@publicInBinary`

julienrf:
May I suggest `@publicABI`?

bjornregnell:
> May I suggest `@publicABI`?

I think an acronym risks being esoteric to most readers, esp. that acronym. Typically and especially, open source library code gets read by many that are not developing library code but using it. I would prefer a name that does not include an acronym. A "binary" is generally referring to our common knowledge of that those ones and zeros are in there somewhere under the hood...

bjornregnell:
A more complete wording would be `@publicInBinaryInterface` but that's maybe too long and I think the last word `Interface` can be inferred implicitly by human readers :)

smarter:
Kotlin uses [PublishedApi](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-published-api/)

bjornregnell:
But there might be some who wants to give their thumbs up for the more complete but lengthy `@publicInBinaryInterface" ?

odersky:
I thought we wanted to go the other way? 

 - Force the user to make the accessed field private[C] (which gets implemented as public)
 - Make sure in the backend the redundant accessors are eliminated in that case (or don't create them in the first place)
 -  have a linter option that makes sure that all accessed fields are of this form? 

Was that not the plan? 

Example: If I write `C` like this
```scala
class C:
  inline def foo = f
  private def f = 1
```
I get this binary API:
```java
public class C {
  public C();
  public final int C$$inline$f();
}
```
Which is a problem since making `f` public will suppress the accessor `C$$inline$f`. But if I write `C` like this:
```scala
class C:
  inline def foo = f
  private[C] def f = 1
```
The binary API becomes
```java
public class C {
  public C();
  public int f();
  public int C$$inline$f();
}
```
Now the inline accessor is redundant since `f` is already public. We could remove the inline accessor in this situation. 
Then all binary accesses will go through the public `f` so making `f` public in source will be binary compatible. 

This would mean we need no new annotation at all.

bjornregnell:
I get it that it is nice to avoid having an annotation at all, but it feels non-intuitive to having to say that it is private[C] when I want it to be *public* in the binary interface.

Stating something that includes the word `public` makes more sense to me. Let's say I know nothing about the strangeness of binary compatibility of API:s and encounter code with `private[C]` then I would think: what does this mean??? if it is private then of course it should be private to this C so why not just write private???

odersky:
The point is that current Scala with a small backend tweak is already capable to support this use case. Whereas an annotation has to be defined and explained. What I am envisioning is a linter option that would give a message like this:
```
  private def f = 1
  ^^^^^
   Warning: Making this method public in the future will not be binary compatible since it is currently accessed
   from inline method `foo` via an inline accessor. You can avoid the problem by declaring `f` as a `private[C]`.
```

sjrd:
Our tools, MiMa and tasty-mima, need to know. They must allow to remove a `private[C]` method, not *not* an `@binaryAPI private[C]` method.

The linter option is there to warn when using a `private[C]` method that is *not* annotated with `@binaryAPI`. Because that is a situation where MiMa/tasty-mima will allow to remove the method.

julienrf:
Can the annotation be defined in MiMa or tasty-mima?

bjornregnell:
> What I am envisioning is a linter option that would give a message like this

But that warning will not be seen by someone that just reads and tries to grok the open source code. So the intent is not clear when it just says `private[C] def f` compared to `@publicInBinary private def f`

nicolasstucki:
> should it be @publicInBinary private def f or @publicInBinary private[C] def f? I think it must be the latter, correct?

It should be the latter if we want to be consistent with overloading rules. `private` definitions can have the same name in subclasses without clashing. A `@publicInBinary` must be encoded a public method in the bytecode and is subject to naming restrictions.

Using `@publicInBinary private` would be nice from a syntax perspective but to state that this definition behaves semantically as a `private[C]`.

nicolasstucki:
This is a good example of why we should not allow `@publicInBinary private def`

```scala
class C:
  @publicInBinary private def f: Int = 1
  def g = f // must be C.f not D.f

class D extends C:
  def f: Int = 2
```

bjornregnell:
@nicolasstucki Also perhaps good to include a reference somewhere in the proposal text to the similar Kotlin annotation given by @smarter above

nicolasstucki:
@bjornregnell, I added coments/references to both of your comments.

nicolasstucki:
> Can the annotation be defined in MiMa or tasty-mima?

No. The compiler needs to know about this annotation when compiling inline methods. Therefore it must be defined in the standard library.

bjornregnell:
> @bjornregnell, I added coments/references to both of your comments.

Looks great to me! 

Just found this strange repetition typo: "but do not guarantee but do not have" that you might want to fix.

lrytz:
The committee was positive about this SIP in its current form when it was discussed on Sep 11 in Madrid. It will be brought up for voting in the next meeting.

I have a question about `-WunstableInlineAccessors`. I think there are two cases:

  1. A library with an existing `inline def` that accesses a private member. Here the author should create a binary copmatible accessor with the same name as was previously compiler-generated. The warning tells the user that accessor name. It was actually suggested that the compiler could insert the accessor to the source file using `-rewrite`.
  2. When someone writes a new `inline def` that accesses a private member, the user should get a warning. Here the warning should not suggest to create an accessor with a mangled name with `$` signs. Is this case also addressed?

lrytz:
This SIP was unanimously accepted in the meeting on Friday, congratulations @nicolasstucki!

As this SIP was never discussed on Discourse, could you open a thread saying that this SIP will be implemented as an experimental feature?

In terms of tooling affected by this SIP, MiMa (cc @dwijnand) and TASTy-MiMa (cc @sjrd) was mentioned in the meeting.

lrytz:
(Also, there was no reply to my [previous comment](https://github.com/scala/improvement-proposals/pull/58#issuecomment-1717909362) from September.)

nicolasstucki:
> As this SIP was never discussed on Discourse, could you open a thread saying that this SIP will be implemented as an experimental feature?

Added in https://contributors.scala-lang.org/t/updates-on-sip-52-publicinbinary-and-wunstableinlineaccessors/6363

nicolasstucki:
> I have a question about `-WunstableInlineAccessors`. I think there are two cases:
> 1. A library with an existing `inline def` that accesses a private member. Here the author should create a binary copmatible accessor with the same name as was previously compiler-generated. The warning tells the user that accessor name. It was actually suggested that the compiler could insert the accessor to the source file using `-rewrite`.

We have all the necessary information to be able to create the accessor using `-rewrite`. I will have a look at how to implement this later.

> 2. When someone writes a new `inline def` that accesses a private member, the user should get a warning. Here the warning should not suggest to create an accessor with a mangled name with `$` signs. Is this case also addressed?

I am not sure if that is possible. I have not found a way to know if an `inline def` is new or not while compiling that source file. It seems we only know this kind of information in MiMa or TASTy MiMa.

lrytz:
> I am not sure if that is possible. I have not found a way to know if an `inline def` is new or not while compiling that source file. It seems we only know this kind of information in MiMa or TASTy MiMa.

It should be fine if the compiler message is clear about the various options, which you implemented already behind `-explain` IIUC.

nicolasstucki:
> It should be fine if the compiler message is clear about the various options, which you implemented already behind -explain IIUC.

This is the `-explain` message @lrytz mentioned: https://github.com/lampepfl/dotty/pull/18402/files#diff-e054695755ff26925ae51361df0f7cd4940bc1fd7ceb658023d0dc38c18178c3R3106-R3111

PR #60, opened 2023-03-10
SIP-54 - Multi-Source Extension Overloads.

sjrd:
@odersky I added your name as author of the proposal, since the proposed spec changes are actually yours.

### DIFF
--- a/content/multi-source-extension-overloads.md
+++ b/content/multi-source-extension-overloads.md
@@ -1,0 +1,178 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: implementation
+status: waiting-for-implementation
+title: SIP-54 - Multi-Source Extension Overloads
+---
+
+**By: Sébastien Doeraene and Martin Odersky**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Mar 10th 2023 | Initial Draft      |
+
+## Summary
+
+We propose to allow overload resolution of `extension` methods with the same name but imported from several sources.
+For example, given the following definitions:
+
+```scala
+class Foo
+class Bar
+
+object A:
+  extension (foo: Foo) def meth(): Foo = foo
+  def normalMeth(foo: Foo): Foo = foo
+
+object B:
+  extension (bar: Bar) def meth(): Bar = bar
+  def normalMeth(bar: Bar): Bar = bar
+```
+
+and the following use site:
+
+```scala
+import A.*
+import B.*
+
+val foo: Foo = ???
+foo.meth() // works with this SIP; "ambiguous import" without it
+
+// unchanged:
+meth(foo)() // always ambiguous, just like
+normalMeth(foo) // always ambiguous
+```
+
+## Motivation
+
+Extension methods are a great, straightforward way to extend external classes with additional methods.
+One classical example is to add a `/` operation to `Path`:
+
+```scala
+import java.nio.file.*
+
+object PathExtensions:
+  extension (path: Path)
+    def /(child: String): Path = path.resolve(child).nn
+
+def app1(): Unit =
+  import PathExtensions.*
+  val projectDir = Paths.get(".") / "project"
+```
+
+However, as currently specified, they do not compose, and effectively live in a single flat namespace.
+This is understandable from the spec--the *mechanism**, which says that they are just regular methods, but is problematic from an intuitive point of view--the *intent*.
+
+For example, if we also use another extension that provides `/` for `URI`s, we can use it in a separate scope as follows:
+
+```scala
+import java.net.URI
+
+object URIExtensions:
+  extension (uri: URI)
+    def /(child: String): URI = uri.resolve(child)
+
+def app2(): Unit =
+  import URIExtensions.*
+  val rootURI = new URI("https://www.example.com/")
+  val projectURI = rootURI / "project/"
+```
+
+The above does not work anymore if we need to use *both* extensions in the same scope.
+The code below does not compile:
+
+```scala
+def app(): Unit =
+  import PathExtensions.*
+  import URIExtensions.*
+
+  val projectDir = Paths.get(".") / "project"
+  val rootURI = new URI("https://www.example.com/")
+  val projectURI = rootURI / "project/"
+  println(s"$projectDir -> $projectURI")
+end app
+```
+
+*Both* attempts to use `/` result in error messages of the form
+
+```
+Reference to / is ambiguous,
+it is both imported by import PathExtensions._
+and imported subsequently by import URIExtensions._
+```
+
+### Workarounds
+
+The only workarounds that exist are unsatisfactory.
+
+We can avoid using extensions with the same name in the same scope.
+In the above example, that would be annoying enough to defeat the purpose of the extensions in the first place.
+
+The only other possibility is to *define* all extension methods of the same name in the same `object` (or as top-level definitions in the same file).
+This is possible, although cumbersome, if they all come from the same library.
+However, it is impossible to combine extension methods coming from separate libraries in this way.
+
+### Problem for migrating off of implicit classes
+
+Scala 2 implicit classes did not suffer from the above issues, because they were disambiguated by the name of the implicit class (not the name of the method).
+This means that there are libraries that cannot migrate off of implicit classes to use `extension` methods without significantly degrading their usability.
+
+## Proposed solution
+
+We propose to relax the resolution of extension methods, so that they can be resolved from multiple imported sources.
+Instead of rejecting the `/` call outright because of ambiguous imports, the compiler should try the resolution from all the imports, and keep the only one (if any) for which the receiver type matches.
+
+Practically speaking, this means that the above `app()` example would compile and behave as expected.
+
+### Non-goals
+
+It is *not* a goal of this proposal to allow resolution of arbitrary overloads of regular methods coming from multiple imports.
+Only `extension` method calls are concerned by this proposal.
+The complexity budget of relaxing *all* overloads in this way is deemed too high, whereas it is acceptable for `extension` method calls.
+
+For the same reason, we do not propose to change regular calls of methods that happen to be `extension` methods.
+
+### Specification
+
+From the [specification of extension methods](https://docs.scala-lang.org/scala3/reference/contextual/extension-methods.html#translation-of-calls-to-extension-methods), we amend step 1. of "The precise rules for resolving a selection to an extension method are as follows."
+
+Previously:
+
+> Assume a selection `e.m[Ts]` where `m` is not a member of `e`, where the type arguments `[Ts]` are optional, and where `T` is the expected type.
+> The following two rewritings are tried in order:
+>
+> 1. The selection is rewritten to `m[Ts](e)`.
+
+With this SIP:
+
+> 1. The selection is rewritten to `m[Ts](e)` and typechecked, using the following slight modification of the name resolution rules:
+>
+>    - If `m` is imported by several imports which are all on the same nesting level, try each import as an extension method instead of failing with an ambiguity.
+>      If only one import leads to an expansion that typechecks without errors, pick that expansion.
+>      If there are several such imports, but only one import which is not a wildcard import, pick the expansion from that import.
+>      Otherwise, report an ambiguous reference error.
+
+### Compatibility
+
+The proposal only alters situations where the previous specification would reject the program with an ambiguous import.
+Therefore, we expect it to be backward source compatible.
+
+The resolved calls could previously be spelled out by hand (with fully-qualified names), so binary and TASTy compatibility are not affected.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


bjornregnell:
For clarity and completeness perhaps add also what happens with "normal" call syntax, i.e. add also this line to the example:
```
meth(foo) // works with this SIP; "ambiguous import" without it
```
Also, I wonder what the situation is if extension methods are used from Java. Is the situation improved there as well?

odersky:
> For clarity and completeness perhaps add also what happens with "normal" call syntax, i.e. add also this line to the example:

I agree. In particular since it does not work as you might expect:
```scala
meth(foo)      // always gives "ambiguous import"
```

bjornregnell:
Hmm. That is surprising. Is it to complex to fix that as well?

soronpo:
> Hmm. That is surprising. Is it to complex to fix that as well?

I would actually expect this to be ambiguous, just like regular definitions are in this kind of application.

odersky:
> Hmm. That is surprising. Is it to complex to fix that as well?

Yes, that would require a completely different approach which would likely interact badly with lots of other things. The trick is that we enter the special import node only if we have already parsed `e.m` and now try a `m(e)`.

bjornregnell:
Ok I see. Well the current docs says more or less that `e.m` is just another way of `m(e)` so people might think that they are equivalent. Hence I think this "irregularity" should be explained in docs as well - should proposed docs text be part of this SIP too? 
https://docs.scala-lang.org/scala3/reference/contextual/extension-methods.html#translation-of-extension-methods

bjornregnell:
"So, the definition of circumference above translates to the following method, and can also be invoked as such"

content/multi-source-extension-overloads.md
```diff
@@ -0,0 +1,178 @@
...
+```scala
+import A.*
+import B.*
+
+val foo: Foo = ???
+foo.meth() // works with this SIP; "ambiguous import" without it
+
+// unchanged:
+meth(foo)() // always ambiguous, just like
+regularMeth(foo) // always ambiguous
```
JD557, line 46: I think this should have been `normalMeth` (as defined above)

```suggestion
normalMeth(foo) // always ambiguous
```

soronpo:
@odersky Can you please elaborate on the motivation why extension methods are available to the namespace like regular methods? I think https://github.com/lampepfl/dotty-feature-requests/issues/342 raises a real issue of namespace pollution, and I wonder if we should just remove the ability to call extension methods like regular methods altogether (it's perfectly fine that underneath the hood they are implemented as regular methods).

odersky:
Extension methods have to be regular methods with _some_ name, since we need to abstract over them. I.e. 

```scala
``scala
trait SemiGroup[T]:
  extension (x: T) def combine (y: T): T

trait Monoid[T] extends SemiGroup[T]:
  def unit: T

given Monoid[String] with
  extension (x: String) def combine (y: String): String = x.concat(y)
  def unit: String = ""
```
I considered several alternatives of mangling the name somehow but these are all unappealing. Mangling would not help in the situation here since we are talking about namespace conflicts with two different extension methods.

bishabosha:
> I considered several alternatives of mangling the name somehow but these are all unappealing. Mangling would not help in the situation here anyway since we are talking about namespace conflicts with two different extension methods.

doesn't the `extension` flag count as mangling? completions/name resolution for unqualified ident `foo` could filter out by the extension flag

sjrd:
> "So, the definition of circumference above translates to the following method, and can also be invoked as such"

Good point. I think we'll need to qualify that a bit. Probably something like: translates to the following method *or a more explicitly qualified version thereof*, and can also be invoked as such.

content/multi-source-extension-overloads.md
```diff
@@ -0,0 +1,178 @@
...
+
+*Both* attempts to use `/` result in error messages of the form
+
+```
+Reference to / is ambiguous,
+it is both imported by import PathExtensions._
+and imported subsequently by import URIExtensions._
+```
+
+### Workarounds
```
prolativ, line 107: Just for the completeness of the SIP - there seems to be one more workaround, probably first (partially) mentioned in https://github.com/lampepfl/dotty/issues/13613. The trick is to define extension methods in given instances of type `{}` (so that they don't clash with other useful givens during implicit search). The given instances should have different names so that they could be imported to the scope where we need to use the extension methods. E.g.
```scala
class Foo
class Bar
object A:
  given fooOps: {} with
    extension (foo: Foo) def meth(): Foo = foo
object B:
  given fooOps: {} with
    extension (bar: Bar) def meth(): Bar = bar
```

```scala
import A.given
import B.given
val foo: Foo = ???
val bar: Bar = ???
val a = foo.meth()
val b = bar.meth()
```

bjornregnell:
> doesn't the `extension` flag count as mangling

I wouldn't think it was too strange if the normal method was named something like `extension_circumfence(c: Circle): Double` if that can help to make multi-imported overloaded normal calls as non-ambiguous as extension-mode calls by this SIP.  And that might also help the "name-space-polution" issue.

prolativ:
@bjornregnell do you currently have any problems with discoverability of extension methods in REPL?  Not sure what you would like to improve here.

Sporarum:
"I considered several alternatives of mangling the name somehow but these are all unappealing."

Would it be possible to expand on that (not necessarily in this thread) ?
For example, why wouldn't `extension$foo` work as the method name ?
(Even though it doesn't solve the problem at hand)

bjornregnell:
> problems with discoverability

image

It is not easy to know, when pressing tab, which methods are extension methods. But that's maybe not so often you want to know that...

soronpo:
> Mangling would not help in the situation here anyway since we are talking about namespace conflicts with two different extension methods.

Yeah, I wasn't referring to that as a possible solution to the extension methods collision, but just the namespace one. 
To solve the namespace pollution problem, it seems like we are missing a modifier/annotation that tells the compiler we want some extension methods to be available as `m.ext(a)` and not `ext(m)(a)`. But that's a different SIP.

bjornregnell:
> but this has little to do with this SIP

@prolativ Except that there are differences in calling it as an extension or as a normal method... (as the first sometimes works but not the second by this SIP...) so it might be important to know that it actually *is* an extension, esp if the namespace that you press TAB on has many names. 

Perhaps it would be a good enough hint if the error message "Reference to meth is ambiguous," would also reveal that it actually *is* an extension method and that it would *indeed* be non-ambigous if it was called that way. Could we have such an error message?

Currently the compiler does not reveal that the normally called method actually is an extension:
```
Welcome to Scala 3.2.2 (17.0.6, Java OpenJDK 64-Bit Server VM).
                                                                                
scala> class Foo
     | class Bar
     | object A:
     |   extension (foo: Foo) def meth(): Foo = foo
     |   def normalMeth(foo: Foo): Foo = foo
     | object B:
     |   extension (bar: Bar) def meth(): Bar = bar
     |   def normalMeth(bar: Bar): Bar = bar
     | 
// defined class Foo
// defined class Bar
// defined object A
// defined object B
                                                                                
scala> import A.*
     | import B.*
                                                                                
scala> val foo = new Foo
val foo: Foo = Foo@36463b09
                                                                                
scala> meth(foo)()
-- [E049] Reference Error: -----------------------------------------------------
1 |meth(foo)()
  |^^^^
  |Reference to meth is ambiguous,
  |it is both imported by import A._
  |and imported subsequently by import B._
  |
-- [E007] Type Mismatch Error: -------------------------------------------------
1 |meth(foo)()
  |     ^^^
  |     Found:    (foo : Foo)
  |     Required: Bar
  |
2 errors found
                 

```

content/multi-source-extension-overloads.md
```diff
@@ -0,0 +1,178 @@
...
+>
+> 1. The selection is rewritten to `m[Ts](e)`.
+
+With this SIP:
+
+> 1. The selection is rewritten to `m[Ts](e)` and typechecked, using the following slight modification of the name resolution rules:
+>
+>    - If `m` is imported by several imports which are all on the nesting level, try each import as an extension method instead of failing with an ambiguity.
+>      If only one import leads to an expansion that typechecks without errors, pick that expansion.
+>      If there are several such imports, but only one import which is not a wildcard import, pick the expansion from that import.
```
smarter, line 155: Are there other situations where we treat named imports differently from wildcard imports?

odersky: Yes, a named import beats a wildcard import when they are in the same scope, just like a definition beats either import.

smarter (approved):

julienrf:
@bjornregnell and @soronpo, do you want to request a vote on this proposal at the next SIP meeting?

bjornregnell:
Yes, I'm prepared to vote and my recommendation is to **accept**, with the added recommendation to update docs and improve the error message as discussed above:
* The doc text "So, the definition of circumference above translates to the following method, and can also be invoked as such" does give the impression that the different ways of calling it are equivalent while they are not.
* The error message on "Reference to meth is ambiguous" could be more helpful, by e.g. saying that it is an extension method and what are the new rules according to this SIP.

sjrd:
I can do it next Monday.

soronpo:
> @bjornregnell and @soronpo, do you want to request a vote on this proposal at the next SIP meeting? In such a case, what is your recommendation?

I recommend to **accept**. Just pointed out minor documentation issues. The proposal is solid and solves the problem it strives to solve.

soronpo:
On Apr 21st the SIP committee discussed the SIP and accepted the proposal. 
It provides a good tradeoff between limited complexity and language consistency when enabling extension methods with the same name to be imported from different scopes while regular methods cannot. It is hoped that this is another good step towards deprecating implicit classes in the future in favor of extension methods. 
Thank you very much for the efforts!

bjornregnell:
In addition to the excellently succinct summary by @soronpo, I'd like to add that there are two specific points from the discussion that I think should be addressed (see details in the thread above):
1. Update of the docs to explain the semantic differences between calling an extension method as an extension or a "normal" method, as introduced by this SIP.
2. Improvement of error messages when ambiguity errors arise due to overloading to inform the user that this is because it is an extension method and explain the reason for the ambiguity (that the extension method is called as a "normal" method etc)

odersky:
@bjornregnell I am not sure how to improve the error messages in this case. Do you have a specific usage scenario where you would like a change?

bjornregnell:
See also above https://github.com/scala/improvement-proposals/pull/60#issuecomment-1468160388

odersky:
Thanks for clarifying! yes, I think we can do that.

julienrf:
Dear @smarter, @bjornregnell, and @soronpo. The authors of the proposals would like the Committee to vote on promoting to a stable feature. Do you agree to request a vote in our next SIP meeting? If yes, what would be your recommendation?

julienrf:
FTR, I’ve tried with IntelliJ (231.8109.175), and it supports the feature. The code example given in the content of the proposal is correctly highlighted (no red squiggle), the extension method is listed in the auto-completion popup, and “go to definition” works as expected.

I wanted to also try with Metals but it seems Metals does not support nightly versions of Scala 3. I expect the feature will just work out of the box in Metals when it becomes available in a non-nightly compiler version.

prolativ:
@julienrf as far as I know Metals should support 5 latest nightly versions of the compiler (probably with a delay of 1 day, for some reason, unless this problem has already got solved). But you would also have to use the nightly version of Metals for this to work.

bjornregnell:
> Do you agree to request a vote in our next SIP meeting? If yes, what would be your recommendation?

Yes I agree and my recommendation is to vote `yes`.

bjornregnell:
A better usability of err messages is not a show-stopper, but it is a bit unsatisfactory that an implementation details prevents nicer UX. I would have assumed that the keyword `extension` would enable the compiler to mark that is an extension method early on and that it should be a simple if-expression to emit a string like "note this is an extension method...". But if this is not the case and it is too much work or too much architectural turmoil for this SIP then the potential for improved UX can e.g. be deferred to an issue for future fix.

odersky:
> and that it should be a simple if-expression to emit a string like "note this is an extension method...

Yes, that would be possible. However, we don't know whether this was called as an extension method or not. If we add the note: "this ambiguity would go away if the method is called as an extension method" and the method was in fact called as an extension method (but the error would not be filtered away for some other reason), this would be embarrassing.

smarter:
> However, we don't know whether this was called as an extension method or not.

Could we compare the span of the extension method with the span of its argument to figure that out? In any case, I'm +1 to accepting this at the next meeting.

bjornregnell:
> this would be embarrassing
Good then that compilers don't have consciousness (yet) and can't get embarrassed :)

I'd assume that the call-site could lookup the declaration and see if it's an extension or not, but I don't know the implementation so this comment might be off. 

Anyway, the comment could be a hint rather than very certain, e.g. "hint: this error may arise if xxx is an extension called as a normal method" which might be less "embarrassing" and just "noise" if its not the case. But I'm not sure if it is worse to introduce noise and false positive. So better if the hint can be reliably given when it actually *is* an extension.

odersky:
I think a hint like that is a good compromise.

PR #63, opened 2023-06-15
SIP-55 - Concurrency with Higher-Order Coroutines

diesalbla:
A SIP for introducing coroutines to Scala using implicit parameters. We deal with usual coroutine problems such as colour-transparency and higher-order functions, as well as with concurrency problems.

This SIP follows the Pre-SIP discussions about [Suspended functions and continuations](https://contributors.scala-lang.org/t/pre-sip-suspended-functions-and-continuations/5801), and addresses some of the issues discussed in that forum. It also succeeds recent presentatios about [Direct Style Scala](https://www.youtube.com/watch?v=0Fm0y4K4YO8), and the ongoing work with [async](https://github.com/lampepfl/async).

### DIFF
--- a/content/sip-55-coroutines.md
+++ b/content/sip-55-coroutines.md
@@ -113,7 +113,7 @@ def gcd(a: Int, b: Int): Int =
 extension [Y] (list: List[Y]) 
   def map[Z](fili: Y => Z): List[Z] = list match 
     case Nil =>  Nil
-    case y :: ys => fili(y) :: ys.map(fun)
+    case y :: ys => fili(y) :: ys.map(fili)
 
 def pipe[X, Y, Z](tick: X => Y, tock: Y => Z): X => Z =
   (x: X) => tock(tick(x))
@@ -219,7 +219,7 @@ extension [Y] (list: List[Y])
     case y :: ys =>
       for
         z <- fili(y)
-        zs <- ys.mapT(fun)
+        zs <- ys.mapT(fili)
       yield z :: zs
 ```
 
--- a/content/sip-55-coroutines.md
+++ b/content/sip-55-coroutines.md
@@ -1,0 +1,965 @@
+# Scala Improvement Process - Coroutines
+
+---
+layout: sip
+permalink: /sips/:title.html
+stage: design
+status: submitted
+title: SIP-55 - Concurrency with Coroutines
+---
+
+**By: Jack Viers and Diego Alonso and Raul Raja**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| 8th July 2022 | [PRE-SIP Proposal](https://contributors.scala-lang.org/t/pre-sip-suspended-functions-and-continuations/5801) |
+| 19th April 2023 | Initial Draft    |
+| 14th June 2023 | SIP submission |
+
+## Summary
+
+This SIP implements concurrency in Scala using coroutines.
+A coroutine is written as a function with an implicit `Suspend` parameter.
+The compiler transforms such a function into a state-machine class.
+We leverage parametric polymorphism to write color-transparent higher order functions,
+and thus address the [two-function-color problem](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/).
+No changes to the syntax or type system of Scala are needed.
+The only front-end changes are generated additions to the standard library. 
+The implementation is
+done in the compiler backend, which does not depend on the runtime platform.
+This is an opt-in feature, which has no effect on any code without coroutines.
+We have [developed a prototype](https://github.com/47deg/scourt) implementation.
+
+**Non-Goals:** This SIP does not improve or deprecate `for` comprehensions. 
+We do not aim to introduce a generalized form of delimited continuations 
+or algebraic effects.
+
+## Motivation
+
+_Concurrency_ is a programming language feature that allows us 1)
+to split a program into parts that can run separately from others; and 2)
+to mix those parts in a larger program that controls how 
+those separate parts are run together and joins them into a final result.
+
+Most operating systems offer managed concurrency solutions 
+such as processes and threads. However, since these are 
+relatively heavy, many languages have a form of self-managed
+lightweight concurrency, such as fibers or coroutines.
+Some examples are
+the asynchronous modality for expressions in F#, 
+[async-await in C#](https://learn.microsoft.com/en-us/events/dotnetconf-2020/asynchronous-courotines-with-c),
+[goroutines in Go](https://go.dev/tour/concurrency/1), 
+[suspended functions in Kotlin](https://kotlinlang.org/docs/coroutines-guide.html#table-of-contents),
+or [Virtual Threads](https://openjdk.org/jeps/444) in Java.
+The solution adopted for each language was designed to fit 
+with the design of the rest of the language.
+
+Scala, on the other hand, lacks direct language support for concurrency.
+Instead, this has been supported by a succession of libraries over the 
+years, such as Twitter util ([2010](https://github.com/twitter/util/commit/34f8876133f7f45ce3499c0e3df9e119e54c1a54#diff-0295d2fd9c8a7570fe2351b8c5c47d346c58bdb4f2c02a91b22a2fd5a682c96d)),
+the Scala concurrent library ([2011](https://github.com/scala/scala/commit/d38ca89f8d50c0d6eafac623f0edc165c01f8bd4)),
+`scalaz` ([2013](https://github.com/scalaz/scalaz/commit/2f49a3b58fb94bb5b0189f97b67982be55af9485)),
+`scalaz-streams` --nowadays `fs2`-- ([2013](https://github.com/scalaz/scalaz-stream/commit/58a00e6836aedeb59616b94b12731b34833485fb)),
+`monix` ([2015](https://github.com/monix/monix/commit/e623f3475fcd7076930004e93fe68a86d2e1eb00)),
+`cats-effect` ([2017](https://github.com/typelevel/cats-effect/commit/74dd07050712f66c4c78d7e1fb056078e013e3b4)),
+and `zio` ([2018](https://github.com/zio/zio/commit/1162f6bb79dc17eaf0dd810361b60cf4acdff22b)). 
+None of these has given a fully satisfactory solution, one which combines
+a syntax that fits Scala design and a clear semantics.
+In this SIP we propose coroutines as the solution for concurrency in Scala.
+
+#### Idiomatic concurrency for Scala
+
+Let us first point out what features an ideal approach to 
+concurrency in Scala should have:
+
+* Since Scala is an expression-oriented language, expressions should be the 
+  level at which concurrency can occur.
+* Scala is a strict and sequential language: expressions are evaluated 
+  in a fixed order, and each evaluation ends before the next one begins.
+  Any shift from this semantics, such as the laziness of local variables 
+  or parallel evaluation, should have a explcit mark such as the `lazy` modifier.
+* Scala is a higher-order language. Functions can take other functions as
+  parameters and return functions as a result. Idiomatic concurrency 
+  in Scala should allow us to define **higher-order functions** (HOF) 
+  that apply to concurrent or non-concurrent functions.
+* A **direct style** syntax helps us mix concurrent expressions with the rest 
+  of Scala as easily as with non-concurrent ones. It 
+  should let us use them in a `val` declaration, a `var` assignment,
+  the conditions or the branches of an `if-then-else`, 
+  the cases of a pattern match, etc.
+* An **applicative** style should let us apply any function, 
+  concurrent or not, to any expression, concurrent or not, 
+  using the same syntax for "normal" functions and expressions,
+  without the need for any special combinators.
+
+#### Running examples
+
+We compare our solution with the _status quo_ using three examples: 
+a modified `gcd` as an example of a first-order function;
+a `map` for lists as an example of a generic HOF; 
+and a `pipe,` as an example of a combinator that composes 
+two input functions and returns another function. 
+Here are they in "plain" code:
+
+``` scala
+def isZero(n: Int): Boolean
+def mod(num: Int, den: Int): Int
+
+def gcd(a: Int, b: Int): Int =
+  if isZero(b) then a else gcd(b, mod(a, b)) + 1
+
+extension [Y] (list: List[Y]) 
+  def map[Z](fili: Y => Z): List[Z] = list match 
+    case Nil =>  Nil
+    case y :: ys => fili(y) :: ys.map(fun)
+
+def pipe[X, Y, Z](tick: X => Y, tock: Y => Z): X => Z =
+  (x: X) => tock(tick(x))
+```
+
+These examples are not meant to give industrial motivation. We use 
+them to show how expressive or verbose each solution can be if we 
+want to make `isZero`, `mod`, `fili`, `tick` or `tock` concurrent.
+
+#### Threads
+
+The Java Virtual Machine (JVM) supports a wrapped form of OS threads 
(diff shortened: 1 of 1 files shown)


He-Pin:
过年了！Epic!

content/sip-55-coroutines.md
```diff
@@ -0,0 +1,965 @@
...
+Just as the result type `List[B]` is tied to the `fili` function by 
+the `B` parameter, we can tie the `Suspend` of `fili` with the `Suspend` 
+of `mapC` by using a "color" type parameter:
+
+``` scala
+sealed trait Color 
+sealed trait Green extends Color
+sealed trait Red extends Green
+
+sealed trait Suspend[+ Col <: Color]
```
He-Pin, line 362: What about name it `Suspendable`?

julienrf:
Thank you for submitting the proposal. I’ve assigned a team of reviewers who will post their feedback within the next couple of weeks.

bjornregnell:
I'm not a reviewer but I think I found a typo:
`fun` should be changed to `fili` in two places (if I'm not incorrect).

doofin:
I didn't find a systematic survey about why the current monadic syntax is not sufficient aside from being deemed too difficult for beginners. It would be great if there's something to justify this new syntax.

content/sip-55-coroutines.md
```diff
@@ -0,0 +1,965 @@
...
+parameters, all we need is to entangle those types.
+Luckily, Scala has a tool to tie two types: parametric polymorphism.
+Just as the result type `List[B]` is tied to the `fili` function by 
+the `B` parameter, we can tie the `Suspend` of `fili` with the `Suspend` 
+of `mapC` by using a "color" type parameter:
+
+``` scala
+sealed trait Color 
+sealed trait Green extends Color
+sealed trait Red extends Green
```
He-Pin, line 360: `Green` and `Read` is not easy understandable.

content/sip-55-coroutines.md
```diff
@@ -0,0 +1,965 @@
...
+for the underlying launcher to handle the execution of the coroutine.
+
+```scala
+/**
+ * @param result Any current or intermediate result between invocations
+ * @param completion The launcher continuation, used for scheduling and returning results at launcher boundaries.
+ */
+class GcdFrame(
+  var state: Int, 
+  var a: Int, var b: Int, var z: Boolean, var m: Int, var g: Int,
```
He-Pin, line 464: What if we have many local varaibles ,will that be a problem?

diesalbla: An object of the `Frame` class stores in heap the method's stack frame. Just as with the stack frame, the size is proportional to the number of local variables in the method, which is known at compile time. We could do some optimisations, similar to the ones the compiler does with stack frames. For instance, we could have two local variables with non-overlapping time-lines (span between declaration or assignment and use) share a same "position". 

Note that the focus of this SIP is 1) the syntax for coroutines, and 2) a possible implementation known to work in other languages. Such optimisations would be addressed down the line.

content/sip-55-coroutines.md
```diff
@@ -0,0 +1,965 @@
...
+def isZero(n: Int): Boolean
+def mod(num: Int, den: Int): Int
+
+def gcd(a: Int, b: Int): Int =
+  if isZero(b) then a else gcd(b, mod(a, b)) + 1
+
+extension [Y] (list: List[Y]) 
+  def map[Z](fili: Y => Z): List[Z] = list match 
+    case Nil =>  Nil
+    case y :: ys => fili(y) :: ys.map(fun)
```
bjornregnell, line 116: fun -> fili

content/sip-55-coroutines.md
```diff
@@ -0,0 +1,965 @@
...
+ `Task[List[Z]]` as its return type. Thus the code that follows:
+
+``` scala
+extension [Y] (list: List[Y])
+  def mapT[Z](fili: Y => Task[Z]): Task[List[Z]] = list match
+    case Nil => Task.pure(Nil)
+    case y :: ys =>
+      for
+        z <- fili(y)
+        zs <- ys.mapT(fun)
```
bjornregnell, line 222: fun -> fili

sjrd:
First, thank you for this extensive, well-research and well-presented SIP.

The SIP addresses an important problem; one for which I think many agree want a good solution. The solution space is however daunting. As a result, the solution proposed in this SIP contains substantial complexity, and it is likely that it will need several iterations, or never pass at all.

Now onto my high-level comments on this SIP.

## Some things I like

Just to highlight some good stuff that I particularly like. ;)

* The launcher abstraction. Thanks to that, the language feature is not tied to specific implementation details that may have performance tradeoffs.
* Structured concurrency.

## Keyword versus contextual parameters

This SIP proposes to use contextual parameters to encode the color of a function. It argues that contextual params are better because they do "not require modifying the lexicon or grammar of Scala, nor adding any major features to its system". While that is true, it does require special *restrictions* on how those contextual parameters can be used. See the section "Front-End Changes".

However necessary, the restrictions are severe enough that they make `Suspend` contextual params significantly depart from regular contextual params. For example, it means that these types can never be abstracted over; they do not really belong in the type lattice. In fact, they cannot even be subtypes of `AnyKind`, which is an entirely new world for the spec.

IMO, this is significant enough that we cannot in fact reason about them in the same way as other contextual params. Given that, it is not clear to me that "only being contextual parameters" is a good thing. In practice, I expect that it would be easier to specify and implement if we took coloring as a separate feature.

This aspect also compounds with my next comment.

## Codegen and Binary/TASTy compatibility

Color-transparent functions, and a fortiori polychromatic functions, require 2 "lanes". Each lane will require a different actual method in the compilation pipeline. How do we guarantee binary compatibility of these generated-but-public methods? Also, it is not clear what happens when such methods are *overridden* (or implement abstract methods). Do we have some guarantee that all the semantics of overriding can be preserved by the code transformation?

Regarding my earlier comment about contextual parameters, it also means that *some* contextual parameters, depending on what type they have, have a profound impact on the binary API of a method. Normally, contextual parameters are all handled in the same way and have no such consequences.

## Integration with the standard library

Integration with the standard library is problematic. As is, no higher-order function from the stdlib, such as `List.map`, will work in async contexts. The proposal mentions that adapting the stdlib is left for future work. Unfortunately, I think this will leave us in an intermediate state that is not really acceptable. It will be confusing that standard methods cannot be used with such a deep-reaching language feature.

Ironically, truly accepting the language-feature aspect of this SIP, instead of encoding it as contextual parameters, may provide a better path forward. If existing methods, with existing signatures, can be marked as color-transparent, we may be able to enhance the existing standard library with async support in a backward compatible way. This is not really possible if the signature of the method has to change, which is the case with contextual parameters.

odersky:
I echo @sjrd: This is a great write-up for a SIP. Thanks to the authors for the clarity of writing! Thanks in particular for the set of example programs to be used in case studies. I found these very well chosen.

The SIP proposes a solution to adding coroutines (or rather: one-shot continuations) to Scala on runtimes that don't support these features directly. Runtimes that would support coroutines directly are JVM 21 and later through Loom and possibly Scala Native. But on earlier JVM versions and JS there is no native support in sight. That said, the proposal stresses that there should be a single language standard that can be implemented on all kinds of runtimes.

The proposal covers three areas:

 1. Changes to the language standard and to the compiler frontend
 2. Transforms needed to implement continuations
 3. A runtime that supports higher level structured concurrency abstractions built on the suspension primitive.

The main purpose of the SIP is to discuss (1). (2) is relevant insofar as it demonstrates that the proposal can be implemented. (3) is relevant insofar as it demonstrates expressiveness - we can build nice abstractions for async computation on top of coroutines. Let me discuss them in reverse order:

## Runtime

I liked the proposed abstractions for Launchers. This holds a lot of promise for cross-platform concurrency libraries.
One question: On JVM, we already have `InterruptedException`. Do you mean `CancellationException` is an alias of that or should it be separate?

## Transforms

The transforms described map suspendable functions to state machines where the state is stored in heap objects. This is similar to the techniques used in Kotlin. It also seems to have a lot of resemblance with the techniques in Scala 2.13's async translation - it would be good to explore this further.

The tradeoff of any implementation technique is between how efficient suspensions are and how much slowdown is accepted for transformed code that does not in the end suspend, or that suspends only rarely. State machines are quite efficient for suspensions but impose a significant overhead for normal code execution. The technique of snapshotting via exceptions mentioned in the proposal is an alternative that does the tradeoff in a different direction, by implementing non-suspending code with almost no slowdown. Before deciding on one or the other it might be good to compare both candidate implementations with benchmarks. As the authors note, this is not a mere implementation detail, since exceptions are exposed to programmers and therefore would have to be specified. 

The tradeoff is also influenced  by how much code gets transformed in the end to make it potentially suspendable. For instance, if we decide that all code should be potentially suspendable (probably not realistic), then we definitely need an implementation that does not penalize straight code that does not suspend. If we have an extremely good prediction of what code might be suspendable we can be less stringent in our requirement. The proposal is to indicate suspendability through an argument of a polymorphic method and to specialize on that parameter, which gives probably a reasonable upper approximation of what can suspend.

## Language and Compiler Frontend

The proposal is to use context parameters to indicate suspendability, but to be sound it needs to restrict `Suspend` context parameters to be [second class values](https://www.cs.purdue.edu/homes/rompf/papers/osvald-oopsla16.pdf). Contrary to what is claimed in the proposal this is a _huge_ change of the type system! A simple `suspend` modifier as in Kotlin would be a much smaller change to the language than this. 

I understand the appeal of context parameters since they tend to get out of the way without needing wrapping or unwrapping, and since they can carry type parameters (which are the colors in this proposal). But still, one cannot simply change the rules like this without risk of feature interactions with everything else. 

There are two other problems:

 - These changes would not be binary backwards compatible (@sjrd already expanded on this, so I won't need to).
 - These changes would make the source definitions a lot more complicated. Compare the definition of `map` with the current one in the standard library. And `map` is not alone. In an OO language like Scala every call on a parameter method is a virtual method call, so  every call could potentially suspend or not. This means we probably need a large scale proliferation of `Color` type parameters in `Suspend` arguments. 

I believe the proposed restriction of context parameters to second class values is not only a big language change and therefore probably unfeasible but also a missed opportunity. We should embrace the fact that `Suspend` parameters can be 
captured in closures since it gives us much cleaner types for higher-order functions like `map` and `pipe`! This is the essence of [capture checking](https://arxiv.org/abs/2207.03402). To demonstrate, here are the three example functions expressed with Scala's capture checking language import:
```scala
def isZero(n: Int)(using Async): Boolean =
  Future(n == 0).await
def mod(num: Int, den: Int)(using Async): Int =
  Future(num % den).await

def gcd(a: Int, b: Int)(using Async): Int =
  if isZero(b) then a else gcd(b, mod(a, b)) + 1

extension [Y] (list: List[Y])
  def map[Z](fili: Y => Z): List[Z] = list match
    case Nil =>  Nil
    case y :: ys => fili(y) :: ys.map(fili)

def pipe[X, Y, Z](tick: X => Y, tock: Y => Z): X ->{tick, tock} Z =
  (x: X) => tock(tick(x))
```
This uses `Async` instead of `Suspend`, but the two are interchangeable in this context. 

**Notes:**

 - We use a slightly higher level abstraction of awaitable futures for `isZero` and `mod` instead of `shift/resume`. (Btw I find the name `shift` a calamity, please let's use `suspend` instead!)
 - The definition of `map` is exactly what it would be without suspensions. More generally, we can keep the definitions of almost all library functions the same as before. The exception is only in lazy abstractions such as `Iterator` which will need a capture annotation. This changes the source code, but not the binaries.
 - The definition of `pipe` now states explicitly that the result function can have precisely the effects of `tick` and `tock` and _nothing else_. So you get not only suspension implementations, but on top of that effect checking for free!

One advantage of the current SIP over capture checking is that it makes it very clear which functions can suspend and which cannot (the disadvantage is that we need new types for every function that can suspend). But I believe that question can be answered for capture checking as well. Namely, a function can suspend if it is in the scope of a `Suspend` (respectively, `Async`) context parameter or if it invokes a method on another object that might capture a `Suspend` capability. It's a bit more indirect, but I believe it can be worked out. 

## Summary

There are lots of things to like about the proposal: A worked out implementation, the Launcher abstractions, the idiomatic code examples used. But it's an extremely impactful and involved change. Before jumping in one direction, we should try out and compare with other type system and implementation techniques.

gabro:
I second what @odersky and @sjrd said, great job on the proposal!

One minor addition: I too worry about the standard library integration and I don't think that should fall under "future work" since that seems to be a drastic change. Even just from a documentation perspective I fear we will return to a `CanBuilFrom`-like situation in which even explaining a simple method like `map` will become tricky since its signature will include the notion of a Color.

diesalbla:
@sjrd Thanks for the feedback received. I wonder if you could you elaborate a bit further regarding the following 

> the restrictions are severe enough that they make Suspend contextual params significantly depart from regular contextual params. For example, it means that these types can never be abstracted over; they do not really belong in the type lattice. In fact, they cannot even be subtypes of AnyKind, which is an entirely new world for the spec.

Could you elaborate a bit further on this? What form of abstractions, or what language constructs, would not be possible with the encoding that would be possible with a different one?

We are not sure if by "abstraction" you are referring to polymorphism. The examples that in the SIP make use of parametric polymorphism, in which the colour is set by the caller. It would also be possible to have an example of Generalised-Algebraic-Data-Type.

```Scala
trait Recorder[C <: Color]: 
  def record(s: String)(using Suspend[C])

object MemoryWriter extends Recorder[Green]
object DiskWriter extends Recorder[Red]
```

diesalbla:
> Integration with the standard library is problematic. 
> As is, no higher-order function from the stdlib, such as List.map, will work in async contexts. 

@sjrd What is referred to with the words "work in async contexts"? I assume that it refers to the fact that, even from within a coroutine, it would not be possible to call a HOF passing it  another coroutine as argument. Note that it would still be possible to call a HOF, applied to a green function, just like any other green function; except that the caller coroutine would not suspend in that HOF call.

doofin:
Frankly speaking, I only see this new syntax as a kind of shorthand for using monads which might be easier for new comers. However, this approach may introduce many complexities as @sjrd said. I tried to find the rationale behind this proposal but failed to be convinced.

anatoliykmetyuk:
@sjrd do you have any updates on this one?

Atry:
With the help of virtual threads, all vanilla functions in JVM is already color-transparent.

odersky:
An EPFL student, Guillem Bartrina I Moreno, will take up implementation work in this direction as part of his master's theses. He might want to re-use some of the parts of the current implementation. The plan is to try to use standard capture checking (which has been implemented for the standard library) to determine which methods need to be async-transformed. 
I propose to put the SIP on hold until we see how that work pans out.

odersky:
The work by Guillem is now finished. See https://github.com/scala/scala3/issues/22474 for a summary. This approach does work with the standard library as is (to be precise: with its capture checked version, which binary compatible with the original and is publicly available). 

The SIP committee believes that it would make sense to to proceed along the lines of Guillem's work rather than pursuing the present SIP. The present SIP was very useful for exploring the design space and implementation techniques. We thank the submitters for making it.

Atry:
This is actually the first color transparent continuation implementation among all main streaming languages!

For comparison. Java 21, Lua and Go have color transparent coroutines, which are one-pass continuations that cannot fork.

diesalbla:
Congratulations to @guillembartrina for this achievement, as well as to his advisors @odersky and Nguyên Cao Pham for pursuing this challenging line of development.

> The present SIP was very useful for exploring the design space and implementation techniques. We thank the submitters for making it.

Thanks for the kind words, and we can all be glad that it was useful. I have skimmed a bit through the thesis, and it seems to have been implemented along similar _lanes_ to those sketched in the SIP. At this point, I we can feel comfortable deferring to @odersky assessment and acknowledge this SIP proposal as having been superseded by Guillem's work as a more feasible way forward.

PR #64, opened 2023-07-25
Update deprecated syntax of using directives

lwronski:
Starting with `scala-cli` version `1.0.0-RC1`, the using directives syntax was simplified - (see https://scala-cli.virtuslab.org/docs/release_notes#dropped-deprecated-using-directive-syntax). As such, it would be beneficial to update SIP 46 proposals, which currently reference the outdated using directives syntax.

### DIFF
--- a/content/scala-cli.md
+++ b/content/scala-cli.md
@@ -624,48 +624,25 @@ As a part of this SIP we propose to introduce Using Directives, a special commen
 
 Using directives can be place on only top of the file (above imports, package definition etx.) and can be proceed only by plain comments (e.g. to comment out an using directive)
 
-Comments containing directives needs to start by `>`, for example:
+Comments containing directives needs to start by `//>`, for example:
 
 ```
-//> using Scala 3
-/*> using 
-  Scala 3
-  option "-Xfatal-warning"
-*/
+//> using scala 3
+//> using platform scala-js
+//> using options -Xasync
 ```
 
 We propose following sytax for Using Directives (within special comments described above):
 
 ```
-UsingDirective ::= "using" (Setting | Settings)
-Settings ::= "{" Setting { ";" Setting } [";"] "}"
-Setting ::= Ident [Values | Settings]
-Ident ::= ScalaIdent { “.” ScalaIdent }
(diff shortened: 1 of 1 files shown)


content/scala-cli.md
```diff
@@ -627,46 +627,23 @@ Using directives can be place on only top of the file (above imports, package de
 Comments containing directives needs to start by `>`, for example:
```
SethTisue, line 627: should be `//>`

lwronski: Thanks!

soronpo:
Since SIP46 was already accepted and marked as complete, I don't think we can change this retroactively. This needs to go through the committee again, IMO (and perhaps under a different number). What do you think @julienrf

tgodzik:
> Since SIP46 was already accepted and marked as complete, I don't think we can change this retroactively. This needs to go through the committee again, IMO (and perhaps under a different number). What do you think @julienrf?

This was discussed during the tooling summit and suggested to be introduced before making Scala CLI an official runner, since using directives were thought of as too complex. This sounded like a requirement for the SIP, since a number of people involved were members of SIP committee, but it was hard to figure out what was needed and was not actually needed to do for it. Especially that we introduced a number of concepts, which were a bit more complex and we had a lot of comments on making them better even after the SIP was accepted. 

So the accepted SIP was one thing and the end result was another, since we really went beyond to try and fix any potential issues in cooperation with a lot of people in the community. And possibly this might have been better result then sticking rigidly to the SIP, but on the other hand this was done kind of outside the SIP process and was not really up to us.

If that was done wrong then I apologize, but the current state is as suggested in this PR. I don't mind leaving the SIP as is since this is not really documentation, but @julienrf actually asked us to do it in https://github.com/VirtusLab/scala-cli/issues/2048

julienrf (approved):
Thank you for updating the document!

I will merge the PR right away because not updating it was an oversight. For the record, the SIP Committee asked the maintainers of Scala CLI to update the proposal when they accepted it (see my comment [here](https://contributors.scala-lang.org/t/sip-46-scala-cli-as-default-scala-command/5996/24)).

PR #65, opened 2023-08-11
SIP-56 - Proper Specification for Match Types.

sjrd:
Currently, match type reduction is not specified, and its implementation is by nature not specifiable.
This is an issue because match type reduction spans across TASTy files (unlike, for example, type inference or GADTs), which can and will lead to old TASTy files not to be linked again in future versions of the compiler.

This SIP proposes a proper specification for match types, which does not involve type inference.
It is based on `baseType` computations and subtype tests involving only fully-defined types.
That is future-proof, because `baseType` and subtyping are defined in the specification of the language.

The proposed specification defines a subset of current match types that are considered legal.
Legal match types use the new, specified reduction rules.
Illegal match types are rejected, which is a breaking change, and can be recovered under `-source:3.3`.

### DIFF
--- a/content/match-types-spec.md
+++ b/content/match-types-spec.md
@@ -419,6 +419,42 @@ It means that if we make the scrutinee of a match type more precise (a subtype) 
 Note: if `⋔` were a "true" disjointness relationship, and not a *provably*-disjoint relationship, that property would trivially hold based on elementary set theoretic properties.
 It would amount to proving that if `S ⊆ T` and `T ⋂ U = ∅`, then `S ⋂ U = ∅`.
 
+#### Reduction
+
+The final piece of the match types puzzle is to define the reduction as a whole.
+
+In addition to matching and `provablyDisjoint`, the existing algorithm relies on a `provablyEmpty` property for a single type.
+It was added a safeguard against matching an empty (`Nothing`) scrutinee.
+The problem with doing so is that an empty scrutinee will be considered *both* as matching the pattern *and* as provably disjoint from the pattern.
+This can result in the match type reducing to different cases depending on the context.
+To sidestep this issue, the current algorithm refuses to consider a scrutinee that is `provablyEmpty`.
+
+If we wanted to keep that strategy, we would also have to specify `provablyEmpty` and prove some properties about it.
+Instead, we choose a much simpler and safer strategy: we always test both matching *and* `provablyDisjoint`.
+When both apply, we deduce that the scrutinee is empty is refuse to reduce the match type.
+
+Therefore, in summary, the whole reduction algorithm works as follows.
+The result of reducing `X match { case P1 => R1; ...; case Pn => Rn }` can be a type, undefined, or a compile error.
+For `n >= 1`, it is specified as:
+
+* If `X` matches `P1` with type capture instantiations `[...ts => ts']`:
+  * If `X ⋔ P1`, do not reduce.
+  * Otherwise, reduce as `[...ts => ts']R1`.
+* Otherwise,
+  * If `X ⋔ P1`, the result of reducing `X match { case P2 => R2; ...; case Pn => Rn }` (i.e., proceed with subsequent cases).
+  * Otherwise, do not reduce.
+
+The reduction of an "empty" match type `X match { }` (which cannot be written in user programs) is a compile error.
+
+#### Subtyping
+
+As is already the case in the existing system, match types tie into subtyping as follows:
+
+* `X match { ... } <: T` if `X match { ... }` reduces to `S1` and `S1 <: T`
+* `S <: X match { ... }` if `X match { ... }` reduces to `T1` and `S <: T1`
+* `X match { ... }  <: T` if `X match { ... }` has the upper bound `H` and `H <: T`
+* `X match { case P1 => A1; ...; case Pn => An } <: Y match { case Q1 => B1; ...; Qn => Bn }` if `X =:= Y` and `Pi =:= Qi` for each `i` and `Ai <: Bi` for each `i`
+
 ### Compatibility
 
 Compatibility is inherently tricky to evaluate for this proposal, and even to define.
--- a/content/match-types-spec.md
+++ b/content/match-types-spec.md
@@ -272,8 +272,8 @@ At the top level, `variance = 1` and `scrutIsWidenedAbstract = false`.
     * Let `innerScrutIsWidenedAbstract` be true if either `scrutIsWidenedAbstract` or `X` is not a concrete type.
     * For each pair of `(Ui, Qi)`, compute `matchPattern(Ui, Qi, vi, innerScrutIsWidenedAbstract)` where `vi` is the variance of the `i`th type parameter of `T`.
   * If `T` is `scala.compiletime.ops.int.S`:
-    * If `n = natValue(X)` is undefined or is `Int.MinValue`, fail as not matching.
-    * Otherwise, compute `matchPattern(n, Q1, 1, scrutIsWidenedAbstract)`.
+    * If `n = natValue(X)` is undefined or `n <= 0`, fail as not matching.
+    * Otherwise, compute `matchPattern(n - 1, Q1, 1, scrutIsWidenedAbstract)`.
   * If `T` is an abstract type constructor:
     * If `X` is not of the form `F[Us]` or `F =:= T` is false, fail as not matching.
     * Otherwise, for each pair of `(Ui, Qi)`, compute `matchPattern(Ui, Qi, vi, scrutIsWidenedAbstract)` where `vi` is the variance of the `i`th type parameter of `T`.
--- a/content/match-types-spec.md
+++ b/content/match-types-spec.md
@@ -1,0 +1,354 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: implementation
+status: waiting-for-implementation
+title: SIP-56 - Proper Specification for Match Types
+---
+
+**By: Sébastien Doeraene**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Aug 11th 2023 | Initial Draft      |
+
+## Summary
+
+Currently, match type reduction is not specified, and its implementation is by nature not specifiable.
+This is an issue because match type reduction spans across TASTy files (unlike, for example, type inference or GADTs), which can and will lead to old TASTy files not to be linked again in future versions of the compiler.
+
+This SIP proposes a proper specification for match types, which does not involve type inference.
+It is based on `baseType` computations and subtype tests involving only fully-defined types.
+That is future-proof, because `baseType` and subtyping are defined in the specification of the language.
+
+The proposed specification defines a subset of current match types that are considered legal.
+Legal match types use the new, specified reduction rules.
+Illegal match types are rejected, which is a breaking change, and can be recovered under `-source:3.3`.
+
+## Motivation
+
+Currently, match type reduction is implementation-defined.
+Matching a scrutinee type `X` against a pattern `P` with captures `ts` works as follows:
+
+1. we create new type variables for the captures `ts'` giving a pattern `P'`,
+2. we ask the compiler's `TypeComparer` (the type inference black blox) to "try and make it so" that `X <:< P'`,
+3. if it manages to do so, we get constraints for `ts'`; we then have a match, and we instantiate the body of the pattern with the received constraints for `ts`.
+
+The problem with this approach is that, by essence, type inference is an unspecified black box.
+There are *guidelines* about how it should behave in common cases, but no actual guarantee.
+This is fine everywhere else in the language, because what type inference comes up with is stored once and for all in TASTy.
+When we read TASTy files, we do not have to perform the work of type inference again; we reuse what was already computed.
+When a new version of the compiler changes type inference, it does not change what was computed and stored in TASTy files by previous versions of the compiler.
+
+For match types, this is a problem, because reduction spans across TASTy file.
+In order to guarantee compatibility, we must ensure that, for any given match type:
+
+* if it reduces in a given way in verion 1 of the compiler, it still reduces in the same way in version 2, and
+* if it decides disjointness in version 1, it still decides disjointness in version 2.
+
+By delegating reduction to the `TypeComparer` black box, it is in practice impossible to guarantee the former.
+
+In order to solve this problem, this SIP provides a specification for match type reduction that is independent of the `TypeComparer` black box.
+It defines a subset of match type cases that are considered legal.
+Legal cases get a specification for when and how they should reduce for any scrutinee.
+Illegal cases are rejected as being outside of the language.
+For compatibility reasons, they can still be accepted with `-source:3.3`; in that case, they reduce using the existing, unspecified (and prone to breakage) implementation.
+
+For legal cases, the proposed reduction specification should reduce in the same way as the current implementation for the majority of cases.
+That is however not possible to guarantee, since the existing implementation is not specified in the first place.
+
+## Proposed solution
+
+### High-level overview
+
+By its nature, this proposal only contains a specification, without any high level overview.
+
+### Specification
+
+#### Syntax
+
+Syntactically, nothing changes.
+The way that a pattern is parsed and type captures identified is kept as is.
+
+Once type captures are identified, we can represent the *abstract* syntax of a pattern as follows:
+
+```
+MatchTypePattern ::= TypeWithoutCapture
+                   | MatchTypeAppliedPattern
+
+MatchTypeAppliedPattern ::= TyconWithoutCapture ‘[‘ MatchTypeSubPattern { ‘,‘ MatchTypeSubPattern } ‘]‘
+
+MatchTypeSubPattern ::= TypeCapture
+                      | TypeWithoutCapture
+                      | MatchTypeAppliedPattern
+
+TypeCapture ::= NamedTypeCapture
+              | WildcardTypeCapture
+```
+
+The cases `MatchTypeAppliedPattern` are only chosen if they contain at least one `TypeCapture`.
+Otherwise, they are considered `TypeWithoutCapture` instead.
+Each named capture appears exactly once.
+
+#### Legal patterns
+
+A `MatchTypePattern` is legal if and only if one of the following is true:
+
+* It is a `TypeWithoutCapture`, or
+* It is a `MatchTypeAppliedPattern` with a legal `TyconWithoutCapture` and each of its arguments is either:
+  * A `TypeCapture`, or
+  * A `TypeWithoutCapture`, or
+  * The type constructor is *covariant* in that parameter, and the argument is recursively a legal `MatchTypeAppliedPattern`.
+
+A `TyconWithoutCapture` is legal if one of the following is true:
+
+* It is a *class* type constructor, or
+* It is the `scala.compiletime.ops.int.S` type constructor, or
+* It is an *abstract* type constructor, or
+* It is a refined type of the form `Base { type Y = t }` where:
+  * `Base` is a `TypeWithoutCapture`,
+  * There exists a type member `Y` in `Base`, and
+  * `t` is a `TypeCapture`.
+* It is a type alias to a type lambda such that:
+  * Its bounds contain all possible values of its arguments, and
+  * When applied to the type arguments, it beta-reduces to a new legal `MatchTypeAppliedPattern` that contains exactly one instance of every type capture present in the type arguments.
+
+#### Examples of legal patterns
+
+Given the following definitions:
+
+```scala
+class Inv[A]
+class Cov[+A]
+class Contra[-A]
+
+class Base {
+  type Y
+}
+
+type YExtractor[t] = Base { type Y = t }
+type ZExtractor[t] = Base { type Z = t }
+
+type IsSeq[t <: Seq[Any]] = t
+```
+
+Here are example of legal patterns:
+
+```scala
+// TypeWithoutCapture's
+case Any => // also the desugaring of `case _ =>` when the _ is at the top-level
+case Int =>
+case List[Int] =>
+case Array[String] =>
+
+// Class type constructors with direct captures
+case scala.collection.immutable.List[t] => // not Predef.List; it is a type alias
+case Array[t] =>
+case Contra[t] =>
+case Either[s, t] =>
+case Either[s, Contra[Int]] =>
+case h *: t =>
+case Int *: t =>
+
+// The S type constructor
+case S[n] =>
+
+// An abstract type constructor
+// given a [F[_]] or `type F[_] >: L <: H` in scope
+case F[t] =>
+
+// Nested captures in covariant position
+case Cov[Inv[t]] =>
+case Cov[Cov[t]] =>
+case Cov[Contra[t]] =>
+case Array[h] *: t => // sugar for *:[Array[h], t]
+case g *: h *: EmptyTuple =>
+
+// Type aliases
+case List[t] => // which is Predef.List, itself defined as `type List[+A] = scala.collection.immutable.List[A]`
+
+// Refinements (through a type alias)
+case YExtractor[t] =>
+```
+
+The following patterns are *not* legal:
+
+```scala
+// Type capture nested two levels below a non-covariant type constructor
+case Inv[Cov[t]] =>
+case Inv[Inv[t]] =>
+case Contra[Cov[t]] =>
+
+// Type constructor with bounds that do not contain all possible instantiations
+case IsSeq[t] =>
+
+// Type refinement where the refined type member is not a member of the parent
+case ZExtractor[t] =>
+```
+
+#### Matching
+
+Given a scrutinee `X` and a match type case `case P => R` with type captures `ts`, matching proceeds in three steps:
+
+1. Compute instantiations for the type captures `ts'`, and check that they are *specific* enough.
+2. If successful, check that `X <:< [ts := ts']P`.
+3. If successful, reduce to `[ts := ts']R`.
+
+The instantiations are computed by the recursive function `matchPattern(X, P, variance, scrutIsWidenedAbstract)`.
+At the top level, `variance = 1` and `scrutIsWidenedAbstract = false`.
+
+`matchPattern` behaves according to what kind is `P`:
+
+* If `P` is a `TypeWithoutCapture`:
+  * Do nothing (always succeed).
+* If `P` is a `WildcardCapture` `ti = _`:
+  * If `X` is of the form `_ >: L <: H`, instantiate `ti := H` (anything between `L` and `H` would work here),
+  * Otherwise, instantiate `ti := X`.
+* If `P` is a `TypeCapture` `ti`:
+  * If `X` is of the form `_ >: L <: H`,
+    * If `scrutIsWidenedAbstract` is `true`, fail as not specific.
+    * Otherwise, if `variance = 1`, instantiate `ti := H`.
+    * Otherwise, if `variance = -1`, instantiate `ti := L`.
+    * Otherwise, fail as not specific.
+  * Otherwise, if `variance = 0` or `scrutIsWidenedAbstract` is `false`, instantiate `ti := X`.
+  * Otherwise, fail as not specific.
+* If `P` is a `MatchTypeAppliedPattern` of the form `T[Qs]`:
+  * Assert: `variance = 1` (from the definition of legal patterns).
+  * If `T` is a class type constructor of the form `p.C`:
+    * If `baseType(X, C)` is not defined, fail as not matching.
+    * Otherwise, it is of the form `q.C[Us]`.
+    * If `p =:= q` is false, fail as not matching.
+    * Let `innerScrutIsWidenedAbstract` be true if either `scrutIsWidenedAbstract` or `X` is not a concrete type.
+    * For each pair of `(Ui, Qi)`, compute `matchPattern(Ui, Qi, vi, innerScrutIsWidenedAbstract)` where `vi` is the variance of the `i`th type parameter of `T`.
+  * If `T` is `scala.compiletime.ops.int.S`:
+    * If `n = natValue(X)` is undefined or is `Int.MinValue`, fail as not matching.
+    * Otherwise, compute `matchPattern(n, Q1, 1, scrutIsWidenedAbstract)`.
+  * If `T` is an abstract type constructor:
+    * If `X` is not of the form `F[Us]` or `F =:= T` is false, fail as not matching.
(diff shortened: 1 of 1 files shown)


content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+Legal match types use the new, specified reduction rules.
+Illegal match types are rejected, which is a breaking change, and can be recovered under `-source:3.3`.
+
+## Motivation
+
+Currently, match type reduction is implementation-defined.
+Matching a scrutinee type `X` against a pattern `P` with captures `ts` works as follows:
+
+1. we create new type variables for the captures `ts'` giving a pattern `P'`,
+2. we ask the compiler's `TypeComparer` (the type inference black blox) to "try and make it so" that `X <:< P'`,
```
SethTisue, line 36:
```suggestion
2. we ask the compiler's `TypeComparer` (the type inference black box) to "try and make it so" that `X <:< P'`,
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+3. if it manages to do so, we get constraints for `ts'`; we then have a match, and we instantiate the body of the pattern with the received constraints for `ts`.
+
+The problem with this approach is that, by essence, type inference is an unspecified black box.
+There are *guidelines* about how it should behave in common cases, but no actual guarantee.
+This is fine everywhere else in the language, because what type inference comes up with is stored once and for all in TASTy.
+When we read TASTy files, we do not have to perform the work of type inference again; we reuse what was already computed.
+When a new version of the compiler changes type inference, it does not change what was computed and stored in TASTy files by previous versions of the compiler.
+
+For match types, this is a problem, because reduction spans across TASTy file.
```
SethTisue, line 45: > reduction spans across TASTy file

unclear, to me anyway — can you reword/expand to be clearer?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+In order to guarantee compatibility, we must ensure that, for any given match type:
+
+* if it reduces in a given way in verion 1 of the compiler, it still reduces in the same way in version 2, and
```
SethTisue, line 48:
```suggestion
* if it reduces in a given way in version 1 of the compiler, it still reduces in the same way in version 2, and
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+* if it decides disjointness in version 1, it still decides disjointness in version 2.
```
SethTisue, line 49: terse to the point of being a bit cryptic. maybe:

```suggestion
* if it judges two types to be disjoint in version 1, it still makes the same judgment in version 2.
```

but I'm not sure what to suggest because I actually don't know what shade of meaning is intended

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+By delegating reduction to the `TypeComparer` black box, it is in practice impossible to guarantee the former.
```
SethTisue, line 51:
```suggestion
If reduction is delegated to the `TypeComparer` black box, then it is impossible in practice to guarantee the first property.
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+In order to solve this problem, this SIP provides a specification for match type reduction that is independent of the `TypeComparer` black box.
+It defines a subset of match type cases that are considered legal.
+Legal cases get a specification for when and how they should reduce for any scrutinee.
+Illegal cases are rejected as being outside of the language.
+For compatibility reasons, they can still be accepted with `-source:3.3`; in that case, they reduce using the existing, unspecified (and prone to breakage) implementation.
```
SethTisue, line 57:
```suggestion
For compatibility reasons, such now-illegal cases will still be accepted under `-source:3.3`; in that case, they reduce using the existing, unspecified (and prone to breakage) implementation.
```

SethTisue: also, I think it would be good to say something about how long `-source:3.3` might be expected to exist. (Is the “LTS” status of 3.3 relevant...?)

Also, if `-source:3.3` enabled, will such uses still generate warnings, or does that flag flip it directly all the way from "doesn't compile at all" to "compiles it without complaint"?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+For legal cases, the proposed reduction specification should reduce in the same way as the current implementation for the majority of cases.
```
SethTisue, line 59: “majority” seems like a pretty weak claim. so, maybe only 51% of cases? I would hope you would be comfortable making a stronger claim such as “the vast majority” or “nearly all”

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+That is however not possible to guarantee, since the existing implementation is not specified in the first place.
+
+## Proposed solution
+
+### High-level overview
+
+By its nature, this proposal only contains a specification, without any high level overview.
```
SethTisue, line 66: huh? I don't understand this at all. am I missing some context?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+That is future-proof, because `baseType` and subtyping are defined in the specification of the language.
+
+The proposed specification defines a subset of current match types that are considered legal.
+Legal match types use the new, specified reduction rules.
+Illegal match types are rejected, which is a breaking change, and can be recovered under `-source:3.3`.
+
+## Motivation
+
+Currently, match type reduction is implementation-defined.
+Matching a scrutinee type `X` against a pattern `P` with captures `ts` works as follows:
```
SethTisue, line 33: I think it would help many readers to explain at least briefly (and/or with a link) what captures are. Does this have to do with “capture checking”, or is it something else?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+By its nature, this proposal only contains a specification, without any high level overview.
+
+### Specification
+
+#### Syntax
+
+Syntactically, nothing changes.
+The way that a pattern is parsed and type captures identified is kept as is.
+
+Once type captures are identified, we can represent the *abstract* syntax of a pattern as follows:
```
SethTisue, line 75: Why the emphasis on “abstract”? Whatever distinction you're making there, I think you could be more explicitly about what the distinction is and why you're making it here.

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+### Compatibility
+
+Compatibility is inherently tricky to evaluate for this proposal, and even to define.
+One could argue that, from a pure specification point of view, it breaks nothing since it only specifies things that were unspecified before.
+However, that is not very practical.
+In practice, this proposal definitely breaks some code that compiled before, due to making some patterns illegal.
+In exchange, it promises that all the patterns that are considered legal will keep working in the future; which is not the case with the current implementation, even for the legal subset.
+
+In order to evaluate the practical impact of this proposal, we conducted a quantitative analysis of *all* the match types found in Scala 3 libraries published on Maven Central.
+We used [Scaladex](https://index.scala-lang.org/) to list all Scala 3 libraries, [coursier](https://get-coursier.io/docs/api) to resolve their classpaths, and [tasty-query](https://github.com/scalacenter/tasty-query) to semantically analyze the patterns of all the match types they contain.
```
SethTisue, line 256: 💪 👏

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+Out of 4,783 libraries that were found and analyzed, 49 contained at least one match type definition.
```
SethTisue, line 258:
```suggestion
Out of 4,783 libraries, 49 contained at least one match type definition.
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+* 1 library used to have 1 `&`-type extractor (which "worked" who knows how?):
+  https://github.com/Katrix/perspective/blob/f1643ac7a4e6a0d8b43546bf7b9e6219cc680dde/dotty/derivation/src/main/scala/perspective/derivation/Helpers.scala#L15-L18
+  but the author already accepted a change with a workaround at
+  https://github.com/Katrix/perspective/pull/1
+* 1 library has 3 occurrences of using an abstract type constructor too "concretely":
+  https://github.com/kory33/s2mc-test/blob/d27c6e85ad292f8a96d7d51af7ddc87518915149/protocol-core/src/main/scala/io/github/kory33/s2mctest/core/generic/compiletime/Tuple.scala#L16
+  defined at https://github.com/kory33/s2mc-test/blob/d27c6e85ad292f8a96d7d51af7ddc87518915149/protocol-core/src/main/scala/io/github/kory33/s2mctest/core/generic/compiletime/Generic.scala#L12
+  It could be replaced by a concrete `class Lock[A](phantom: A)` instead.
+
+The only case for which there exists no workaround that we know of is the extractor for `scala.Enumeration`-based `Value` classes.
```
SethTisue, line 278: So. Should we be concerned? You think this... entirely acceptable, probably acceptable, probably a blocker that needs to be resolved?

I am curious what your own opinion is, but I'm also interested in having a bit more assistance in formulating my own opinion. What is this, what is the fundamental issue here?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+None at this point.
+
+## Alternatives
+
+The specification is more complicated than we initially wanted.
+At the beginning, we were hoping that we could restrict match cases to class type constructors only.
+The quantitative study however revealed that we had to introduce support for abstract type constructors and for type member extractors.
+
+As already mentioned, the standard library itself contains an occurrence of an abstract type constructor in a pattern.
+Making that an error would mean declaring the standard library itself bankrupt, which was not a viable option.
```
SethTisue, line 330: “bankrupt” is a colorful metaphor here, but I'm left wondering what underlying literal meaning is intended. like, we'd need to make an incompatible change to the standard library that would break some code? (how much code? compile-time breakage, runtime breakage?) is making such a change impossible? highly undesirable? at least somewhat undesirable?

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+We tried to restrict abstract type constructor to never match on their own.
```
SethTisue, line 332:
```suggestion
We tried to restrict abstract type constructors to never match on their own.
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+Instead, we wanted them to stay *stuck* until they could be instantiated to a concrete type constructor.
```
SethTisue, line 333:
```suggestion
Instead, we wanted them to stay “stuck” until they could be instantiated to a concrete type constructor.
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+However, that led some existing tests to fail even for match types that were declared legal, because they did not reduce anymore in some places where they reduced before.
+
+Type member extractors are our biggest pain point.
+Their specification is complicated, and the implementation as well.
+Our quantitative study showed that they were however "often" used (10 occurrences spread over 4 libraries).
```
SethTisue, line 338:
```suggestion
Our quantitative study showed that they were however used at least somewhat often (10 occurrences spread over 4 libraries).
```

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+In each case, they seem to be a way to express what Scala 2 type projections (`A#T`) could express.
+While not quite as powerful as type projections (which were shown to be unsound), match types with type member extractors delay things enough for actual use cases to be meaningful.
```
SethTisue, line 340: @dwjinand you might have something to say about this? I know it's come up repeatedly in our discussions about match types that it's possible to use them to smuggle in something akin to abstract type projections.

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+
+As far as we know, those use cases have no workaround if we make type member extractors illegal.
+
+## Related work
+
+This section should list prior work related to the proposal, notably:
```
SethTisue, line 346:
```suggestion
Notable prior work related to this proposal includes:
```

soronpo:
I gave **only** the specification part to GPT-4 for review. Here is what it mentioned:
The provided specification for the proposed SIP regarding Match Types in Scala 3 is thorough and quite detailed. However, based on the text you've provided, here are a few observations and potential issues:

1. **Clarity on Syntactical Components**:
    - When you mention various constructs like `TypeCapture`, `TypeWithoutCapture`, etc., it would be helpful to provide some simple examples or even a brief definition for each. This ensures that the reader is clear on what constitutes each of these syntactical elements.
  
2. **Nested Captures in Legal Patterns**:
    - The specification mentions that the following is not a legal pattern: `case Inv[Cov[t]] =>` because a type capture is nested two levels below a non-covariant type constructor. However, it's not immediately obvious why this is illegal based solely on the legal pattern criteria provided. An elaboration on this could be beneficial.

3. **Concrete Vs. Abstract**:
    - There's mention of checking whether a type is 'concrete' versus 'abstract', especially in the matching section. It might be helpful to clarify how the proposal differentiates between these two.

4. **The Matching Process**:
    - The matching procedure is explained well in terms of steps and conditions. Still, due to its complexity, adding a visual representation, like a flowchart, might help in better understanding the flow of the process.
  
5. **Refined Type Constructs**:
    - There's a mention of refined types, like `Base { type Y = t }`. However, an explanation of the intricacies of these refined types, especially within the context of match types, would be beneficial.

6. **Caveats or Limitations**:
    - If there are any caveats or limitations associated with this proposal, it would be beneficial to outline them explicitly. This helps in setting clear expectations for developers using this feature.

7. **Examples**:
    - While there are ample examples given, providing more complex scenarios and edge cases might be useful. This can help in highlighting the intricacies and power of the new match type system. Moreover, a practical use-case of where such a feature would be beneficial in real-world code can be a useful addition.

8. **Performance Implications**:
    - Given the complexity of the match types and the specified algorithm, what are the expected performance implications? Would matching against these types result in a considerable slowdown in compile time?

9. **Error Handling & Messages**:
    - The proposal doesn't talk about the quality or detail of error messages that would be provided when developers write illegal patterns or mismatched types. For a feature of this complexity, clear error messages are crucial.

10. **Disjointness**:
    - The section on disjointness is quite brief. Elaborating more on how provably disjoint types are handled might provide clarity, especially in edge cases.

In summary, while the proposal is detailed and gives a good overview of the changes to Match Types, addressing the above points could make it even clearer and more actionable. It's essential that when proposing such changes, developers at all levels of expertise can understand and work with the feature effectively.

anatoliykmetyuk:
As per the [recent changes](https://github.com/scala/docs.scala-lang/commit/81d3ed7c0d8a1a958e0161f4735b1bbbb4bcaaf9) to the SIP process, each SIP now needs to have a SIP Manager assigned. I propose @bjornregnell to be a manager of this SIP - is that OK with you?

As per the process spec, this role entails responsibility for all the communications around the SIP during its lifecycle, including requesting a vote on the SIP from the whole Committee, presenting the SIP to the Committee at the plenary meetings, merging or closing the corresponding PR, reporting to the community on the vote outcome, and announcing when it is available for testing.

bjornregnell:
@anatoliykmetyuk I'm not sure I'm the best pic for this as I have never seriously tried Match types or any other Scala meta-programming feature so I have some catch up + learning to do in order to write intelligibly about this in a community post etc. 

But if there is no one more knowledgeable in Match types available, I can do it iff @sjrd is willing to give me a beginner-level personal tutorial in Zoom when possible ;)   (It will have to be after the start of the fall semester as I'm currently swamped with teaching preparations...)

content/match-types-spec.md
sjrd: I expanded a bit. Is it better now? Should I go all the way to showing a concrete example?

content/match-types-spec.md
sjrd: I have significantly elaborated those paragraphs. Is it better now?

content/match-types-spec.md
sjrd: Yeah that was overly pessimistic of me. We actually did not find any case where it broke, so I made that clearer.

content/match-types-spec.md
sjrd: There's a section "High-level overview" in the SIP template and I felt compelled to fill it in, but it makes no sense here. The feature already exists and is shipped. I removed the section.

content/match-types-spec.md
sjrd: I added some references to the spec, which has both concrete syntax and abstract syntax for types. We are talking about the abstract syntax here.

content/match-types-spec.md
sjrd: I found a workaround in the meantime, though I had to allow matching class members with type member extractors. I added a commit with those changes.

content/match-types-spec.md
SethTisue: I think so.

content/match-types-spec.md
SethTisue: That seems like a requirement in order to have the "proper specification for match types" the SIP name promises.

anatoliykmetyuk:
@bjornregnell got it, thanks for the heads-up!

@soronpo has agreed to be the SIP Manager, so, in that case, he'll manage this SIP.

content/match-types-spec.md
```diff
@@ -0,0 +1,354 @@
...
+Instead, we wanted them to stay *stuck* until they could be instantiated to a concrete type constructor.
+However, that led some existing tests to fail even for match types that were declared legal, because they did not reduce anymore in some places where they reduced before.
+
+Type member extractors are our biggest pain point.
+Their specification is complicated, and the implementation as well.
+Our quantitative study showed that they were however "often" used (10 occurrences spread over 4 libraries).
+In each case, they seem to be a way to express what Scala 2 type projections (`A#T`) could express.
+While not quite as powerful as type projections (which were shown to be unsound), match types with type member extractors delay things enough for actual use cases to be meaningful.
+
+As far as we know, those use cases have no workaround if we make type member extractors illegal.
```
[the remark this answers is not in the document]
sjrd: Hum no, the sentence says: if we remove type member extractors *from the proposed spec*, then we will have no workaround anymore. With the proposed spec, we can extract type members.

The originally submitted SIP/spec did not allow to extract the type member of `SomeEnumeration#Value` because `Value` is a *class* member (not a type alias), but the commit "Allow class members to be matched by type member extractors." made that actually possible.

content/match-types-spec.md
sjrd: > something akin to abstract type projections

With the proposed spec, only *concrete* type projections actually make it. As long as it's abstract, `ExtractTOf[A]` cannot reduce to `T` like `A#T` does. The difference is that `A#T` has the "best" *bounds* of `T` in `A`, whereas `ExtractTOf[A]` has only the match type's bound as long as `T` is not concrete (through subsequent substitution of `A` by something more precise in which `T` is a type alias whose rhs does not depend on its prefix).

For example,
```scala
class Base {
  type T
  def someT: T = ???
}

type ExtractTOfAux[A] = Base { type T = A }
type ExtractTOf[X <: Base] = X match
  case ExtractTOfAux[t] => t

// Scala 2 abstract type projection, works
def fooOfBaseProj[B <: Base](b: B): B#T = b.someT

// Scala 3 match type extractor: does not work because the match type does not reduce
// as T is not concrete in B.
def fooOfBaseMT[B <: Base](b: B): ExtractTOf[B] = b.someT
```

The match-type-based type member extractor still features enough power because you can "carry around" values of that type, and at the edge your `B` becomes concrete enough that you can reduce. This is what happens with `scala.Enumeration`s.

IMO, people should not use that. It's all a hack. Instead, we should use path-dependent types with actual values of the enclosing class. For example the proper definition of `fooOfBase` is:
```scala
def fooOfBasePathDep(b: Base): b.T = b.someT
```
Unfortunately, people have been using type projections/type member extractors instead. Yes, sometimes it avoids one extra layer of nesting: if you need a class that `extends Foo[b.type]`, you would `b` to be declared one level up, and so that class needs to be nested in another class that has a `b` field. This is why the json4s workaround is what it is, instead of turning it into path-dependent types. The correct solution would be to have
```scala
class EnumSerializers(val enumeration: Enumeration) {
  class EnumSerializer extends Serializer(enumeration.Value) {
    ...
  }
}
```
but that would have changed the public API, so it was not an acceptable PR to make.

content/match-types-spec.md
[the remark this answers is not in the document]
bjornregnell: OK thanks for the clarification!

content/match-types-spec.md
```diff
@@ -0,0 +1,399 @@
...
+* It is a `TypeWithoutCapture`, or
+* It is a `MatchTypeAppliedPattern` with a legal `TyconWithoutCapture` and each of its arguments is either:
+  * A `TypeCapture`, or
+  * A `TypeWithoutCapture`, or
+  * The type constructor is *covariant* in that parameter, and the argument is recursively a legal `MatchTypeAppliedPattern`.
+
+A `TyconWithoutCapture` is legal if one of the following is true:
+
+* It is a *class* type constructor, or
+* It is the `scala.compiletime.ops.int.S` type constructor, or
```
bishabosha, line 147: shouldn't this also cover `compiletime.ops.long.S` [Source code](https://github.com/lampepfl/dotty/blob/3.3.0/library/src/scala/compiletime/ops/long.scala#L18)

sjrd: Perhaps, if we ignore the fact that [scala.compiletime.ops.long.S doesn't work](https://github.com/lampepfl/dotty/issues/18342) :man_shrugging:

content/match-types-spec.md
benhutchison: 👏

sjrd:
Well, it took me quite a while to get to the bottom of it, but I added a proposed specification for the provably-disjoint test, as requested. Note that in the end, I actually *do* recommend that we change it compared to the current implementation. I have a PR for those additional changes at https://github.com/lampepfl/dotty/pull/18416

content/match-types-spec.md
sjrd: `long.S` was [officially deprecated](https://github.com/lampepfl/dotty/pull/18426), so this is not an issue anymore.

sjrd:
I have polished the proposal a bit further with type lambda support in `provablyDisjoint`, and with a spec for the full reduction algorithm as well.

content/match-types-spec.md
```diff
@@ -0,0 +1,579 @@
...
+* Singleton paths to distinct `enum` case values are nonintersecting
+
+However, a precise definition of provably-disjoint is complicated and requires some helpers.
+We start with the notion of "simple types", which are a minimal subset of Scala types that capture the concepts mentioned above.
+
+A "simple type" is one of:
+
+* `Nothing`
+* `AnyKind`
+* `p.C[...Xi]` a possibly parameterized class type, where `p` and `...Xi` are arbitrary types (not just simple types)
```
bishabosha, line 332: I propose to treat `scala.compiletime.ops.int.S` as a parameterised class type for purposes of disjointness. Then will it will be considered disjoint to `0` and any other negative integer literal?

sjrd: I don't think we can pretend that it is a parameterized class type. It is not. For one thing, its existence would invalidate the `final` status of `scala.Int`, which then prevents `Int` from being disjoint from other traits, although currently it is.

If we want to add `S[n]` here, it needs to have its own category, and all the rules (and proof) need to be adapted to account for it.

dwijnand:
@sjrd or @bishabosha could you add `AnyKind` to this, please? I.e. #18510

content/match-types-spec.md
```diff
@@ -0,0 +1,579 @@
...
+* `c` a literal type
+* `p.C.x` where `C` is an `enum` class and `x` is one of its value `case`s
+* `X₁ & X₂` where `X₁` and `X₂` are both simple types
+* `X₁ | X₂` where `X₁` and `X₂` are both simple types
+* `[...ai] =>> X₁` where `X₁` is a simple type
+
+We define `⌈X⌉` a function from a full Scala type to a simple type.
```
dwijnand, line 339: How do you say `⌈`?

content/match-types-spec.md
```diff
@@ -0,0 +1,579 @@
...
+
+The following properties hold about `⌈X⌉` (we have paper proofs for those):
+
+* `X <: ⌈X⌉` for all type `X`.
+* If `S <: T`, and the subtyping derivation does not use the "lower-bound rule" of `<:` anywhere, then `⌈S⌉ <: ⌈T⌉`.
+
+The "lower-bound rule" states that `S <: T` if `T = q.X `and `q.X` is a non-class type designator and `S <: L` where `L` is the lower bound of the underlying type definition of `q.X`".
+That rule is known to break transitivy of subtyping in Scala already.
+
+Second, we define the relation `⋔` on *classes* (including traits and hidden classes of objects) as:
```
dwijnand, line 366: "hidden classes of objects" means module classes?

content/match-types-spec.md
```diff
@@ -0,0 +1,579 @@
...
+  * `p.C[...Si] ⋔ q.D.y` if `baseType(q.D.y, C)` is not defined
+  * `c ⋔ q.D[...Ti]` if `baseType(c, D)` is not defined
+  * `p.C[...Si] ⋔ d` if `baseType(d, C)` is not defined
+* Two class types are disjoint if the classes themselves are disjoint, or if there exist a common super type with conflicting type arguments.
+  * `p.C[...Si] ⋔ q.D[...Ti]` if `C ⋔ D`
+  * `p.C[...Si] ⋔ q.D[...Ti]` if there exists a class `E` such that `baseType(p.C[...Si], E) = a.E[...Ai]` and `baseType(q.D[...Ti], E) = b.E[...Bi]` and there exists a pair `(Ai, Bi)` such that
+    * `Ai ⋔ Bi` and it is in invariant position, or
+    * `Ai ⋔ Bi` and it is in covariant position and there exists a field of that type parameter in `E`
+
+It is worth noting that this definition disregards prefixes entirely.
```
dwijnand, line 418: Disregards _differing_ prefixes?  Just to be clear that not all prefixes are disregarded.

content/match-types-spec.md
sjrd: Its real name can probably be "ceil", which is its name in math as the approximation towards positive infinity.

I call it "up" in my head, for some reason.

content/match-types-spec.md
sjrd: Yes.

sjrd:
> @sjrd or @bishabosha could you add `AnyKind` to this, please? I.e. [lampepfl/dotty#18510](https://github.com/lampepfl/dotty/pull/18510)

What do you think is missing? It's already in the spec and in the implementation for this SIP.

dwijnand:
> > @sjrd or @bishabosha could you add `AnyKind` to this, please? I.e. [lampepfl/dotty#18510](https://github.com/lampepfl/dotty/pull/18510)
> 
> What do you think is missing? It's already in the spec and in the implementation for this SIP. (Note that in the spec, `Nothing` and `AnyKind` are not _classes_; they are special types.)

That any type and `AnyKind` are never disjoint, which it is in the current implementation and that I'm fixing in my PR.  If that's what's already said in this spec, then nothing more.

sjrd:
The `provablyDisjoint` (`⋔`) relation is defined in terms of all the possible cases that make it true. When one of the sides is `AnyKind`, the only possible rules that can apply are `Nothing ⋔ T` and `S ⋔ Nothing`, as well as rules on union/intersection types, which recurse. So `AnyKind ⋔ Nothing` and conversely, but no other type can ever be provably disjoint from `AnyKind`. There is no need for a rule that says something about when `provablyDisjoint` is *not* true.

soronpo:
@bjornregnell @Kordyjan Do you agree to request a vote in our next SIP meeting? If yes, what would be your recommendation?

bjornregnell:
> Do you agree to request a vote in our next SIP meeting? If yes, what would be your recommendation?

Yes, I recommend to accept iff @sjrd is happy with the review feedback given so far and feel fairly certain that the spec does not contain major problems that cannot be fixed later. I guess we are voting for it to "experimental" and that we later vote for it be be accepted as "stable" when a tested implementation is available.

I have read through the proposal and find the motivation for the need of a spec very well written and understandable. However, my knowledge in the underlying theory, type systems. meta-programming, etc, is too limited to give deeper feedback to the actual specification, esp. the section "Matching". 

If I mark the entire natural language bullet list after "matchPattern behaves according to what kind is P:" and copy-paste it into a `val s = """..."""` and calculate
```
scala> s.split("\n").map(_.trim).filter(_.startsWith("If ")).length
```
then I get 21 (sic!) "If "-clauses that are up for review. I humbly wonder if there is some other human with at least the knowledge of @sjrd that have double check each rule for omissions or bugs? 

I *have* read all 21 of them and "LGTM", but I cannot say that I am able to check them, although they look reasonable to me on a surface-understanding level.  

Also, perhaps it could be explicitly stated in the SIP preamble that this proposed spec is the source of a complete re-implementation from scratch of this language feature in a coming compiler version (if that is the case (?) - just to make it clear upfront). 

(And then I wonder how to verify that the new implementation complies with this spec - but that is a matter of how to implement this and how to accept is as "stable" in a later stage...)

Anyway, thank you **very much** indeed @sjrd for embarking on this seemingly very intricate and tedious but also very important work.

bjornregnell:
(And I guess that when implementing this in the compiler based on the spec, omissions and bugs in the spec are likely to be found, if any.)

sjrd:
> I humbly wonder if there is some other human with at least the knowledge of @sjrd that have double-checked each rule for omissions or bugs?

FWIW, there is at least one other human: Olivier Blanvillain, who created match types in the first place.

I hoped there would be fewer "If"s. Unfortunately, it seems each one of them is necessary to have something complete enough. :man_shrugging: 

> (And I guess that when implementing this in the compiler based on the spec, omissions and bugs in the spec are likely to be found, if any.)

There is already an implementation at https://github.com/lampepfl/dotty/pull/18262, which indeed, discovered omissions in the early versions of the spec. The latest spec is up-to-date with that implementation, and conversely. The implementation passes the (small) community build, but we have not tried it yet on the open community build.

bjornregnell:
> The implementation passes the (small) community build, but we have not tried it yet on the open community build.

OK great. Perhaps each "If" can be formulated as a test case in the code base?

bjornregnell:
I looked at https://github.com/lampepfl/dotty/pull/18262 and indeed there are tests, but I could not trace them to the rules in the spec - would that kind of traceability be helpful? @sjrd

bjornregnell:
Perhaps the rules in the spec should be numbered?

content/match-types-spec.md
```diff
@@ -0,0 +1,579 @@
...
+class Contra[-A]
+
+class Base {
+  type Y
+}
+
+type YExtractor[t] = Base { type Y = t }
+type ZExtractor[t] = Base { type Z = t }
+
+type IsSeq[t <: Seq[Any]] = t
```
joroKr21, line 183: How do we express something similar with the new algorithm?

PR #67, opened 2023-12-08
SIP-57 - Replace non-sensical @unchecked annotations

bishabosha:
Currently, a user can silence warnings that a scrutinee may not be matched by a pattern, by annotating the scrutinee with the `@unchecked` annotation.

This SIP proposes to use a new annotation `@RuntimeCheck` to replace `@unchecked` for this purpose. For convenience, an extension method will be added to `Predef` that marks the receiver with the annotation (used as follows: `foo.runtimeCheck`).

Functionally it behaves the same as the old annotation, but improves readability at the callsite.

### DIFF
--- a/content/replace-nonsensical-unchecked-annotation.md
+++ b/content/replace-nonsensical-unchecked-annotation.md
@@ -4,7 +4,7 @@ permalink: /sips/:title.html
 stage: implementation
 status: waiting-for-implementation
 presip-thread: https://contributors.scala-lang.org/t/pre-sip-replace-non-sensical-unchecked-annotations/6342
-title: SIP-NN - Replace non-sensical @unchecked annotations
+title: SIP-57 - Replace non-sensical @unchecked annotations
 ---
 
 **By: Martin Odersky and Jamie Thompson**
--- a/content/replace-nonsensical-unchecked-annotation.md
+++ b/content/replace-nonsensical-unchecked-annotation.md
@@ -1,0 +1,176 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: implementation
+status: waiting-for-implementation
+presip-thread: https://contributors.scala-lang.org/t/pre-sip-replace-non-sensical-unchecked-annotations/6342
+title: SIP-NN - Replace non-sensical @unchecked annotations
+---
+
+**By: Martin Odersky and Jamie Thompson**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Dec 8th 2023  | Initial Draft      |
+
+## Summary
+
+We propose to replace the mechanism to silence warnings for "unchecked" patterns, in the cases where silencing the warning will still result in the pattern being checked at runtime.
+
+Currently, a user can silence warnings that a scrutinee may not be matched by a pattern, by annotating the scrutinee with the `@unchecked` annotation. This SIP proposes to use a new annotation `@RuntimeCheck` to replace `@unchecked` for this purpose. For convenience, an extension method will be added to `Predef` that marks the receiver with the annotation (used as follows: `foo.runtimeCheck`). Functionally it behaves the same as the old annotation, but improves readability at the callsite.
+
+## Motivation
+
+As described in [Scala 3 Reference: Pattern Bindings](https://docs.scala-lang.org/scala3/reference/changed-features/pattern-bindings.html), under `-source:future` it is an error for a pattern definition to be refutable. For instance, consider:
+```scala
+def xs: List[Any] = ???
+val y :: ys = xs
+```
+
+This compiled without warning in 3.0, became a warning in 3.2, and we would like to make it an error by default in a future 3.x version.
+As an escape hatch we recommend to use `@unchecked`:
+```
+-- Warning: ../../new/test.scala:6:16 ------------------------------------------
+6 |  val y :: ys = xs
+  |                ^^
+  |pattern's type ::[Any] is more specialized than the right hand side expression's type List[Any]
+  |
+  |If the narrowing is intentional, this can be communicated by adding `: @unchecked` after the expression,
+  |which may result in a MatchError at runtime.
+```
+Similarly for non-exhaustive `match` expressions, where we also recommend to put `@unchecked` on the scrutinee.
+
+But `@unchecked` has several problems. First, it is ergonomically bad. For instance to fix the exhaustivity warning in
+```scala
+xs match
+  case y :: ys => ...
+```
+we'd have to write
+```
+(xs: @unchecked) match
+  case y :: ys => ...
+```
+Having to wrap the `@unchecked` in parentheses requires editing in two places, and arguably harms readability: both due to the churn in extra symbols, and because in this use case the `@unchecked` annotation poorly communicates intent.
+
+Nominally, the purpose of the annotation is to silence warnings (_from the [API docs](https://www.scala-lang.org/api/3.3.1/scala/unchecked.html#)_):
+> An annotation to designate that the annotated entity should not be considered for additional compiler checks.
+
+
+
+In the following code however, the word `unchecked` is a misnomer, so could be confused for another meaning by an inexperienced user:
+
+```scala
+def xs: List[Any] = ???
+val y :: ys = xs: @unchecked
+```
+ After all, the pattern `y :: ys` _is_ checked, but it is done at runtime (by looking at the runtime class), rather than statically.
+
+As a direct contradiction, in the following usage of `unchecked`, the meaning is the opposite:
+```scala
+xs match
+  case ints: List[Int @unchecked] =>
+```
+Here, `@unchecked` means that the `Int` parameter will _not_ be checked at runtime: The compiler instead trusts the user that `ints` is a `List[Int]`. This could lead to a `ClassCastException` in an unrelated piece of code that uses `ints`, possibly without leaving a clear breadcrumb trail of where the faulty cast originally occurred.
+
+## Proposed solution
+
+### High-level overview
+
+This SIP proposes to fix the ergnomics and readability of `@unchecked` in the usage where it means "checked at runtime", by instead adding a new annotation `scala.internal.RuntimeCheck`.
+
+```scala
+package scala.annotation.internal
+
+final class RuntimeCheck extends Annotation
+```
+
+In all usages where the compiler looks for `@unchecked` for this purpose, we instead change to look for `@RuntimeCheck`.
+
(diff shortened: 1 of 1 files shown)


content/replace-nonsensical-unchecked-annotation.md
```diff
@@ -0,0 +1,176 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: implementation
+status: waiting-for-implementation
+presip-thread: https://contributors.scala-lang.org/t/pre-sip-replace-non-sensical-unchecked-annotations/6342
+title: SIP-NN - Replace non-sensical @unchecked annotations
```
anatoliykmetyuk, line 7:
```suggestion
title: SIP-57 - Replace non-sensical @unchecked annotations
```

JD557:
A bit of bike shedding, but all annotations that I can find in `scala.annotation` seem to use lower camel case. 
Is there any particular reason for this one to use upper camel case?

odersky:
It's an `internal` annotations. Other annotations in `internal` also start with an upper case letter, e.g. `Child`, `Body`, `InlineParam`.

bishabosha:
I will need to add an amendment that we would also like to change the following behavior:
```scala
scala> val xs = List(1: Any)
val xs: List[Any] = List(1)
                                                                                                    
scala> xs match {
     |   case is: ::[Int] => is.head
     | }
2 warnings found
-- [E029] Pattern Match Exhaustivity Warning: ----------------------------------
1 |xs match {
  |^^
  |match may not be exhaustive.
  |
  |It would fail on pattern case: List(_, _*), Nil
  |
val res11: Int = 1
-- Unchecked Warning: ----------------------------------------------------------
2 |  case is: ::[Int] => is.head
  |       ^
  |the type test for ::[Int] cannot be checked at runtime because its type arguments can't be determined from List[Any]
```
when we try again with the `@unchecked` on the scrutinee, all the warnings disappear
```scala
scala> (xs: @unchecked) match {
     |   case is: ::[Int] => is.head
     | }
val res12: Int = 1
```
with `xs.runtimeCheck` we should still produce an unchecked warning for `case is: ::[Int] =>`
```scala
scala> xs.runtimeChecked match {
     |   case is: ::[Int] => is.head
     | }
1 warning found
-- Unchecked Warning: ----------------------------------------------------------
2 |  case is: ::[Int] => is.head
  |       ^
  |the type test for ::[Int] cannot be checked at runtime because its type arguments can't be determined from List[Any]
val res13: Int = 1
```

chrisandrews-ms:
_(I'm an assigned reviewer and manager for this SIP)_ 

I think purpose of this SIP is clearly explained and valid, and the proposed solution is good. The name `runtimeChecked` is also natural to read and captures the behavior well. It's a little bit long to _write_ but that's probably a good thing.

@sjrd my understanding is that your [concerns about breaking macros](https://contributors.scala-lang.org/t/pre-sip-replace-non-sensical-unchecked-annotations/6342/82) are resolved by the `transparent inline` approach, because macros will now see the annotation rather than the method call, and will hopefully leave it well alone. But please let us know if that isn't the case.

@Kordyjan as the other assigned reviewer, do you have any thoughts?

Unless someone raises new concerns, I'm inclined to vote to approve this proposal at the next meeting.

lrytz:
In Scala 3, `@unchecked` on the scrutinee seems to also disable reachability

```scala
// Scala 3
scala> def f(x: Any) = (x: @unchecked) match { case Some(_) => 0; case Some(1) => 1 } // no warning
def f(x: Any): Int
```

Would that remain the same?

Note that's different in Scala 2 where `@unchecked` only affects exhaustivity.

@bishabosha regarding this comment https://github.com/scala/improvement-proposals/pull/67#issuecomment-1857994355, in Scala 2 `@unchecked` on the scrutinee doesn't silence unchecked type arguments:

```scala
// Scala 2
scala> def f(xs: List[Any]) = (xs: @unchecked) match { case is: ::[Int] => is.head }
                                                                ^
       warning: non-variable type argument Int in type pattern scala.collection.immutable.::[Int] (the underlying of ::[Int]) is unchecked since it is eliminated by erasure
def f(xs: List[Any]): Int
```

Maybe that can be considered a bug in Scala 3.

lrytz:
To someone unfamiliar, `x match` vs `x.runtimeCheck match` might look like some check is introduced, or maybe some check is delayed from compiletime to runtime; either way it looks like something is different at runtime.

But this is not the case. What it's doing is disabling compiletime exhaustivity checking. The warning for a non-exhaustive match says "match may not be exhaustive", so "exhaustive" is already a term that users need to learn. Can we use it in the new syntax introduced by this SIP, e.g. `x.nonExhaustive match { ... }`?

odersky:
But the check is delayed, no? If we have a selector without a matching case, we get a `MatchError` at runtime. So disabling exhaustivity checking reveals a runtime check.

lrytz:
It's maybe a matter of perspective; the runtime checks are always there and unchanged, whether we do the compiletime checks or not.

bjornregnell:
The main thing here is that we tell the compiler that we want to wait until runtime to check this. So I think `runtimeCheck` is more intuitive than `nonExhaustive` (it *could* be exhaustive anyway even if it is not checked by the compiler) and a beginner should at least start to appreciate that there is a difference between compile time and runtime.

odersky:
But everything we do also has to work for the case of a `val` where there is no exhaustive match.

```scala
  val x: Any = ???
  val y: Int = x.runtimeCheck
```
is more obvious than
```scala
  val x: Any = ???
  val y: Int = x.nonExhaustive
```

lrytz:
There's no `match` keyword, but it's still exhaustiveness checking that's being suppressed in `val` pattern bindings, so it doesn't look wrong to me.

bishabosha:
I have integrated the comments about scala 2.13 behavior of unchecked, in comparison to the proposed semantics of `runtimeCheck`

He-Pin:
So this will not bring in any runtime overhead？does runtime check means unsafecast？

bishabosha:
> So this will not bring in any runtime overhead？does runtime check means unsafecast？

runtime behavior is identical, it only silences the pattern match exhaustiveness check

bishabosha:
If we are to solve the naming issue, then let's spell out some of the ways of interpreting the meaning (feel free to add more):
- cast the value to the type expected by the pattern(s)
- assert I know better than the compiler that this will always match
- I declare match is exhaustive (at runtime) because the invariants of my code mean that other cases will never appear
- delay pattern match exhaustivity checking until runtime (which may result in MatchError)

sjrd:
My favorite is actually the "assert" meaning. Mixed with the 3rd one I would actually think of it as "assert exhaustive". Then I'm wondering if it should throw an `AssertionError` rather than a `MatchError`. 🤔

He-Pin:
refs: Kotlin for cast
https://kotlinlang.org/docs/typecasts.html#unsafe-cast-operator
https://github.com/JakeWharton/confundus

soronpo:
> My favorite is actually the "assert" meaning. Mixed with the 3rd one I would actually think of it as "assert exhaustive". Then I'm wondering if it should throw an `AssertionError` rather than a `MatchError`. 🤔

I was thinking about `assumeChecked`. If the assumption is incorrect then an assertion is triggered.

JD557:
> > My favorite is actually the "assert" meaning. Mixed with the 3rd one I would actually think of it as "assert exhaustive". Then I'm wondering if it should throw an `AssertionError` rather than a `MatchError`. 🤔
> 
> I was thinking about `assumeChecked`. If the assumption is incorrect then an assertion is triggered.

What about `assumeExhaustive`?

`assumeChecked` makes me want to ask "checked for what?"

Kordyjan:
I'm for `assumeMatching,` `assertMatching`, or maybe even `assume` alone. For me all of 
```scala
xs.assume match
  case x0 :: tail => ???

val x0 :: tail = xs.assume

assume(xs) match
  case x0 :: tail => ???

val x0 :: tail = assume(xs)
```

look pretty good.

bjornregnell:
Hmm... `assume` perhaps looks good and reads well and align with `assert` and `require` but to me it is not obvious what happens, compared to `runtimeCheck`, which tells me better what is happening even if I haven't read up on it. I'm not really sure (yet), but I think I'd vote for `runtimeCheck` rather than `assume`.

chrisandrews-ms:
I think `assumeMatching` or `assume match { ...` are a little misleading because we're only assuming anything at compile time; at runtime we are actually checking. To me, `assume` sounds a bit more like the _other_ meaning of `@unchecked` (i.e. erased type argument matching) where we're asking both compiler and runtime to assume that we're correct, and not actually checking. Distinguishing those two meanings of the old `@unchecked` is an important part of this SIP as I understand it.

I prefer @JD557's suggestion `assumeExhaustive` because that describes more precisely what we're doing: it says only that we're assuming that we've covered all of the cases. To me, exhaustivity checking is naturally a compile-time thing, so it wouldn't make me think that there would be no checking at runtime.

odersky:
There is the possibility that we want to expand this further. We now allow:

```scala
val (x: T) = y.runtimeChecked
```
since `(x: T)` is technically a pattern. It seems silly to require the parenthesis and natural to also allow
```scala
val x: T = y.runtimeChecked
```
With `runtimeChecked` this is fine, but anything that mentions "matching" or "exhaustive" will have a problem.

I still think `runtimeChecked` is by far the best of all proposals.

bjornregnell:
I like the trailing `ed` and `runtimeChecked` reads better that `runtimeCheck` as it means "it is runtimeChecked", so better from an English grammar perspective, I think.

chrisandrews-ms:
I'm fine with `runtimeChecked` or `assumeExhaustive`. Both convey the meaning to me that Scala is not checking at compile time but is checking at runtime

chrisandrews-ms:
I'm pleased to announce that the SIP Committee voted in favor of this SIP to be "Accepted for implementation"

bishabosha:
> since (x: T) is technically a pattern. It seems silly to require the parentheses and natural to also allow
> 
> val x: T = y.runtimeChecked
> With runtimeChecked this is fine, but anything that mentions "matching" or "exhaustive" will have a problem.
> 
> To be sure, this is not part of the current SIP, but it would be a likely next step.

Is there a precedent for using annotations to rewrite code?, i.e. would this 
```scala
val x: T = y: @RuntimeChecked // inlining of y.runtimeChecked
```

rewrite to

```scala
val x: T = y match
  case y: T => y
  case _ => throw ClassCastException(...)
```

PR #68, opened 2023-12-19
Fix using directives link in SIP 46

bishabosha:
not sure if we do the change here or directly in the docs.scala-lang repo


bishabosha:
chose #70 over this one

PR #70, opened 2023-12-22
Fix link to "using directives"

Philippus:
I see https://github.com/scala/improvement-proposals/pull/68 also fixes the link, with different url's. Just pick the best fix. :)


SethTisue:
@bishabosha want to make the call here?

bishabosha:
I think this one is more useful :)

PR #72, opened 2024-01-13
SIP-58 - Named Tuples

odersky:
This is the promised SIP for named tuples


content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+By contrast to named tuples, structural types are unordered and have width subtyping. This comes with the price that no natural element ordering exist, and that one usually needs some kind of dictionary structure for access. We believe that the following advantages of named tuples over structural types outweigh the loss of subtyping flexibility:
+
+ - Better integration since named tuples and normal tuples share the same representation.
+ - Better efficiency, since no dictionary is needed.
+ - Natural traversal order allows the formulation of generic algorithms such as projections and joins.
+
+### Conformance
+
+A large part of Pre-SIP discussion centered around subtyping rules,. whether ordinary tuples should subtype named-tuples (as in this proposal) or {\em vice versa} or maybe no subtyping at all.
```
bishabosha, line 257:
```suggestion
A large part of Pre-SIP discussion centered around subtyping rules,. whether ordinary tuples should subtype named-tuples (as in this proposal) or _vice versa_ or maybe no subtyping at all.
```
italics in markdown

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Conformance
+
+A large part of Pre-SIP discussion centered around subtyping rules,. whether ordinary tuples should subtype named-tuples (as in this proposal) or _vice versa_ or maybe no subtyping at all.
+
+Looking at precedent in other languages it feels like we we do want some sort of subtyping for easy convertibility and possibly an implicit conversion in the other direction.
+
+The discussion established that both forms of subtyping are sound. My personal opinion is that the subtyping of this proposal is both more useful and safer than the one in the other direction. There is also the problem that changing the subtyping direction would be incompatible with the current structure of `Tuple` and `NamedTuple` since for instance `zip` is already an inline method on `Tuple` so it could not be overridden in `NamedTuple`. To make this work requires a refactoring of `Tuple` to use more extension methods, and the questions whether this is feasible and whether it can be made binary backwards compatible are unknown. I personally will not work on this, if others are willing to make the effort we can discuss the alternative subtyping as well.
```
lihaoyi, line 261: Can you summarize the discussion here? There were three options (named <: unnamed, unnamed <: named, and no subtyping) and someone reading this SIP should be able to understand the pros and cons of each and why one selection was chosen, even if they may not agree with the choice

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+Backward source compatibility is partially preserved since additions to types and patterns come with new syntax that was not expressible before. When looking at tuple expressions, we have one instance of a source incompatibility:
+
+```scala
+var age: Int
+(age = 1)
+```
+This was an assignment in parentheses before, and is a named tuple of arity one now. It is however not idiomatic Scala code, since assignments are not usually enclosed in parentheses. The problem could also be detected and diagnosed fairly straightforwardly: When faced with a unary named tuple, try to interpret it as an assignment, and if that succeeds, issue a migration error.
+
+### Open questions
```
lihaoyi, line 237: I'd like at least one section one `Use Cases`. A number were discussed in the original thread, and should be summarized here, with strawman implementations of each one demonstrating how modelling the problem using names tuples is superior to the alternatives

In particular interest of me is the discussion around modelling database rows. How would duplicate column names handled if we join two named tuples with overlapping keys?

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+SimpleExpr        ::=  ...
+                    |  '(' NamedExprInParens {‘,’ NamedExprInParens} ')'
+NamedExprInParens ::=  id '=' ExprInParens
+
+Patterns          ::=  Pattern {‘,’ Pattern}
+                    |  NamedPattern {‘,’ NamedPattern}
+NamedPattern      ::=  id '=' Pattern
+```
+
+### Compatibility
```
lihaoyi, line 223: Is there ambiguity in the following case?

```scala
foo bar (qux = ???)
```

Previously it was a call to `bar` with one names param `???`, now it also looks like an infix call passed the named tuple `(qux = ???)` as a positional param. Which will it be interpreted as?

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+Named constructor patterns are analogous to named tuple patterns. In both cases
+
+ - every name must match the name some field of the selector,
+ - names can come in any order,
+ - not all fields of the selector need to be matched.
+
+This revives SIP 43, with a much simpler desugaring than originally proposed.
+Named patterns are compatible with extensible pattern matching simply because
+`unapply` results can be named tuples.
```
lihaoyi, line 184: Can we make user defined class instances support named pattern matching without conversion to named tuples? I feel like "pattern matching without allocation" has been a direction Scala has been going into, with name-based pattern matching and Scala's new pattern matching desugaring, and forcing people to allocate a named tuple to support named pattern matching feels out of place

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+
+### Restrictions
```
lihaoyi, line 187: Can we also have a section enumerating what is _allowed_? Scala 3 tuples have a lot of operations; do named tuples support some of them? all of them? what does that look like? Extending named tuples, Appending named tuples, generic recursion over named tuples?

Even if the answer to this can be derived from
other parts of this doc, better to show concrete examples to help explain to readers what this SIP will allow

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Alternatives
+
+### Structural Types
+
+We also considered to expand structural types. Structural types allow to abstract over existing classes, but require reflection or some other library-provided mechanism for element access. By contrast, named tuples have a separate representation as tuples, which can be manipulated directly. Since elements are ordered, traversals can be defined, and this allows the definition of type generic algorithms over named tuples. Structural types don’t allow such generic algorithms directly. Be could define mappings between structural types and named tuples, which could be used to implement such algorithms. These mappings would certainly become simpler if they map to/from named tuples than if they had to map to/from user-defined "HMap"s.
+
+By contrast to named tuples, structural types are unordered and have width subtyping. This comes with the price that no natural element ordering exist, and that one usually needs some kind of dictionary structure for access. We believe that the following advantages of named tuples over structural types outweigh the loss of subtyping flexibility:
```
lihaoyi, line 249: Do structural types truly always need dictionaries? Can we have e.g. a synthetic trait for every structural member, such that

```scala
def x: {def foo: Int, def bar: String}
```

desugars to

```scala
package scala.structural
trait WithFoo[T]{ def foo: T}
trait WithBar[T]{ def bar: T}
```
```scala
def x: scala.structural.WithFoo[Int] with scala.structural.WithBar[String]
```
 
 The `scala.structual` traits could/would be duplicated by various compilation units, but they would all be the same implementation, and so the duplicate classpath entries would be benign since they would all resolve to the same interfaces

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+Bob match
+  case (age = x, name = y) => ...
+```
+
+### Expansion
+
+Named tuples are in essence just a convenient syntax for regular tuples. In the internal representation, a named tuple type is represented at compile time as a pair of two tuples. One tuple contains the names as literal constant string types, the other contains the element types. The runtime representation of a named tuples consists of just the element values, whereas the names are forgotten. This is achieved  by declaring `NamedTuple`
+in package `scala` as an opaque type as follows:
+```scala
+  opaque type NamedTuple[N <: Tuple, +V <: Tuple] >: V = V
```
lihaoyi, line 129: Is there any way we constrain the `N` type parameter more, given that it can only be literal strings, and not arbitrary types like the `V` paramater next to it?

content/named-tuples.md
```diff
@@ -0,0 +1,282 @@
...
+```
+This was an assignment in parentheses before, and is a named tuple of arity one now. It is however not idiomatic Scala code, since assignments are not usually enclosed in parentheses. The problem could also be detected and diagnosed fairly straightforwardly: When faced with a unary named tuple, try to interpret it as an assignment, and if that succeeds, issue a migration error.
+
+### Open questions
+
+ 1. What is the precise set of types and operations we want to add to `NamedTuple`. This could also evolve after this SIP is completed.
+
+ 2. Should there be an implicit conversion from named tuples to ordinary tuples?
+
+## Alternatives
```
lihaoyi, line 243: An earlier version of this discussion modelled named tuples as tuples whose members were`NamedValue` types, basically a "array of structs" type in contrast to the current proposals "struct of arrays" approach.

What were the pros and cons that resulted in the current approach being chosen over that one?

content/named-tuples.md
odersky: The compiler checks that all labels are string literals. But we can't encode such a constraint in the type bound.

odersky:
There are a lot of good suggestions to add more sections to the proposal, but I am out of time to do this. Maybe someone else wants to help?

content/named-tuples.md
odersky: That has been tried from time to time by us and others. I believe one such paper is: https://dblp.org/rec/conf/oopsla/GilM08.html

But nothing ever materialized. Generally, creating globally visible classes on the fly is a can of worms.

Also, we still need a traversal principle, which named tuples give us, but structural types don't.

content/named-tuples.md
odersky: Unapply could also return some case class, or an option of it. That would work as well for named matching. The only corner case where case classes don't work but tuples do is if there's only one name. See the discussion in SIP 43, which is quite involved.

content/named-tuples.md
odersky: Good point. It would be interpreted as a named tuple. I'll add that point to the compatibility section.

content/named-tuples.md
odersky: The test `named-tuples-strawman-2` in the linked PR [19174](https://github.com/lampepfl/dotty/pull/19174) has a type `UpdateWith` and an associated method `updateWith`, which shows a kind of `join`. The techniques are completely general, so other kinds of joins can be defined in the same way.
```scala
  /** A joint named tuple where
   *   - The names are the names of named tuple `NT1` followed by those names of `NT2` which
   *     do not appear in `NT1`
   *   - The values are the values of `NT1` and `NT2` corresponding to these names.
   *     If a name is present in both `NT1` and `NT2` the value in `NT2` is used.
   */
  type UpdateWith[NT1 <: AnyNamedTuple, NT2 <: AnyNamedTuple] =
    Merge[ConcatDistinct[Names[NT1], Names[NT2]], DropNames[NT1], Names[NT2], DropNames[NT2]]
```
The test also contains the definitions of the helper types used in the definition above.

content/named-tuples.md
lihaoyi: So for normal pattern matching, my understand is that it's dependent on the `ProductN` trait. That means that it is not hardcoded to case classes and tuples, but even normal classes/traits can benefit from zero allocation pattern matching as well. Please correct me if I'm wrong

Would normal classes/traits have some trait they can implement to support names pattern matching in the same way? Or is this a known limitation that we are consciously deciding to live with?

odersky: `ProductN` is no longer a thing in Scala 3. An `unapply` can return a `Product` trait. That's typically the argument itself, which means no allocations. If the argument is a case class (the most common case), we can do named pattern matching on it. If not, we don't have the names, so it's not possible.

lihaoyi: I guess my question is: given position pattern matching has a `Product` trait, that positional `Tuple`s implement, should named pattern matching also have a `NamedProduct` trait that `NamedTuple`s implement? 

At least superficially, I can imagine

1. `NamedProduct[N <: Tuple] extends Product` where `N` is a tuple of string literals
2. `NamedTuple[N <: Tuple, +V <: Tuple] extends `NamedProduct[N]`
3. Define named pattern matching on `NamedProduct`, so that `NamedTuple`s get it for free due to inheriting from `NamedProduct`, but user-land normal classes and traits can inherit from `NamedProduct` and get it as well

I'm not saying we should do this, just that it seems like the natural way of extending the existing `Product`/`Tuple`/positional-pattern-matching relationship to `NamedTuple`/named-pattern-matching. So even if the answer is "no this cannot work" or "maybe it can work but we can always do it later in a follow up" it deserves to be called out.

content/named-tuples.md
soronpo: I think the question is since we are relying on opaque types to represent `NamedTuples`, what stops the user from creating a badly formed NamedTuples manually or via macros? The Scala 3 compiler protects us from creating bad trees, but it won't in this case, unless the checker special-cases `NamedTuples` opaque types in its checks.

content/named-tuples.md
odersky: It's an interesting suggestion. A product match looks for the presence of `_1`, `_2`, ...., selectors. If the scrutinee implements `NamedProduct` , we could then alternatively use the associated names in patterns and selectors instead. But the problem is we can't abstract over it. If we want to have a generic type that reflects names _and_ types, we are back to `NamedTuple`. 

Maybe someone could try that out, see whether it would work? I agree we can do it independently later.

content/named-tuples.md
bishabosha: there was the think that each element of the tuple has to be unwrapped explicitly, so incongruence between `tup(0).value` and `tup.x`

content/named-tuples.md
bishabosha: the consequence would be like passing around `Tuple` instead of a concrete case

lihaoyi: It's probably worth listing out in the proposal all the "magic" constraints and checks that we are adding here, since they are not part of the normal type system and would be special-case support for `scala.NamedTuple` . From what I understand:

- `N <: Tuple` must only contain literal strings
- `N <: Tuple` and `V <: Tuple` must have the same length
- `N <: Tuple` cannot contain duplicates

Is that all, or are there other constraints I'm missing?

content/named-tuples.md
lihaoyi: What's a `traversal principle`?

content/named-tuples.md
odersky: That should be it.

content/named-tuples.md
odersky: A way to inductively visit all the members of a type at the type level, typically using a match type.

content/named-tuples.md
odersky: In fact, those constraints are not enforced by the compiler. You can form arbitrary `NamedTuple` types yourself. There's no additional magic. It's just that they won't fit any named tuple value you write if they are ill-formed.

Sporarum: I think this would be a good occasion to add a `TupleOf[T]` (with the semantics `TupleOf[T]<: EmptyTuple | T *: TupleOf[T]`) to the standard library

While not necessary, it would make the meaning of `N` clearer (even if it is renamed to something else)

content/named-tuples.md
```diff
@@ -0,0 +1,300 @@
...
+
+Values of named tuple types can also be be defined using regular tuples. For instance:
+```scala
+val x: Person = ("Laura", 25)
+
+def register(person: Person) = ...
+register(person = ("Silvain", 16))
+register(("Silvain", 16))
+```
+This follows since a regular tuple `(T_1, ..., T_n)` is treated as a subtype of a named tuple `(N_1 = T_1, ..., N_n = T_n)` with the same element types. On the other hand, named tuples do not conform to unnamed tuples, so the following is an error:
```
jeremyrsmith, line 80: The proposed named tuples are pretty close to shapeless records in principle (just with new syntax at the type and value level to define them). But, the subtyping relationship is the opposite.

E.g. where `(FieldType["foo", Int], FieldType["bar", String])` would be a subtype of `(Int, String)`, in this proposal the equivalent to the former `(foo: Int, bar: String)` would be a supertype of the latter `(Int, String)`.

I think the named value as a refinement of the unnamed value makes a lot more sense. But in the proposal, there are instead special rules to allow the subtyping relationship to be backwards without total insanity.

Please take a look at `FieldType` and the very successful (ignoring the symbol-vs-string issue) decade-long history of named tuples (i.e. records) in shapeless, before making a decision here.

@milessabin I wonder if you have an opinion about this.

utenma:
would name tuples be incubated as experimenta feature for a period of time? guess there could be more feedback after people start to play with them

would really like to see how they play with type level programming

soronpo:
> would name tuples be incubated as experimenta feature for a period of time?

Every new Scala feature is first approved for experimentation, and after a period of time is approved to be shipped. What is approved as experimental is only testable under a nightly or snapshot versions of the compiler and usually involves enabling some flag or importing `language.experimental`.

content/named-tuples.md
```diff
@@ -0,0 +1,300 @@
...
+The translation of named tuples to instances of `NamedTuple` is fixed by the specification and therefore known to the programmer. This means that:
+
+ - All tuple operations also work with named tuples "out of the box".
+ - Macro libraries can rely on this expansion.
+
+### The FieldsOf Type
+
+The `NamedTuple` object contains a type definition
+```scala
+  type FieldsOf[T] <: AnyNamedTuple
```
smarter, line 153: It looks like this was renamed to `From` in https://github.com/lampepfl/dotty/pull/19174. We have a few existing magical type operators such as (https://github.com/lampepfl/dotty/blob/main/library/src/scala/compiletime/ops/int.scala):
```scala
  /** Successor of a natural number where zero is the type 0 and successors are reduced as if the definition was:
   *
   *  ```scala
   *  //{
   *  import compiletime.ops.int.*
   *  //}
   *  type S[N <: Int] <: Int = N match {
   *    case 0 => 1
   *    case 1 => 2
   *    case 2 => 3
   *    // ...
   *    case 2147483646 => 2147483647
   *  }
   *  ```
   *  @syntax markdown
   */
  type S[N <: Int] <: Int
```

Conceptually, can we specify `From` in a similar way as a match type on every case class in the world? Otherwise we'd need to explicitly specify how it handles abstract types (that might be eventually known to be specific case classes) and known invalid inputs.

Additionally, from the usage in https://github.com/dotty-staging/dotty/blob/named-tuples-tcs/tests/run/tyql.scala I gather that it's possible to use `From` on a named tuple in which case it behaves like the identity, is that correct?

content/named-tuples.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Open questions
+
+ 1. What is the precise set of types and operations we want to add to `NamedTuple`. This could also evolve after this SIP is completed.
+
+ 2. Should there be an implicit conversion from named tuples to ordinary tuples?
+
+## Alternatives
+
+### Structural Types
```
smarter, line 263: https://github.com/lampepfl/dotty/pull/19174 now also augments the scala.Selectable mechanism normally used for structural typing to support named tuples as proposed by @soronpo in https://contributors.scala-lang.org/t/expanding-changing-selectable-based-on-upcoming-named-tuples-feature/6395 This is currently documented in https://github.com/dotty-staging/dotty/blob/named-tuples-2/docs/_docs/reference/experimental/named-tuples.md#computed-field-names which I believe could just be copy-pasted in this SIP so we can consider these changes at the same time.

content/named-tuples.md
odersky: I updated the definition of `From`. I am not sure it would make sense to define it as a hypothetical match type.

content/named-tuples.md
odersky: The currently supported operations on named tuples are defined in object `scala.NamedTuple` ultimately what is defined in the `NamedTuple.scala` source, which is a part of the library. I am not sure we should specify the source file in this SIP.

content/named-tuples.md
odersky: I added the section

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+// This is an @main method
+@main def foo(x: Int): Unit =
+  println(x)
+~~~
+
+### Conformance and Convertibility
+
+The order of names in a named tuple matters. For instance, the type `Person` above and the type `(age: Int, name: String)` would be different, incompatible types.
+
+Values of named tuple types can also be be defined using regular tuples. For instance:
```
soronpo, line 72: Remove one "be"

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+
+ 1. What is the precise set of types and operations we want to add to `NamedTuple`. This could also evolve after this SIP is completed.
+
+ 2. Should there be an implicit conversion from named tuples to ordinary tuples?
+
+## Alternatives
+
+### Structural Types
+
+We also considered to expand structural types. Structural types allow to abstract over existing classes, but require reflection or some other library-provided mechanism for element access. By contrast, named tuples have a separate representation as tuples, which can be manipulated directly. Since elements are ordered, traversals can be defined, and this allows the definition of type generic algorithms over named tuples. Structural types don’t allow such generic algorithms directly. Be could define mappings between structural types and named tuples, which could be used to implement such algorithms. These mappings would certainly become simpler if they map to/from named tuples than if they had to map to/from user-defined "HMap"s.
```
soronpo, line 336: "We could" instead of "Be could"

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+(zip: Int, name: String, population: Int)
+```
+The same works for enum cases expanding to case classes, abstract types with case classes as upper bound, alias types expanding to case classes
+and singleton types with case classes as underlying type (in terms of the implementation, the `classSymbol` of a type must be a case class).
+
+`From` is also defined on named tuples. If `NT` is a named tuple type, then `From[NT] = NT`.
+
+### Pattern Matching with Named Fields in General
+
+We allow named patterns not just for named tuples but also for case classes. For instance:
```
lihaoyi, line 230: How does this work? Wouldn't we have to replace 

```scala
def unapply(x: CaseClass): CaseClass
```

with 

```scala
def unapply(x: CaseClass): (foo: Bar, qux: Baz, ...)
```

? 

Or is this more generic, and works with anything returned from `def unapply` as long as the type has a zero-param-list member defined with the correct name?

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+```scala
+city match
+  case c @ City(name = "London") => println(p.population)
+  case City(name = n, zip = 1026, population = pop) => println(pop)
```
lihaoyi, line 234: Should we require a trailing `*` e.g. `case City(name = "London", *) =>` to make it clear that we are only matching a subset of fields?

Is there some way to do named-pattern-matching against _zero_ fields? This is different from `case c: City` because it's `unapply` based rather than `isInstanceOf`-based, and is more convenient than writing `case City(_, _, _) =>` which is the status quo

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+### Pattern Matching
+
+When pattern matching on a named tuple, the pattern may be named or unnamed.
+If the pattern is named it needs to mention only a subset of the tuple names, and these names can come in any order. So the following are all OK:
+```scala
+Bob match
+  case (name, age) => ...
+
+Bob match
+  case (name = x, age = y) => ...
```
lihaoyi, line 131: Is something like this allowed?

```scala
case (name, age = y) =>

```

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+| Date          | Version            |
+|---------------|--------------------|
+| Jan 13th 2024 | Initial Draft      |
+
+## Summary
+
+We propose to add new form of tuples where the elements are named.
+Named tuples can be types, terms, or patterns. Syntax examples:
+```scala
+type Person = (name: String, age: Int)
```
lihaoyi, line 23: What happens if someone writes the following:

```scala
type Person = (name: String, name: String)
```

Is this a syntax error, type error, or something else? The runtime implementation of `(String, String)` doesn't care about names, but then when someone calls `.name` the compiler cannot know which tuple entry it is referring to

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+```scala
+package scala
+import annotation.experimental
+import compiletime.ops.boolean.*
+
+@experimental
+object NamedTuple:
+
+  opaque type AnyNamedTuple = Any
+  opaque type NamedTuple[N <: Tuple, +V <: Tuple] >: V <: AnyNamedTuple = V
```
lihaoyi, line 402: I probably need help from someone else to review this part; this sort of more advanced type-level computation is beyond my expertise

content/named-tuples.md
soronpo: > Is there some way to do named-pattern-matching against _zero_ fields? This is different from `case c: City` because it's `unapply` based rather than `isInstanceOf`-based, and is more convenient than writing `case City(_, _, _) =>` which is the status quo

I also think that `case City(*)` and would opt for that kind of syntax almost always than `_ : City`. This can be a different SIP, unless we require that indeed we need a trailing `*` if not all fields are specified.

odersky: I am not fond of explicit * here. The situation for me is analogous to named function arguments. We don't use * there either to make clear that we have given all names and it's OK to use defaults for the rest. For pattern matching, the default is simply `_`, i.e. wildcard and no binding. 

Also, I think almost all named pattern matches will be partial. That's the whole point. If I have a full match for e.g. `Person`, then I would likely write:
```scala
case Person(name, age)
```
and not
```scala
case Person(name = name, age = age)
```
So in that sense, the fact that it is a named pattern match also indicates that it is partial, and the `*` would just be annoying and redundant.

content/named-tuples.md
odersky: `case (name, age = y)` is currently not allowed.  Named fields no not need to be in the right order. 

Not caring about order comes from the named pattern matching part, and behaves the same for case classes and named tuples. So order matters for construction but not for named deconstruction.

content/named-tuples.md
odersky: > Or is this more generic, and works with anything returned from def unapply as long as the type has a zero-param-list member defined with the correct name?

Correct.

content/named-tuples.md
odersky: It's a type error. It's listed in the Restrictions section of the SIP.

content/named-tuples.md
odersky: As an appendix, it would not need to be reviewed in depth. It just gives additional info that an implementation of `NamedTuple` with a useful API is possible. It's premature to freeze that API, so I believe the NamedTuple API itself should not be part of the SIP, except for things that are explicitly mentioned (like the `NamedTuple` and `From` types).

content/named-tuples.md
lihaoyi: What about members that take implicits? I'm imagining e.g. the Dotty internal types, where just about everything takes a `(given Context)`, even though for most intents and purposes I think of them as fields

What if the return type of `unapply` is a Java type? Do we accept named pattern matching on zero-parameter "getter" methods, or only on Java fields? Or do we prohibit named pattern matching on Java types entirely?

odersky: Implicits are handled by the existing rules. To support a named pattern match, an `unapply` must return one of the following (after handling the get/isDefined cases)

 - A reference to a case class. The fields are the fields in the first parameter list of the case class. (Note that implicits come later)
 - A reference to a named tuple.

lihaoyi: What is the reason we want to limit the return type of `unapply` to case-classes and named-tuples types here? It seems the "look up properties on type returned by `unapply` would work for normal `class`es or `trait` values as well?

As a point of contrast Python's structural pattern matching on class instances works across arbitrary types ([link](https://docs.python.org/3/reference/compound_stmts.html#class-patterns)), as long as they have the relevant properties that are being matched against.

odersky: > Allow unapply to return values of any type, not just case classes or named tuples, and then proceed unchanged with looking up the matched fields on the returned value

That's in a sense already true. `unapply` allows to return instances of any `Product` type. It then proceeds to select the fields of that product type by the `_1`, `_2` selectors. We have a link between these selectors and actual field names only for case classes, since the `_1` selectors are auto-generated for them. For other `Product` instances, these things can be user defined in arbitrary ways so we don't have that info.

Another, deeper reason why pattern matching does not work out of the box for normal classes is that in Scala the constructor elements of a normal class are retained only if prefixed by `val`. So there might be nothing to select. E.g.

```
class Point(x: Int, y: Int)

p match
  case Point(x, y) // does not work, `x` and `y` are not fields of the class.
```
We could consider relaxing that restriction, so that we also allow pattern matching on all classes that have only `val` parameters in their first section. But I believe that's out of the scope for this SIP since it has nothing to do with named matching.

lihaoyi: The reason I think it's relevant to named matching is because the idea was to allow something like this

```scala
class Point(x: Int, y: Int){
  val z = x + y
}

p match{
  case Point(z = 0) => 
}
```

This doesn't work with positional pattern matching, because that relies on the `_1` `_2` `_3` selectors, which do not exist for most non-case classes. Named pattern matching relaxes the reliance: we can just go directly to pattern matching on the member named `.z` directly. Previously such a thing didn't make sense, but with named pattern matching it does

soronpo: The main issue with enabling non-product (fields) pattern matching is the complete loss of exhaustivity indicators.

odersky: @haoyi I see. That use case makes sense. Today we would write
```scala
p match
  case p: Point if p.z == 0 =>
```
The named pattern notation gets more convenient if there are more named elements like this. 

Still, it's a rather big change to the normal pattern matching rules, so I'd prefer we leave that for a possible future time.

content/named-tuples.md
```diff
@@ -0,0 +1,777 @@
...
+`apply`, `head`, `tail`, `take`, `drop`, `++`, `map`, or `zip`.
+Similar to `Tuple`, the `NamedTuple` object also contains types such as `Elem`, `Head`, `Concat`
+that describe the results of these extension methods.
+
+The translation of named tuples to instances of `NamedTuple` is fixed by the specification and therefore known to the programmer. This means that:
+
+ - All tuple operations also work with named tuples "out of the box".
+ - Macro libraries can rely on this expansion.
+
+### Computed Field Names
```
smarter, line 165: I like this feature since it makes Selectable a lot nicer to use. Combined with `NamedTuple.From`, it makes it possible to create powerful proxy types like the `Expr` example in the appendix.

However, there's something a bit odd about using NamedTuple (where the order of the elements matters) to define the members of a Selectable (where there is no ordering). This is fine as long as we can construct a NamedTuple that contains the members we'd like to select on, but if we rely on `NamedTuple.From` we're limited to the parameters of case classes and tuples because they have a notion of ordering but other types don't.

To lift this limitation, I think it'd be useful to define another magic operator to construct a NamedTuple from the public fields of a class or trait, the elements of the NamedTuple should be sorted alphabetically so that reordering `vals` in a class would keep being a source and binary compatible change:
```scala
trait Base:
  val y: Int = 3
class Foo extends Base:
  val z: Int = 1
  val x: Int = 2

NamedTuple.FieldsOf[Foo] =:= (x: Int, y: Int, z: Int)
```

soronpo: >To lift this limitation, I think it'd be useful to define another magic operator to construct a NamedTuple from the public fields of a class or trait

If we add this, I don't see why `NamedTuple.From` cannot be used for both. Or at least a more clear name that will highlight the difference (`FieldsOf` does not).

smarter:
A few points raised at the last SIP meeting:
- There is currently a soundness bug with NamedTuple.From: https://github.com/scala/scala3/issues/20517
  - we believe the fix will require refining the specification of NamedTuple.From to reuse a check for reduction from the match type logic: https://github.com/scala/scala3/blob/0a3497bf7f0ac032f98d36cf1713e8926b5d176c/compiler/src/dotty/tools/dotc/core/TypeComparer.scala#L3397-L3410. This should be reflected in the spec, which should link to the corresponding section of https://docs.scala-lang.org/sips/match-types-spec.html (but it seems to me that right now the match type spec does not precisely defined "concrete" either? Note also that the current match type concreteness check implementation is too lax in some cases: https://github.com/scala/scala3/issues/20518)
- @sjrd suggested that NamedTuple should be implemented as a class (with a custom erasure) rather than an abstract type to avoid unintuitive restrictions related to match types (e.g., https://github.com/scala/scala3/issues/20453)
  - I've also recently realized that pattern matching on a quoted type is broken for parameterized abstract types (you get a MatchError: https://github.com/scala/scala3/issues/16782), this means that meta-programming with NamedTuple is significantly harder than with regular Tuple currently.

odersky:
[Disallow some opaque type aliases in match type patterns #20913
](https://github.com/scala/scala3/pull/20913) should address the problem with failing match type reductions. It serves as a safety net ensuring that match type reduction on opaque aliases will not fail later.

julienrf:
> I changed the stage label to stage:implementation - as that is required by our state transition diagram (and the script that generates the sips website)

Does that mean that the design has been accepted by the SIP Committee? If that’s the case, it would be good to merge the PR so that the proposal gets its own page like e.g. https://docs.scala-lang.org/sips/polymorphic-eta-expansion.html.

soronpo:
> > I changed the stage label to stage:implementation - as that is required by our state transition diagram (and the script that generates the sips website)
> 
> Does that mean that the design has been accepted by the SIP Committee? If that’s the case, it would be good to merge the PR so that the proposal gets its own page like e.g. https://docs.scala-lang.org/sips/polymorphic-eta-expansion.html.

Only as experimental. Not final yet.

soronpo:
@lihaoyi can we vote to accept this tomorrow?

lihaoyi:
Sure let's vote

bjornregnell:
It would be good if the recommendations of each assignee could be stated here prior to the vote.

kyouko-taiga:
I agree. Could the assignees update the agenda?

soronpo:
I recommend *YES* to accept.

PR #74, opened 2024-01-18
SIP-60 - Alternative bind patterns

yilinwei:
This is the associated contributors [thread](https://contributors.scala-lang.org/t/pre-sip-bind-variables-for-alternative-patterns/6321/13).

The linked implementation is incomplete - I am happy to provide a more complete implementation at a later stage of the process.

### DIFF
--- a/content/alternative-bind-variables.md
+++ b/content/alternative-bind-variables.md
@@ -2,7 +2,8 @@
 layout: sip
 permalink: /sips/:title.html
 stage: pre-sip
-status: waiting-for-implementation
+status: submitted
+presip-thread: https://contributors.scala-lang.org/t/pre-sip-bind-variables-for-alternative-patterns/6321/13
 title: SIP-NN - Bind variables within alternative patterns
 ---
 
--- a/content/alternative-bind-variables.md
+++ b/content/alternative-bind-variables.md
@@ -1,0 +1,248 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: pre-sip
+status: waiting-for-implementation
+title: SIP-NN - Bind variables within alternative patterns
+---
+
+**By: Yilin Wei**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| Sep 17th 2023 | Initial Draft      |
+
+## Summary
+
+Pattern matching is one of the most commonly used features in Scala by beginners and experts alike. Most of
+the features of pattern matching compose beautifully — for example, a user who learns about bind variables
+and guard patterns can mix the two features intuitively.
+
+One of the few outstanding cases where this is untrue, is when mixing bind variables and alternative patterns. The part of
+current [specification](https://scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html) which we are concerned with is under section **8.1.12** and is copied below, with the relevant clause
+highlighted.
+
+> … All alternative patterns are type checked with the expected type of the pattern. **They may not bind variables other than wildcards**. The alternative …
+
+We propose that this restriction be lifted and this corner case be eliminated.
+
+Removing the corner case would make the language easier to teach, reduce friction and allow users to express intent in a more natural manner.
+
+## Motivation
+
+## Scenario
+
+The following scenario is shamelessly stolen from [PEP 636](https://peps.python.org/pep-0636), which introduces pattern matching to the
+Python language.
+
+Suppose a user is writing classic text adventure game such as [Zork](https://en.wikipedia.org/wiki/Zork). For readers unfamiliar with
+text adventure games, the player typically enters freeform text into the terminal in the form of commands to interact with the game
+world. Examples of commands might be `"pick up rabbit"` or `"open door"`.
+
+Typically, the commands are tokenized and parsed. After a parsing stage we may end up with a encoding which is similar to the following:
+
+```scala
+enum Word
+  case Get, North, Go, Pick, Up
+  case Item(name: String)
+    
+  case class Command(words: List[Word])
+```
+
+In this encoding, the string `pick up jar`, would be parsed as `Command(List(Pick, Up, Item("jar")))`.
+
+Once the command is parsed, we want to actually *do* something with the command. With this particular encoding,
+we would naturally reach for a pattern match — in the simplest case, we could get away with a single recursive function for
+our whole program.
+
+Suppose we take the simplest example where we want to match on a command like `"north"`. The pattern match consists of
+matching on a single stable identifier, `North` and the code would look like this:
+
+~~~ scala
+import Command.*
+    
+def loop(cmd: Command): Unit =
+  cmd match
+    case Command(North :: Nil) => // Code for going north
+~~~
+
+However as we begin play-testing the actual text adventure, we observe that users type `"go north"`. We decide
+our program should treat the two distinct commands as synonyms. At this point we would reach for an alternative pattern `|` and
+refactor the code like so:
+
+~~~ scala
+  case Command(North :: Nil | Go :: North :: Nil) => // Code for going north
+~~~
+
+This clearly expresses our intent that the two commands map to the same underlying logic.
+
+Later we decide that we want more complex logic in our game; perhaps allowing the user to pick up
+items with a command like `pick up jar`. We would then extend our function with another case, binding the variable `name`:
+
+~~~ scala
+  case Command(Pick :: Up :: Item(name) :: Nil) => // Code for picking up items
+~~~
+
+Again, we might realise through our play-testing that users type `get` as a synonym for `pick up`. After playing around
+with alternative patterns, we may reasonably write something like:
+
+~~~ scala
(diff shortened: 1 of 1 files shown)


content/alternative-bind-variables.md
```diff
@@ -0,0 +1,321 @@
...
+
+This begs the question of what to do in the case of an explicit `@` binding where the user binds a variable to the same _name_ but to different types. We can either expose a `String | Int` within the contextual scope, or simply reject the code as invalid.
+
+~~~ scala
+  enum Foo:
+    case Bar(s: String)
+    case Baz(i: Int)
+    
+    def fun = this match
+      case Bar(x @ given String) | Baz(x @ given Int) => ???
```
smarter, line 238: I would stay simple here and require the types to match, the user can always write `given (Int | String)` or use a type alias to avoid repetition.

content/alternative-bind-variables.md
```diff
@@ -0,0 +1,321 @@
...
+Let $`p_1 | \ldots | p_n`$ be an alternative pattern at an arbitrary depth within a case pattern and $`\Gamma_n`$ is the named scope associated with each alternative.
+
+Let the named variables introduced within each alternative $`p_n`$, be $`x_i \in \Gamma_n`$ and the unnamed contextual variables within each alternative have the type $`T_i \in \Gamma_n`$.
+
+Each $`p_n`$ must introduce the same set of bindings, i.e. for each $`n`$, $`\Gamma_n`$ must have the same **named** members $`\Gamma_{n+1}`$ and the set of $`{T_0, ... T_n}`$ must be the same.
+
+If $`X_{n,i}`$, is the type of the binding $`x_i`$ within an alternative $`p_n`$, then the consequent type, $`X_i`$, of the 
+variable $`x_i`$ within the pattern scope, $`\Gamma`$ is the least upper-bound of all the types $`X_{n, i}`$ associated with
+the variable, $`x_i`$ within each branch.
+
```
smarter, line 297: Looks like the spec doesn't mention the effect of the presence of Ti on the pattern scope.

smarter (approved):
Very well-written SIP, thank you! One thing that would be worth investigating is: how does this interact with [https://docs.scala-lang.org/scala3/guides/macros/quotes.html#quoted-patterns](quoted patterns) ? /cc @nicolasstucki  I think it's fine to just not support them at least initially, but that should be explicit in the SIP.

nicolasstucki:
#### About quoted patterns 
> One thing that would be worth investigating is: how does this interact with [quoted patterns](https://docs.scala-lang.org/scala3/guides/macros/quotes.html#quoted-patterns)? /cc @nicolasstucki I think it's fine to just not support them at least initially, but that should be explicit in the SIP.

**For the sake of this SIP, I would say that we do not support quoted patterns.**

That said, the current specification should also work for quoted patterns (maybe partially). I would assume that this would be one of the use-cases
```
  case '{ f($x: Int) } | '{ g($x: Double) } => x: Expr[Int] | Expr[Double]
```

The content of expression splices (the `p` in `case '{ f(${p}) }`) are patterns that could follow the same rules laid out in this SIP. While it should not be hard on the language level, there might be some extra complexity in the implementation due to the non-standard way quote patterns are typed. 

Type variables in quoted patterns are another story, I am not sure if we can apply the same specification to these constructs, we would have to check if it is sound or can lead to confusion on the user level.

content/alternative-bind-variables.md
yilinwei: That was originally my intention, but I didn't like the fact that it "clashes" with the rules set up for named bindings and thought it might be better to be consistent with the named binding rule and hence wouldn't have to carve out an exception in the specification.

I don't feel strongly either way; in my opinion if I was binding a `given` as a union type, I would use an alias or write it out explicitly anyway.

content/alternative-bind-variables.md
yilinwei: @smarter 

Since the specification isn't complete for Scala 3 (in order for the amendments to be complete, I'd first need to change the specification to reflect the current behaviour in dotty), should I include the changes to the spec to reflect the current behaviour within this SIP as well? Or is a separate PR to the spec enough?

smarter: A separate PR to the spec would be great and highly appreciated!

content/alternative-bind-variables.md
```diff
@@ -0,0 +1,331 @@
...
+## Implementation
+
+The author has a current in-progress implementation focused on the typer which compiles the examples with the expected types. Interested
+ parties are welcome to see the WIP [here](https://github.com/lampepfl/dotty/compare/main...yilinwei:dotty:main).
+
+### Further work
+
+#### Quoted patterns
+
+More investigation is needed to see how quoted patterns with bind variables in alternative patterns.
```
smarter, line 326: Missing words?
```suggestion
More investigation is needed to see how quoted patterns with bind variables in alternative patterns could be supported.
```

content/alternative-bind-variables.md
smarter: After discussing it in the SIP meeting, we feel that it might be best to not include support for given bind variables in alternatives at least initially, since it opens a can of worms and it's not clear if there is a lot of usecases for it. (But for the record, it seems that if we wanted to support them we would have to require the types to be identical to handle `case Foo(given A, given B) | Bla(given B, given A)` which works out because the givens get desugared into `given_A @ _`, `given_B @ _`).

content/alternative-bind-variables.md
```diff
@@ -0,0 +1,331 @@
...
+
+~~~scala
+def test = this match
+  case Bar(id: Int) => id
+  case Baz(id) => id
+~~~
+
+On the other hand, if it is decided that each bound variable ought to be the same type, then arguably "sharing" explicit type ascriptions across branches would reduce boilerplate.
+
+#### Missing variables
```
soronpo, line 278: Regarding missing variables, we need a decision. The spec should either specify this is an error or assign the type `A | Null`. Because currently explicit nulls are under a flag, I'm inclined to think that this should be an error at least until explicit nulls are always enabled.

odersky: Even with explicit nulls I would not mix `null` with other features like it would be done here. I think it is entirely reasonable to make this an error.

content/alternative-bind-variables.md
yilinwei: Relevant PR is [here](https://github.com/lampepfl/dotty/pull/19723).

For the record, should the synthesized names part of the spec or are they an implementation detail i.e. should the following be valid code?

```scala

type MyString = String

("foo", "foo") match:
  case (given MyString, given String) => ???
```

smarter: Good question, I'd say they should be part of the spec since they have to be for non-local definitions anyway (for ABI stability) and since it doesn't seem like we need to treat them differently in local patterns.

yilinwei: I would be in favour of that interpretation.

It leads to a more straightforward interpretation of both:

`x @ given String |  y @ given String` 

and

`given MyString  |  given String`

Both are invalid now due to fact that the names introduced in each branch are different (if we don't take the alternative for missing variables), regardless of whatever instances are available in the implicit scope.

content/alternative-bind-variables.md
```diff
@@ -0,0 +1,331 @@
...
+
+Unlike in other languages, we could assign a type, `A | Null`, to a bind variable which is not present in all of the alternative branches. Rust, for example, is constrained by the fact that the size of a variable must be known and untagged unions do not exist.
+
+Arguably, missing a variable entirely is more likely to be an error — the absence of a requirement for `var` declarations before assigning variables in Python means that beginners can easily assign variables to the wrong variable.
+
+It may be, that the enforcement of having to have the same bind variables within each branch ought to be left to a linter rather thana a hard restriction within the language itself.
```
bjornregnell, line 284: Typo "thana"

content/alternative-bind-variables.md
```diff
@@ -0,0 +1,331 @@
...
+enum Foo:
+  case Bar(x: Int)
+  case Baz(y: Int)
+    
+  def fun = this match
+    case Bar(z) | Baz(z) => ... // z: Int
+~~~
+
+For the expression to make sense with the current semantics around pattern matches, `z` must be defined in both branches; otherwise the
+case body would be nonsensical if `z` was referenced within it (see [missing variables](#missing-variables) for a proposed alternative).
```
bjornregnell, line 156: The link to the "Missing variable" section does not seem to work.

smarter:
The SIP committee voted to **accept** this proposal as experimental, congratulations! There was consensus among the committee for the following changes:
- Drop support for given bind variables (if there is demand for them, they could be added back later)
- Rewrite the "Missing variables" section to always make it an error to miss a variable

@yilinwei Would you be willing to make these changes?

yilinwei:
@smarter 

Certainly; I am happy to make the changes. I am a little busy at the moment, so will get to within the next month or so.

smarter:
Sounds good, thanks!

smarter:
Hi @yilinwei, do you think you'll have time to make those changes? Otherwise no worries, I should be able to find some time soon.

PR #78, opened 2024-02-15
SIP-61 - Unroll Default Arguments for Binary Compatibility

lihaoyi:


content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+   def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0) = s + n + b + l
+
+   def foo(s: String, n: Int, b: Boolean) = foo(s, n, b, 0)
+   def foo(s: String, n: Int) = foo(s, n, true, 0)
+}
+```
+
+As a result, old callers who expect `def foo(String, Int, Boolean)` or `def foo(String, Int, Boolean, Long)`
+can continue to work, even as new parameters are added to `def foo`. The only restriction is that 
+new parameters can only be added on the right, and they must be provided with a default value.
```
odersky, line 263: It would be good to add an explanation of what happens when there are multiple extensions of a method in different versions. Say in version 2 write 
```scala
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0)
```
and code like `foo("", 1, false, -1L)` compiles against that. Then in version 3  I would like another parameter with default to `foo` like this:
```scala
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0, f: Float = 0.0f)
```
How do I make the previous call `foo("", 1, false, -1L)` work? I assume I'll need another `@unroll`. What would the generated code be in this case?

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+provide benefits. Thus, unless we can find some solution, we cannot properly support
+virtual methods and overrides in `@unroll`.
+
+It may be possible to loosen this restriction to also allow abstract methods that
+are implemented only once by a final method. See the section about 
+[Abstract Methods](#abstract-methods) for details.
+
+## Major Alternatives
+
+The major alternatives to `@unroll` are listed below:
```
odersky, line 550: I would add an implementation with macro annotations to that list. Macro annotations are about to land as a SIP. We should evaluate whether and how they could implement this behavior. If they can do it, we should go down that road, I think. Then the discussion would be where the new macro should live. I would argue that it should go into the standard library in that case.

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+
+| Date          | Version            |
+|---------------|--------------------|
+| Feb 14th 2024 | Initial Draft      |
+
+## Summary
+
+This SIP proposes an `@unroll` annotation lets you add additional parameters
+to method `def`s,`class` construtors, or `case class`es, without breaking binary 
+compatibility. `@unroll` works by generating "unrolled" or "telescoping" forwarders:
```
odersky, line 21: Backwards compatible default parameters are definitely a pain point and it would be good if we could address this.

content/unroll-default-arguments.md
lihaoyi: As currently written, `@unroll` generates all forwarders for parameters to the right of the annotation, and `@unrollOnly` generates a single forwarder for only that parameter.

We could also flip it around, and call them `@unrollAll` and `@unroll`, but the semantics would be the same. Just a question of naming

content/unroll-default-arguments.md
lihaoyi: Yes that's definitely worth a consideration. I will need to wait for the macro annotations proposal to appear before I can really say, but if we can implement this in a macro annotation rather than hardcoded in the compiler that would definitely be preferable.

content/unroll-default-arguments.md
odersky: So, assuming `unroll`, would I write
```scala
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0, @unroll f: Float = 0.0f)
```
? And what would that generate?

content/unroll-default-arguments.md
odersky: We can propose it as a test case for the implementers of macro annotations /cc @nicolasstucki @hamzaremmal.

content/unroll-default-arguments.md
bjornregnell: I think a combination of something similar to `@unrollAll` + `@unroll` is perhaps most intuitive and with a lesser risk for misinterpretation. Or perhaps the latter should be called `@unrollSingle` to be explicit - I mean it is not easy to know what just `@unroll` means...

odersky: Can we clarify what would happen under `@unroll` before discussing alternatives?

lihaoyi: As the proposal is currently written, it's

```scala
// Invalid, multiple `@unroll`s not allowed
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0, @unroll f: Float = 0.0f)

// Single `@unroll` generates forwarders for every method to the right
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0, f: Float = 0.0f)
// Generated
def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0)
def foo(s: String, n: Int = 1, @unroll b: Boolean = true)
def foo(s: String, n: Int = 1)

// `@unrollOnly`s generates forwarders only for annotated parameters
def foo(s: String, n: Int = 1, @unrollOnly b: Boolean = true, l: Long = 0, @unrollOnly f: Float = 0.0f)
// Generated
def foo(s: String, n: Int = 1, @unrollOnly b: Boolean = true, l: Long = 0)
def foo(s: String, n: Int = 1)
```

We could swap out `@unroll`/`@unrollOnly` with `@unrollAll`/`@unroll` or `@unrollAll`/`@unrollOnly` or `@unroll.all`/`@unroll.only` if people think that's more intuitive. The naming is pretty arbitrary and I don't want to spend too much time discussing it up front

bjornregnell: Sorry for entering "naming mode" :) 
(I just wanted to give my initial feedback/thoughts...)

content/unroll-default-arguments.md
lihaoyi: I do think we do not need to block on macro annotations before reviewing this proposal. 

Whether `@unroll` ends up being baked into the compiler, implemented as a compiler plugin (as it is today), or implemented using macro annotations, the feature is still the same. It's just a matter of what line you need in your build file to enable it.

e.g., we could decide that we want `@unroll` as a feature, leave it a compiler plugin while macro annotations are figured out (possibly under the `github.com/scala` org?), and eventually fold it into the standard library as a macro annotation when those are ready. The Scala 2 implementation will need to be done differently regardless, either baked-in or remaining as a compiler plugin.

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+
+7. `@unroll` only supports `final` methods. `object` methods and constructors are naturally
+   final, but `class` or `trait` methods that are `@unroll`ed need to be explicitly marked `final`.
+   It has proved difficult to implement the semantics of `@unroll` in the presence of downstream 
+   overrides, `super`, etc. where the downstream overrides can be compiled against by different 
+   versions of the upstream code. If we can come up with some implementation that works, we can
+   lift this restriction later, but for now I have not managed to do so and so this restriction
+   stays.
+
+### Challenges of Non-Final Methods and Overriding
```
lihaoyi, line 485: I'd like to highlight this section as worth discussing.

I would like to be able to support non-final methods, abstract-methods-implemented-by-final-methods, and abstract-methods-implemented-by-non-final-methods. But I haven't managed to come up with an implementation strategy that I can convince myself is correct.

Right now, I'm reasonably confident that:

- `@unroll`ing final methods works
- `@unroll`ing override-able concrete methods cannot be done safely

There's one case I'm not sure about

- `@unroll`ing abstract methods that are only implemented by final methods (i.e. implemented with no subsequent overrides)

If we could make this case work, it would open up the use of `@unroll` in traits and abstract classes, which would be a big win over the status quo of only allowing final concrete methods

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+   def this(s: String, n: Int, b: Boolean) = this(s, n, b, 0L)
+   
+   def copy(s: String, n: Int) = copy(s, n, true, 0L)
+   def copy(s: String, n: Int, b: Boolean) = copy(s, n, b, 0L)
+   
+   def foo = s + n + b
+}
+object Unrolled{
+   def apply(s: String, n: Int) = apply(s, n, true, 0L)
+   def apply(s: String, n: Int, b: Boolean) = apply(s, n, b, , 0L)
```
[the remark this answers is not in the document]
lihaoyi: Thanks, fixed

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+
+```scala
+import scala.annotation.unroll
+
+case class Unrolled(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0L){
+   def this(s: String, n: Int) = this(s, n, true, 0L)
+   def this(s: String, n: Int, b: Boolean) = this(s, n, b, 0L)
+   
+   def copy(s: String, n: Int) = copy(s, n, true, 0L)
+   def copy(s: String, n: Int, b: Boolean) = copy(s, n, b, 0L)
```
JD557, line 377: I wonder if the `copy` can be unrolled like this..., but maybe I'm just confused on how the normal desugaring and default arguments work.

My understanding is that, on a first version of `case class Unrolled(s: String, n: Int = 1)`, the copy method would be declared as `def copy(s: String = this.s, n: Int = this.s)`

Now, say that the user code calls `unrolled.copy(n = 2)`, it will link to that 2 argument version.

Now, on the new version with `case class Unrolled(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0L)`, we will get 3 versions of copy:

- `def copy(s: String = this.s, n: Int = this.s, b: Boolean = this.b, l: Long = this.l)`
- `def copy(s: String, n: Int) = copy(s, n, true, 0L)`
- `def copy(s: String, n: Int, b: Boolean) = copy(s, n, b, 0L)`

Which one will `unrolled.copy(n = 2)` call? Is it able to call the 4 argument version (I imagine not, as that is the point of this SIP, right?)

---

On another note, I'm not sure if `def copy(s: String, n: Int) = copy(s, n, true, 0L)` is correct... Shouldn't it propagate the current values, like `def copy(s: String, n: Int) = copy(s, n, this.b, this.l)`?

Even if the end user is not able to modify those (as they are expecting the old version), the library that is pulling latest version might, and should expect the context to survive under some operations. 🤔

lihaoyi: > Which one will unrolled.copy(n = 2) call? Is it able to call the 4 argument version (I imagine not, as that is the point of this SIP, right?)

I actually don't know haha. It depends on overload resolution, and I've found it sometimes gives me one and sometimes the other, and I can't figure out why.

Regardless, however, the forwarders and primary implementation *should* have the same semantics, so it's probably fine either way

> On another note, I'm not sure if def copy(s: String, n: Int) = copy(s, n, true, 0L) is correct... Shouldn't it propagate the current values, like def copy(s: String, n: Int) = copy(s, n, this.b, this.l)?

Yes, I think you are correct. It should refer to the current value, not the default value

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+given default parameter values:
+
+```scala
+import scala.annotation.unroll
+
+object Unrolled{
+   def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0) = s + n + b + l
+
+   def foo(s: String, n: Int, b: Boolean) = foo(s, n, b, 0)
+   def foo(s: String, n: Int) = foo(s, n, true, 0)
```
lihaoyi, line 257: This is actually a bit simplified. We do not inline the constants `true` and `0`, but instead get them from the `this.foo$default$1` and `this.foo$default$2` methods. This means that if someone overrides the default with a different, the overriden value is used. But for the vast majority of scenarios that does not make a difference to the user

odersky: Understood. But why do we need two generated `foo` functions? I would have thought if we add a new `b` parameter, we put the unroll on the `b` and get:

```scala
def foo(s: String, n: Int) = foo(s, n, true, 0)
```
Isn't that enough? Why do we need the three parameter `foo`?

lihaoyi: The assumption was tha new default parameters were added one at a time, so two/three/four parameter versions of `foo` exist and were published, and so we need the corresponding forwarders for code compiled against each version

If that assumption is not true, then `@unrollOnly` gives more fine grained support on what forwarders need to be generated

odersky: I still don't get it. Can you walk me through step by step. The way I see this is:

Version 1 has
```scala
def foo(s: String, n: Int = 0)
```
I threw in a default, just to cover that case.

Version 2 has:
```scala
def foo(s: String, n: Int = 0, @unroll b: Boolean = true, l: Long = 0)
def foo(s: String, n: Int) = foo(s, n, true, 0) // auto-generated
```
The default parameter on the generated version is gone. But that's OK because we have the 
same default accessors, so old code calling `foo("hello")` expanded to `foo("hello", foo$default$2)` and that also works fine against the new generated version. 

Version 3 has:
```scala
def foo(s: String, n: Int = 0, @unroll b: Boolean = true, l: Long = 0, @unroll f: Float = 0.0f), c: Char = 'a')
def foo(s: String, n: Int., b: Boolean, l: Long) = foo(s, n, b, l, 0.0f, 'a')  // auto-generated
def foo(s: String, n: Int) = foo(s, n, true, 0) // auto-generated
```

I must be missing something, since there's always exactly one function generated, but I don't see what it is.

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,817 @@
...
+   selected: that is up to the Scala compiler and how overloading interacts with default
+   parameter values. I have encountered scenarios with manually-written forwarders where
+   selected over the "primary" implementation
+
+3. The forwarders are meant to have the exact same semantics as the "primary" implementation.
+   Thus it _does not matter_ whether you call the primary and pass it a default value,
+   or you call a forwarder which then calls the primary and passes it the same default value.
+
+For now, I have left the generated methods "as is", though choosing to hide them or deprecate
+them or something else is definitely an option
```
lrytz, line 734: Visible synthetic overloads would affect the user experience in IDEs, scaladocs, etc.

content/unroll-default-arguments.md
lihaoyi: Got it. 

alAn earlier version of the prototype had an `@unroll("x")` annotation on the method, where the `"x"` string parameter specified the parameter name from which to begin unrolling.

We could also do two annotations: `@unroll` on the method, and then `@unroll2` or something on the starting parameter.

I don't have a strong opinion between these options. The current API was chosen as the most minimal while being able to provide the flexibility necessary, but the others are fine too IMO (if marginally more verbose)

content/unroll-default-arguments.md
lrytz: 👍

Are macro annotation generated methods also hidden in separate compilation?

content/unroll-default-arguments.md
lihaoyi: What you described is exactly the (as-currently-written) semantics for `@unrollOnly`. You are not missing anything,  it's correct. With 3 versions, adding two default arguments in versions 2 and 3, only 2 generated forwarders are necessary to maintain binary compatibility.

I think where your understanding differs from the current proposal, is that the current proposal calls what you described `@unrollOnly` rather than `@unroll`. In the current proposal, where `@unroll` is shorthand for _"add `@unrollOnly` to the parameter annotated and every parameter to the right"_.

The naming should definitely be changed if it's unintuitive. If you and @bjornregnell both find the current naming confusing, that's probably the case haha

content/unroll-default-arguments.md
odersky: They are.

content/unroll-default-arguments.md
odersky: Ah, but why would you use `unroll` then over `unrollOnly`?  If you add multiple parameters in one version, there's no need to add a new forwarder for each prefix. I think I would prefer just one annotation. Call it `unroll` with the semantics of `unrollOnly`. Then if you want multiple forwarders, just put multiple @unrolls on your parameters. That's not a big burden, and would be less confusing.

lihaoyi: I have updated the document to use `@unroll`/`@unrollAll`

I have not updated the prototype implementation, but it should be easy to do so when the time comes

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,797 @@
...
+It may be possible to loosen this restriction to also allow abstract methods that
+are implemented only once by a final method. See the section about 
+[Abstract Methods](#abstract-methods) for details.
+
+## Major Alternatives
+
+The major alternatives to `@unroll` are listed below:
+
+1. [data-class](https://index.scala-lang.org/alexarchambault/data-class)
+2. [SBT Datatype](https://www.scala-sbt.org/1.x/docs/Datatype.html)
```
eed3si9n, line 552: Could you link to the current incarnation of sbt-datatype, which is [Contraband](https://www.scala-sbt.org/contraband/) instead please? Contraband primarily uses [GraphQL Schema](https://www.scala-sbt.org/contraband/schema.html) as the schema language with a few extensions like using `@since("1.0.0")` to denote when a field was added, though we do support the JSON-style schema for internal usage.

In addition to being binary-JAR compatible, Contraband supports generation of serialization codec, which is an important aspect similar to other schemas like Protobuf and Avro. Ideally, we should support cross talk between two services with all combinations of v1 -> v2, v2 -> v1, etc.

alexarchambault (commented):
That looks great, has a larger scope than data-class, and incidentally, also has a Scala 3 implementation (unlike data-class). I hope to be able to find time to give it a try on the coursier code base, to add Scala 3 support to it.

content/unroll-default-arguments.md
odersky: Or, if we go with an annotation on the method, it could be 
```scala
@unrollFrom("n") 
def foo(s: String, n: Int = 0, b: Boolean = true)
```
vs
```scala
@unrollAt("n") 
def foo(s: String, n: Int = 0, b: Boolean = true)
```

(Now it's me who caught the naming bug)

lihaoyi: Yes, that works, but I would argue that having a secondary annotation on the parameter itself is much better than duplicating the parameter name above the method def. 

A secondary annotation on the parameter is much more clear which parameter it applies to, avoids the need for duplicating the parameter name, and is much easier to line up when the number of parameters annotated grows

content/unroll-default-arguments.md
lihaoyi: Thanks! I've updated the link to contraband.

lihaoyi:
@alexarchambault please give it a try! Do note that the current published compiler plugin uses an older version of the proposal. You can take a look at the linked example PRs to see how to use it

lihaoyi:
I have updated the writeup and prototype implementation with support for abstract method `def`s, with the same limitation that any downstream implementations must be final (i.e. no `override` alliowed). This opens up `@unroll` for usage in interfaces, which will definitely prove handy: interface methods have to evolve just as much as `final` methods do!

I am reasonably confident the treatment of abstract method `def`s is correct, but please help me take a look at the section `Abstract Methods` and help me double check things.

lihaoyi:
Also CC @lefou and @lolgab as the folks who have been pushing for binary-compatibility and semantic versioning within the `com.lihaoyi` ecosystem; would be great if you guys could help give this proposal a read through and see if it makes sense

lihaoyi:
> Adding an abstract method to an open hierarchy is never a backward binary compatible change: it breaks existing subclasses. It's OK if the whole hierarchy is sealed.

in this case, `@unroll` generates reverse forwarders such that no abstract methods are added or removed. MiMa passes, and my own ad-hoc classpath mynging tests pass. 

I don't know whether you've read through the updated proposal, but there's a whole section on the unrolling of abstract methods and why it allows addition of default parameters to be backwards binary compatible and semantically compatible

lefou:
I think having at least the option to mark the generated forwarder methods as deprecated is important. Although binary compatibility is the driving concern here, APIs are use by developers who need to understand and decide which method they should go for. Without any `@deprecation` notice, they may choose the wrong, resulting in unnecessarily limited compatibility. And I'm also having polyglot APIs, reflection acces and ScalaDoc in mind, where it matters, which explicit API is recent and what is deprecated; in contrast to pure Scala code, where the compile is most likely deciding which override to call.

And since `@deprecated` takes a `since` parameter which is IMHO important, we most likely need a `since` in `@unroll` too.

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,921 @@
...
+```scala
+object UnrolledObj extends Unrolled{ // version 3
+  def foo(s: String, n: Int = 1, @unroll b: Boolean = true, @unroll l: Long = 0) = s + n + b
+}
+```
+
+This unrolls to:
+```scala
+trait Unrolled{ // version 3
+  def foo(s: String, n: Int = 1, @unroll b: Boolean = true, @unroll l: Long = 0): String = foo(s, n, b)
```
sjrd, line 486: This unroll will not cause a `LinkageError`, but it is semantically incorrect anyway. *Ignoring* the value of the parameter at definition site is most likely not allowed by the new spec of the method. It is certainly not at all what a default value stands for at the source level.

lihaoyi:
@lrytz suggested that in Scala 2 we could generate the forwarders post pickler phase, which makes them invisible to downstream Scala compilation runs. And Scala 3 there is an Invisible annotation that does the same thing

I havent implemented these in the prototype yet, bit they sound like they would work

content/unroll-default-arguments.md
lihaoyi: > but it is semantically incorrect anyway. Ignoring the value of the parameter at definition site is most likely not allowed by the new spec of the method.

I think the first line here doesn't really follow from the second. `@unroll` doesn't make any guarantees over the spec of the method; that's up to you to specify and implement. 

* Some methods' specs will require that you specifically handle the new default parameter. These are fundamental incompatibilities, and `@unroll` can do nothing to help

* Some methods' specs will be fine with you ignoring additional default parameters. For this category, there is no fundamental incompatibility, and `@unroll` makes it so there is no technical/implementation incompatibility either

Now, exactly what the split is between these two categories is an empirical question, but it seems plausible that the latter category is non-zero, and for those this style of reverse forwarder would be useful. But this is up for discussion. If people think this treatment of unrolled abstract methods is unintuitive/confusing/not-worth-the-complexity, I'd be happy to drop it

sjrd: I suggest dropping it. I don't doubt that there are situations where the specs would be fine, but I also expect them to be pretty rare. Regardless, it is infinitely more dangerous than `@unroll` everywhere else, since everywhere else any spec will be accommodated by `@unroll`. I suggest we stick to situations where we can do no harm.

lihaoyi: One way of thinking about `@unroll`ing abstract methods is that it's meant for adding _optional_ parameters to an invocation _protocol_, rather than any specific caller or implementor.

These are parameters that is optional both for the caller to _pass_, as well as the implementor to _handle_. Beyond this specific case of scala `trait`s with abstract `def`s, this is a relatively common concept in software engineering. 

Examples include:

- HTTP headers, which servers and clients exchange, are usually optional: e.g. `Content-Type` `SHOULD` be provided, not `MUST` [link](https://stackoverflow.com/questions/15860742/is-content-type-mandatory-in-http-post-request), and it is expected that people can extend the HTTP protocol with custom headers that are optional to both clients and servers

- Common invocation protocols like GRPC treat addition of a fields with defaults as backwards compatible ([link](https://learn.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-8.0)), allowing either caller or implementor to omit the field. 
    - The linked documentation for a the .NET GRPC server states "*Adding a field to a request message - Fields added to a request message are deserialized with the default value on the server when not set. To be a non-breaking change, the service must succeed when the new field isn't set by older clients.*". This is similar to the limits around `@unroll`ing abstract method `def`s we have discussed here

- `case class`es themselves are often invocation protocols! 
    - A method taking a single `case class` instance is isomorphic to a method taking a loose parameter list
    - The `case class` definition itself is isomorphic to an abstract method `def`: both specify precisely what typed values the implementor method will have to work with, but now _how_ it will work with those values
    - Thus, adding fields to the `case class` is isomorphic to adding additional parameters to the abstract method `def`
    - If we accept that we should be able to pass newer `case class`es instances with more fields to older versions of code that would simply ignore the newer/unknown fields, why should we reject using a newer abstract method to pass the equivalent loose parameter list to an older version of the concrete implementation, that would ignore the newer/unknown parameters in the same way?

In any protocol-evolution there is a risk of breaking semantic compatibility: it doesn't matter whether you are evolving a Scala abstract-`trait`/implementing-`object`/callsites, or you are evolving a bunch of Protobuf/GRPC schema files and some C++ binary servers and clients that implements it. But the fundamental act of "adding an optional parameter to a protocol that can be ignored by both client or server" is not itself breaking, and in fact seems to be a pretty common pattern/convention in software.

`@unroll` on abstract method `def`s thus matches up with a very common pattern in protocol evolution, just like `@unroll`ing `case class`es matches up with common patterns in schema evolution, and `@unroll`ing concrete method `def`s matches up with common patterns in API evolution. Each of the three challenges can be solved independently of the others, but I think it's a reasonable case to say that solving all three with the same syntax and (almost) the same semantics is a reasonable choice to make

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,934 @@
...
+[Backwards Compatibility](#backwards-compatibility), but nevertheless is a real
+concern due to the requirement that [All Overrides Are Equivalent](#all-overrides-are-equivalent).
+
+It may be possible to loosen this restriction to also allow abstract methods that
+are implemented only once by a final method. See the section about 
+[Abstract Methods](#abstract-methods) for details.
+
+## Major Alternatives
+
+The major alternatives to `@unroll` are listed below:
```
sideeffffect, line 714: This proposal revolves around explicitly annotating items with `@unroll`.
Would a thinkable alternative be that the compiler unrolls all default parameters just by itself, by default?

sjrd: Not thinkable unless you want to break binary and TASTy compatibility of the entire world, which would mean Scala 4.

lihaoyi: It would also probably result in a ton of redundant forwarders that aren't actually necessary, because the compiler cannot reason across published artifact versions to figure out which forwarders are necessary and which are not

lihaoyi:
I have updated the prototype implementation to use the post-pickler/invisible approach to hiding the generated forwarder methods, and updated the doc accordingly

lihaoyi:
@alexarchambault I have published the latest version of `@unroll`, in accordance to this proposal. You can use it in Mill via

```scala
  def scalacPluginIvyDeps = super.scalacPluginIvyDeps() ++ Agg(ivy"com.lihaoyi::unroll-plugin:0.1.12")
  def ivyDeps = super.ivyDeps() ++ Agg(ivy"com.lihaoyi::unroll-annotation:0.1.12")
```

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,933 @@
...
+
+## Summary
+
+This SIP proposes an `@unroll` annotation lets you add additional parameters
+to method `def`s,`class` construtors, or `case class`es, without breaking binary 
+compatibility. `@unroll` works by generating "unrolled" or "telescoping" forwarders:
+
+```scala
+// Original
+def foo(s: String, n: Int = 1, @unroll b: Boolean = true, l: Long = 0) = s + n + b + l
```
lrytz, line 25: `@unroll` missing on the `l: Long` parameter (now that there's `@unrollAll`)

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,933 @@
...
+2. Abstract methods are properly implemented, and no method would fail with an 
+   `AbstractMethodError` when called
+3. We properly forward the necessary argument and default parameter values when
+   calling the respective implementation.
+
+## Proposed solution
+
+
+The proposed solution is to provide a `scala.annotation.unroll` annotation, that
+can be applied to methods `def`s, `class` constructors, or `case class`es to generate
```
sideeffffect, line 207: In the description I've seen only `@unroll` in case class constructor.

Would it be possible to `@unroll` a case class as such? Meaning it would `@unroll` from the first parameter with default value.

lihaoyi: @sideeffffect can you elaborate on what you mean by "*`@unroll` a whole case class*" and how it differs from the behavior described under **Case Classes**?

bjornregnell: I guess this is the idea:
```scala
@unroll case class Data(x: Int, y: Int = 0, z: Int = 0)
```
So it is unrolled from y to z.

lrytz: In the current proposal that would be expressed

```scala
case class Data(x: Int, @unrollAll y: Int = 0, z: Int = 0)
```

bjornregnell: Yes, but I guess the idea is that it may be easier to spot/read if it's annotation on the whole class. I'm not sure.

bjornregnell: Also, I'm still a bit hesitant if the term "unroll" is the best possible, but I have failed to come up with something better. It would be nice if the term said something concrete about what is happening, like "autoOverloaded" or something.

It is not that the name matters very much fpr lib authors, but for newcomers that come across "cryptic" annotations, it can be difficult.

bjornregnell: If I duckduck "unroll" I'm suggested a service to "unsubscribe from emails, instantly" and much other stuff unrelated to programming.

bjornregnell: Perhaps `@compatOverload` can reveal both something about the intent (compatibility) and mechanism (overloading). (I think "unroll" says little about neither intent nor mechanism - that's why I'm brainstorming names...)

bjornregnell: And if the annotation is allowed outside the thing it would cover the typical use case without cluttering the paramlist, so the reading and grokking of the thing is not disturbed:
```scala
@compatOverload 
case class Thing(x: Int, y: Int = 0, z: Int = 0)

@compatOverload 
def thing(x: Int, y: Int = 0, z: Int = 0)
```

sideeffffect: Indeed, that's exactly what I had in mind, that
```scala
@unroll case class Data(x: Int, y: Int = 0, z: Int = 0)
```
would behave as
```scala
case class Data(x: Int, @unrollAll y: Int = 0, z: Int = 0)
```
does currently.

I think many people would prefer this over thinking where exactly in the constructor list to put the `@unrollAll`. I certainly would.

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,933 @@
...
+
+| Date          | Version            |
+|---------------|--------------------|
+| Feb 14th 2024 | Initial Draft      |
+
+## Summary
+
+This SIP proposes an `@unroll` annotation lets you add additional parameters
+to method `def`s,`class` construtors, or `case class`es, without breaking binary 
+compatibility. `@unroll` works by generating "unrolled" or "telescoping" forwarders:
```
sideeffffect, line 21: Regarding the naming, I don't mind very much either way. But _compatibility_ is a concept which doesn't exist within one class, but only as a relationship between that class in two versions of that code base. So maybe it would be better to name the annotation only after the mechanism (which incidentally helps with compatibility). The two Scala mechanisms at play here are "Default Parameters" and "Overloading". If we use them, the name could be
 * `@overloadDefaultParams`
 * `@defaultParamsOverload`
 * `@overloadDefaultParameters`
 * `@defaultParametersOverload`

But I have to admit that it's longer than `@unroll`.

bjornregnell: If you can annotate the whole thing the annotation can be kept on its own line and then the length of the name is not that big of a deal.

bjornregnell: I agree that compatibility intent is difficult after thinking for a while. 

I think I like `@overloadDefaultParam` best as the name of a "whole-thing-annotation".

content/unroll-default-arguments.md
lihaoyi: Main view here is that "*unroll everything*" is most often *not* what you want. 

You only want to `@unroll` the individual parameters that correspond to a new release of the library, which is a usually subset of the default parameters a method or case class. Often methods are defined with default parameters to begin with, and new versions of a library may add more than one default parameter to a method. You only want to `@unroll` the first default parameter for each new version of the library.

It doesn't do any harm to "`@unroll` everything", but it is wasteful in terms of generated code, and thus both compile times and classfile size. This was discussed when the SIP was initially proposed, and we decided to go with `@unroll` on specific parameters rather than `@unroll`ing entire `def` or `case class`es as the default as it better reflects what you would actually want in reality despite being a bit more verbose and intrusive

bjornregnell: Could we then have both the shorter `@unroll` on individual parameters and a more explanatory similar to `@unrollDefaultParams` on the whole thing to cover both use cases?

You say "most often" but I'd assume that if you plan for binary compatibility then as soon as you add things all the params from the first update would need overloaded forwarders, so it seems to me that the use case of "from now on generate forwarders for all added params" would be the most common?

content/unroll-default-arguments.md
lihaoyi: Yes if we annotate the whole method the length does not matter, but if we annotation individual parameters the length does matter.

One option is to specify individual parameter names in a method-level annotation `@unroll("param1", "param2") def foo...` , but that involves duplication of the parameter names and introduces an extra degree of freedom for people to make mistakes.

With the current approach of annotating the parameters themselves, it is important to keep the annotation short. And I'll say again that any one confused about `@unroll` is one click away from a detailed explanation in the Scaladoc, so I don't think we need to have a long and verbose name. 

It's not like people unfamiliar can look at `.pipe` or `.tap` and immediately know what it means either, but because the detailed Scaladoc is one click away the initial confusion is easy to overcome and being concise can take priority over being verbose and spelling out exactly what the functionality does in a long and verbose identifier. I believe the same logic applies to `@unroll`

content/unroll-default-arguments.md
lihaoyi: > You say "most often" but I'd assume that if you plan for binary compatibility then as soon as you add things all the params from the first update would need overloaded forwarders, so it seems to me that the use case of "from now on generate forwarders for all added params" would be the most common?

That is only true if you add one default parameter at a time per stable release of your library. That is sometimes the case, sometimes isn't. Even if no single PR adds multiple default parameters, a stable release often includes multiple PRs. If a stable release includes multiple default parameters, then you should only generate an overload for the first one and not all of them

lihaoyi: Another thing to consider is that `@unrollAll` *still has to specify a parameter to start unrolling from*. Often methods come with default parameters right out of the box, and you do *not* want to unroll any of those. You only want to unroll default parameters added in *subsequent stable releases*, which means you need to specify a parameter from which unrolling will take place, which means putting the annotation on the entire method `def`/`class`/`case class` does not allow us to convey enough information to make that work

content/unroll-default-arguments.md
bjornregnell: Thanks for taking the time to explain all rationale! As usual, you have good answers to everything! 

OK, I can live with `@unroll` on individual params. But we could still *also* have an annotation on the whole thing and the param to that annotation could be from where the unrolling starts, like in:
```scala
@unrollDefaultFrom("y")
case class Thing(x: Int, y: Int = 0, z: Int= 0)
``` 
as a more readable (when applicable) alternative to this:
```scala
case class Thing(x: Int, @unroll y: Int = 0, @unroll z: Int= 0)
```

lihaoyi: >  But we could still also have an annotation on the whole thing and the param to that annotation could be from where the unrolling starts, like in:

I would personally be against this. Having two ways to spell the same thing seems worse than having either one alone. When there's backwards compatibility or migration concerns, having some duplication seems fine, but since we have a chance to specify this from scratch we should really just pick one.

It basically comes down to the age-old tradeoff between "intrusive annotation" or "lookaside table", and the current proposal of "intrusive annotation" is a decent spot in a solution space with no "correct" answer.

bjornregnell: Well, the proposal already has two ways o doing things: `@unroll` and `@unrollAll`. But I agree that it's a matter of taste. I think I prefer that the unrollAll variant is placed on the whole thing with a param from where it starts. That way we both get the cake and can eat it, so to speak.

lihaoyi:
@lrytz I have moved the Abstract Methods section out of the main proposal into the Alternatives section, as discussed during the meeting since we won't be implementing it as part of this proposal

content/unroll-default-arguments.md
```diff
@@ -0,0 +1,934 @@
...
+code to become binary compatible (see [Major Alternatives](#major-alternatives)),
+`@unroll` allows you to write Scala with vanilla `def`s/`class`es/`case class`es, add 
+a single annotation, and your code will maintain binary compatibility as new default
+parameters and fields are added over time. 
+
+`@unroll`'s only constraints are that: 
+
+1. New parameters need to have a default value 
+2. New parameters can only be added on the right
+3. The `@unroll`ed methods must be abstract or final
```
lrytz, line 42:
```suggestion
3. The `@unroll`ed methods must be final
```

lrytz:
The proposal was unanimously "Accepted for implementation" in the May SIP meeting, given these latest changes:
  - `@unroll` is only supported on final methods; the "Abstract Methods" section is now under "Alternatives"
  - `@unrollAll` is also under "Alternatives", so not part of the accepted proposal

bishabosha:
@lihaoyi does the text in the merged file represent the current state of things?

lihaoyi:
@bishabosha yes that's the current state of the approved spec. The Scala 3 compiler plugin in https://github.com/com-lihaoyi/unroll should mostly work, though we'd need to disable to abstract method support and replace it with error reporting  for that and any non-final methods

lihaoyi:
FYI for everyone, @bishabosha will be picking up the work integrating the current POC compiler plugin into the dotty mainline repo

lihaoyi:
I've been playing around with this, and so far hit two minor issues

* https://github.com/scala/scala3/issues/22832
* https://github.com/scala/scala3/issues/22833

Unfortunately due to the nature of the problem, we won't be able to test this out in more realistic workloads "in the wild" for a while, since we would need a Scala 3.7.x-only codebase that is long-lived and widely used enough that binary compatibility of downstream libraries compiled against it becomes a problem, which none of my small libraries satisfy (they are all scala 2/3 cross-compiled) and Mill-on-Scala-3 is too young to satisfy (since it was ported to Scala 3 too recently to have to deal with binary compatible evolution)

content/unroll-default-arguments.md
hamzaremmal: @lrytz was this proposed change part of the accepted spec?

lihaoyi: @hamzaremmal yes, we decided against allowing `abstract` methods because we didn't have a semantics for them that we were confident was correct. So only `final` methods should be unrollable

PR #79, opened 2024-02-19
SIP-62 - For comprehension improvements

KacperFKorban:
## SIP-62 - For comprehension improvements

Author: Kacper Korban (VirtusLab)

### Summary of the proposed changes

**For more details read the committed markdown file.**

This proposal introduces improvements to `for` comprehensions in Scala to improve usability and simplify desugaring. The main changes are:

1. **Starting `for` comprehensions with aliases**: 
   - **Current Syntax**:
     ```scala
     val a = 1
     for {
       b <- Some(2)
       c <- doSth(a)
     } yield b + c
     ```
   - **New Syntax**:
     ```scala
     for {
       a = 1
       b <- Some(2)
       c <- doSth(a)
     } yield b + c
     ```

2. **Simpler Desugaring for Pure Aliases**:
   - **Current Desugaring**:
     ```scala
     for {
       a <- doSth(arg)
       b = a
     } yield a + b
     ```
     Desugars to:
     ```scala
     doSth(arg).map { a =>
       val b = a
       (a, b)
     }.map { case (a, b) =>
       a + b
     }
     ```
   - **New Desugaring**: (where possible)
     ```scala
     doSth(arg).map { a =>
       val b = a
       a + b
     }
     ```

3. **Avoiding Redundant `map` Calls**:
   - **Current Desugaring**:
     ```scala
     for {
       a <- List(1, 2, 3)
     } yield a
     ```
     Desugars to:
     ```scala
     List(1, 2, 3).map(a => a)
     ```
   - **New Desugaring**:
     ```scala
     List(1, 2, 3)
     ```

### DIFF
--- a/content/better-fors.md
+++ b/content/better-fors.md
@@ -133,7 +133,7 @@ This SIP suggests the following changes to `for` comprehensions:
     will be desugared to
 
     ```scala
-    Some(1).map { a =>
+    doSth(arg).map { a =>
       val b = a
       (a, b)
     }.withFilter { case (a, b) =>
@@ -234,7 +234,7 @@ yield a + b
 will be desugared to:
 
 ```scala
-Some(1).map { a =>
+doSth(arg).map { a =>
   val b = a
   (a, b)
 }.map { case (a, b) =>
@@ -319,7 +319,7 @@ yield a + b
 will be desugared to
 
 ```scala
-Some(1).map { a =>
+doSth(arg).map { a =>
   val b = a
   (a, b)
 }.withFilter { case (a, b) =>
@@ -372,4 +372,4 @@ As far as I know, there are no widely used Scala 3 libraries that depend on the 
 2. Github issue discussion about for desugaring: https://github.com/lampepfl/dotty/issues/2573
 3. Scala 2 implementation of some of the improvements: https://github.com/oleg-py/better-monadic-for
 4. Implementation of one of the simplifications: https://github.com/lampepfl/dotty/pull/16703
-5. WIP implementation branch: https://github.com/dotty-staging/dotty/tree/improved-fors
+5. Draft implementation branch: https://github.com/dotty-staging/dotty/tree/improved-fors
--- a/content/better-fors.md
+++ b/content/better-fors.md
@@ -358,9 +358,15 @@ List(1, 2, 3)
 
 ### Compatibility
 
-This change is binary and TASTY compatible since for-comprehensions are desugared in the Typer. Thus, both class and TASTY files only ever use the desugared versions of programs.
+This change may change the semantics of some programs. It may remove some `map` calls in the desugared code, which may change the program semantics (if the `map` implementation was side-effecting).
 
-While this change is forward source compatible, it is not backward compatible, as it accepts more syntax.
+For example the following code will now have only one `map` call, instead of two:
+```scala
+for
+  a <- doSth(arg)
+  b = a
+yield a + b
+```
 
 ### Other concerns
 
--- a/content/better-fors.md
+++ b/content/better-fors.md
@@ -1,0 +1,375 @@
+---
+layout: sip
+permalink: /sips/:title.html
+stage: design
+status: submitted
+title: SIP-62 - For comprehension improvements
+---
+
+**By: Kacper Korban (VirtusLab)**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| June 6th 2023 | Initial Draft      |
+| Feb 15th 2024 | Reviewed Version   |
+
+## Summary
+
+`for`-comprehensions in Scala 3 improved their usability in comparison to Scala 2, but there are still some pain points relating both usability of `for`-comprehensions and simplicity of their desugaring.
+
+This SIP tries to address some of those problems, by changing the specification of `for`-comprehensions. From user perspective, the biggest change is allowing aliases at the start of the `for`-comprehensions. e.g.
+
+```
+for {
+  x = 1
+  y <- Some(2)
+} yield x + y
+```
+
+## Motivation
+
+There are some clear pain points related to Scala'3 `for`-comprehensions and those can be divided into two categories:
+
+1. User-facing and code simplicity problems
+
+    Specifically, for the following example written in a Haskell-style do-comprehension
+
+    ```haskell
+    do
+      a = largeExpr(arg)
+      b <- doSth(a)
+      combineM(a, b)
+    ```
+    in Scala we would have to write
+
+    ```scala
+    val a = largeExpr(b)
+    for
+      b <- doSth(a)
+      x <- combineM(a, b)
+      yield x
+    ```
+
+    This complicates the code, even in this simple example.
+2. The simplicity of desugared code
+    
+    The second pain point is that the desugared code of `for`-comprehensions can often be surprisingly complicated.
+
+    e.g.
+    ```scala
+    for
+      a <- doSth(arg)
+      b = a
+    yield a + b
+    ```
+
+    Intuition would suggest for the desugared code will be of the form
+
+    ```scala
+    doSth(arg).map { a =>
+      val b = a
+      a + b
+    }
+    ```
+
+    But because of the possibility of an `if` guard being immediately after the pure alias, the desugared code is of the form
+
+    ```scala
+    doSth(arg).map { a =>
+      val b = a
+      (a, b)
+    }.map { case (a, b) =>
+      a + b
+    }
+    ```
+
+    These unnecessary assignments and additional function calls not only add unnecessary runtime overhead but can also block other optimizations from being performed.
+
+## Proposed solution
+
+This SIP suggests the following changes to `for` comprehensions:
+
+1. Allow `for` comprehensions to start with pure aliases
+  
+    e.g.
+    ```scala
+    for
+      a = 1
+      b <- Some(2)
+      c <- doSth(a)
+    yield b + c
+    ```
+2. Simpler conditional desugaring of pure aliases. i.e. whenever a series of pure aliases is not immediately followed by an `if`, use a simpler way of desugaring.
+
+    e.g. 
+    ```scala
+    for
+      a <- doSth(arg)
+      b = a
+    yield a + b
+    ```
+
+    will be desugared to
+
+    ```scala
+    doSth(arg).map { a =>
+      val b = a
+      a + b
+    }
+    ```
+
+    but
+
+    ```scala
+    for
+      a <- doSth(arg)
+      b = a
+      if b > 1
+    yield a + b
+    ```
+
+    will be desugared to
+
+    ```scala
+    Some(1).map { a =>
+      val b = a
+      (a, b)
+    }.withFilter { case (a, b) =>
+      b > 1
+    }.map { case (a, b) =>
+      a + b
+    }
+    ```
+
+3. Avoiding redundant `map` calls if the yielded value is the same as the last bound value.
+
+    e.g.
+    ```scala
+    for
+      a <- List(1, 2, 3)
+    yield a
+    ```
+
+    will just be desugared to
+
+    ```scala
+    List(1, 2, 3)
+    ```
+
+### Detailed description
+
+#### Ad 1. Allow `for` comprehensions to start with pure aliases
+
+Allowing `for` comprehensions to start with pure aliases is a straightforward change.
+
+The Enumerators syntax will be changed from:
+
+```
+Enumerators ::= Generator {semi Enumerator | Guard}
+```
+
+to
+
+```
+Enumerators ::= {Pattern1 `=' Expr semi} Generator {semi Enumerator | Guard}
+```
+
+Which will allow adding 0 or more aliases before the first generator.
+
+When desugaring is concerned, a for comprehension starting with pure aliases will generate a block with those aliases as `val` declarations and the rest of the desugared `for` as an expression. Unless the aliases are followed by a guard, then the desugaring should result in an error.
+
+New desugaring rule will be added:
+
+```scala
+For any N:
+  for (P_1 = E_1; ... P_N = E_N; ...)
+    ==>
+  {
+    val x_2 @ P_2 = E_2
+    ...
+    val x_N @ P_N = E_N
+    for (...)
+  }
+```
+
+e.g.
+
+```scala
+for
+  a = 1
+  b <- Some(2)
+  c <- doSth(a)
+yield b + c
+```
+
+will desugar to
+
+```scala
+{
+  val a = 1
+  for
+    b <- Some(2)
+    c <- doSth(a)
+  yield b + c
+}
+```
+
+#### Ad 2. Simpler conditional desugaring of pure aliases. i.e. whenever a series of pure aliases is not immediately followed by an `if`, use a simpler way of desugaring.
+
+Currently, for consistency, all pure aliases are desugared as if they are followed by an `if` condition. Which makes the desugaring more complicated than expected.
+
+e.g.
+
+The following code:
+
+```scala
+for
+  a <- doSth(arg)
+  b = a
+yield a + b
+```
+
+will be desugared to:
+
+```scala
+Some(1).map { a =>
+  val b = a
+  (a, b)
+}.map { case (a, b) =>
+  a + b
+}
+```
+
+The proposed change is to introduce a simpler desugaring for common cases, when aliases aren't followed by a guard, and keep the old desugaring method for the other cases.
+
+A new desugaring rules will be introduced for simple desugaring.
+
+```scala
+For any N:
+  for (P <- G; P_1 = E_1; ... P_N = E_N; ...)
+    ==>
+  G.flatMap (P => for (P_1 = E_1; ... P_N = E_N; ...)) 
+
+And:
+
+  for () yield E  ==>  E
+
+(Where empty for-comprehensions are excluded by the parser)
+```
+
+It delegares desugaring aliases to the newly introduced rule from the previous impreovement. i.e.
+
+```scala
+For any N:
+  for (P_1 = E_1; ... P_N = E_N; ...)
+    ==>
+  {
+    val x_2 @ P_2 = E_2
+    ...
+    val x_N @ P_N = E_N
+    for (...)
+  }
+```
+
+One other rule also has to be changed, so that the current desugaring method, of passing all the aliases in a tuple with the result, will only be used when desugaring a generator, followed by some aliases, followed by a guard.
+
+```scala
+For any N:
+  for (P <- G; P_1 = E_1; ... P_N = E_N; if E; ...)
+    ==>
+  for (TupleN(P, P_1, ... P_N) <-
+    for (x @ P <- G) yield {
+      val x_1 @ P_1 = E_2
+      ...
+      val x_N @ P_N = E_N
+      TupleN(x, x_1, ..., x_N)
+    }; if E; ...)
+```
+
+This changes will make the desugaring work in the following way:
+
+```scala
+for
+  a <- doSth(arg)
+  b = a
+yield a + b
+```
+
+will be desugared to
+
+```scala
+doSth(arg).map { a =>
+  val b = a
+  a + b
+}
+```
+
+but
+
+```scala
+for
+  a <- doSth(arg)
+  b = a
+  if b > 1
+yield a + b
+```
+
+will be desugared to
+
+```scala
+Some(1).map { a =>
+  val b = a
+  (a, b)
+}.withFilter { case (a, b) =>
+  b > 1
+}.map { case (a, b) =>
+  a + b
+}
+```
+
+#### Ad 3. Avoiding redundant `map` calls if the yielded value is the same as the last bound value.
+
+This change is strictly an optimization. This allows for the compiler to get rid of the final `map` call, if the yielded value is the same as the last bound pattern. The pattern can be either a single variable binding or a tuple.
+
+One desugaring rule has to be modified for this purpose.
+
+```scala
+  for (P <- G) yield P  ==>  G
+If P is a variable or a tuple of variables and G is not a withFilter.
+
+  for (P <- G) yield E  ==>  G.map (P => E)
+Otherwise
+```
+
+e.g.
+```scala
+for
+  a <- List(1, 2, 3)
+yield a
+```
+
+will just be desugared to
+
+```scala
+List(1, 2, 3)
+```
+
+### Compatibility
+
+This change is binary and TASTY compatible since for-comprehensions are desugared in the Typer. Thus, both class and TASTY files only ever use the desugared versions of programs.
+
+While this change is forward source compatible, it is not backward compatible, as it accepts more syntax.
+
+### Other concerns
+
+As far as I know, there are no widely used Scala 3 libraries that depend on the desugaring specification of `for`-comprehensions.
+
+## Links
+
+1. Scala contributors discussion thread (pre-SIP): https://contributors.scala-lang.org/t/pre-sip-improve-for-comprehensions-functionality/3509/51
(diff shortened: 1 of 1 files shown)


SethTisue:
about allowing `a = ` as the first clause, there is previous discussion (and links to other previous discussions) at https://github.com/scala/bug/issues/907

nafg:
Just to be clear, is this 3 orthogonal changes or do they build on each other? Just curious.

bishabosha:
Also perhaps some code out there has side effecting map that could be dropped and change semantics?

KacperFKorban:
> Just to be clear, is this 3 orthogonal changes or do they build on each other? Just curious.

Theses can be 3 independent changes. `1.` and `2.` use some of the same new desugaring rules and are implemented in the same commit in the [draft implementation](https://github.com/dotty-staging/dotty/tree/improved-fors), but it is definitely possible to implement just one of them.

KacperFKorban:
> perhaps some code out there has side effecting map that could change semantics if dropped?

Yes, that is possible. But if that's the case, then using `map` for such an operation probably isn't a good idea in the first place.

content/better-fors.md
```diff
@@ -0,0 +1,375 @@
...
+
+```scala
+for
+  a <- doSth(arg)
+  b = a
+  if b > 1
+yield a + b
+```
+
+will be desugared to
```
lihaoyi, line 319: This desugaring seems incorrect. How did `doSth` become `Some`?

content/better-fors.md
```diff
@@ -0,0 +1,375 @@
...
+
+```scala
+List(1, 2, 3)
+```
+
+### Compatibility
+
+This change is binary and TASTY compatible since for-comprehensions are desugared in the Typer. Thus, both class and TASTY files only ever use the desugared versions of programs.
+
+While this change is forward source compatible, it is not backward compatible, as it accepts more syntax.
```
lihaoyi, line 363: I think it's pretty clear that many of the changes in this proposal can change the behavior of user code. We could argue that such user code is badly written, but nevertheless it is a concern, and we should be rigorous about listing out what kinds of user code would have their behavior changed by this proposal.

Given that we are already choosing to potentially change the behavior of user code via these changes, I wonder if we should take it further. e.g. rather than generating

```scala
    Some(1).map { a =>
      val b = a
      (a, b)
    }.withFilter { case (a, b) =>
      b > 1
    }.map { case (a, b) =>
      a + b
    }
```

we could generate something like

```scala
    Some(1).flatMap { a =>
      val b = a
      if (!(b > 1)) Option.empty
      else Some(a + b)
    }
```

If we're going to be changing the desugaring and potentially breaking user code, I feel like it makes sense to change it directly to a desugaring that makes sense, rather than changing it half-way and paying the cost in breakage without reaping the full benefits of a simpler desugaring

content/better-fors.md
KacperFKorban: Yes, thanks for pointing that out. Fixed now. :smiley:

content/better-fors.md
JD557: In that particular example, how would the compiler know that it should desugar to `Option.empty` instead of `Some.empty`? 🤔

KacperFKorban: I might be wrong, but AFAIK this would require some refactoring around desugaring of `for`s. Right now, the desugaring is purely syntactic and it deals with untyped trees. So to implement the desugaring based on `empty` in an elegant way, `for`s would have to become fully fledged (typed) trees.

lihaoyi: Maybe we could provide `implicitly[Empty[Some[Int]]]` or something?

KacperFKorban: Hmmm, this would still require the type info. Implementing `empty` as a method on the class instead of the module might work, but that seems the opposite of elegant.

I think that this change might be way less disruptive than it may seem. Maybe we could run the open community build with vs without the feature flag and check the results. (It's far from good testing, but better than speculation)

KacperFKorban: Also, I updated the compatibility part of the doc to point to possible semantics changes (based on Your comment)

lihaoyi: Yeah I'm not saying we can't change things, just that we need to be careful in doing so. Some breakage is worse than others, and the community build would be a great way to check how severe the theoretical breakage actually is in reality

SethTisue: Perhaps the old desugaring could remain supported under a flag.

odersky: > Hmmm, this would still require the type info. Implementing empty as a method on the class instead of the module might work, but that seems the opposite of elegant.

In fact we already have it. There is an `empty` method on most (all?) stdlib collections. But we also need a singleton method and we don't have that (the `Some` in the last line of the example)
```scala
Some(1).flatMap { a =>
      val b = a
      if (!(b > 1)) this.empty
      else this.singleton(a + b)
    }
```
Otherwise we'd need the types, and we definitely do not want that! The untyped translation of for expression is one of the major distinguishing features of Scala.

lihaoyi: If we wanted to go the implicit route, we might be able to do so via an untyped desugaring as well. Something like:

```scala
trait Empty[T]{
  def apply(): T
}
object Empty{
  def of[T](v: => T)(implicit e: Empty[T]) = e
}


trait Single[T]{
  def apply(t: T): T
}
object Single{
  def of[T](v: => T)(implicit e: Empty[T]) = e.value
}

Some(1).flatMap { a =>
  val b = a
  if (!(b > 1)) Empty.of(this).apply()
  else Single.of(this).apply(a + b)
}
```

This could be useful if e.g. we wanted to provide the `Single.of` value without needing to add a `def singleton` on every type, allowing us to move forward without modifications to existing standard library classes

KacperFKorban: Being able to make this change independent from unfreezing the stdlib would be great. But I'm not sure if it's that straightforward to create those typelcasses. For the sake of inference, we would like the typeclass to be contravariant (we want `Empty[Option[?]] <: Empty[Some[?]]`). But then IMHO we can't implement it, since it's **not** the case that `∀(Opt <: Option[?]). None <: Opt`. On the other hand, if we make the typeclass invariant, then it will just infer the more specific type. (But maybe I'm missing something obvious)

One additional concern I have here is that `this` isn't a correct lookup. In this case, it's an owner class/package reference. To get the "monad" lookup, we would have to lift the first `GenFrom` expression and use it as a lookup afterwards, like so:

```scala
val a$lifted = Some(1)
a$lifted.flatMap { a =>
  val b = a
  if (b > 1) a$lifted.singleton(a + b)
  else a$lifted.empty
}
```

If done correctly, the lifted `val` should always be immediately before the reference, but might add a `lazy` to be safe.

bjornregnell:
Related about for-comprehensions but with focus on performance of for-do:

https://august.nagro.us/scala-for-loop.html

A quote from the above blog post by @AugustNagro  :
"I wonder if for-do could be updated to behave like the loop() overloads, falling back to foreach(..) when required. I’m sure it would make some codebases a lot faster."

If we revisit de-sugaring of for-comprehensions perhaps we should also include possible performance optimizations to for-do in the same go?

bjornregnell:
> We simply make `foreach` on `Range` an inline method.

Aha! If it can be solved in the std library rather than by the compiler this is yet another cool evidence about the power of the language :)  (Again so glad that I'm doing Scala).

content/better-fors.md
bishabosha: to avoid the by name type, as long as the value is bound to a variable so you could have something like `Empty.of[b.type]`

odersky:
I think changes (1) and (2) are definitely worthwhile. 

(3) (eliding map) is tricky since it changes behavior if the function passed to `map` is side-effecting. A discussion took place in https://github.com/lampepfl/dotty/pull/16703 which is a (closed) PR linked from the SIP. 

I wonder whether we should wait with (3) until we have a better grip on effects. There are a number of places where this would be beneficial. But it's clearly a long-term goal only.

Or, try a cooperative solution: If the final expression is `x <- xs yield x`, translate that to `map(Identity)` with a special library defined `Identity` object. Then the _implementations_ of `map` can decide to short-circuit identity.

JD557:
> Or, try a cooperative solution: If the final expression is x <- xs yield x, translate that to xs.map(Identity) with a special library defined Identity object with an identity apply method. Then the implementations of map can decide to short-circuit identity.

I wonder if this behavior would make more sense in `mapConserve`, with `for` simply desugaring the last `map` to `mapConserve` instead.
Although this would obviously be a huge breaking change.

odersky:
About the general problem of translating `val`s in a natural way even in the presence of conditionals: I thought up a scheme that fits all the criteria, but it is not simple. 

The criteria being:

 1. Untyped translation
 2. No requirement that methods taking part in the translation have any specific type. 
 3. It should be possible to define a lifting behavior that simply maps a for expression to an AST describing its contents.

The new scheme does not need `withFilter` anymore, but needs three other functions instead (I am using collection meaning for intuition):

 - `mapDefined`: Takes a function and returns all defined elements obtained by applying the function to the receiver collection. What is "defined" is up to the implementor of `mapDefined` to decide.
 - `flatMapDefined`: Takes a function and concatenates all defined collections obtained by applying the function to the receiver collection.  
 - `applyFilter`: Takes a predicate `p` and a body returning a collection, and marks all elements for which `p` is false as undefined. 

## Example

The following for expression mixes generators, val definitions, and conditionals:
```scala
    for
      x <- List(1, 2, 3)
      y = x + x
      if x >= 2
      i <- List.range(0, y)
      z = i * i
      if z % 2 == 0
    yield
      i * x
```
The expression would be translated to the following code:
```scala
    val xs = List(1, 2, 3)
    xs.flatMapDefined: x =>
      val y = x + x
      xs.applyFilter(x >= 2):
        val is = List.range(0, y)
        is.mapDefined: i =>
          val z = i * i
          is.applyFilter(z % 2 == 0):
            i * x
```
I verified that the two expressions give the same results if the three methods are defined like this:
```scala
extension [A](as: List[A])

  def applyFilter[B](p: => Boolean)(b: => B): Option[B] =
    if p then Some(b) else None

  def flatMapDefined[B](f: A => Option[IterableOnce[B]]): List[B] =
    as.flatMap: x =>
      f(x).getOrElse(Nil)

  def mapDefined[B](f: A => Option[B]): List[B] =
    as.flatMap(f)
```
Here we use an `Option` type for tracking definedness. That's not the only possible choice. here is another one:
```scala
object UNDEFINED

extension [A](as: Vector[A])

  def applyFilter[B](p: => Boolean)(b: => B): B | UNDEFINED.type =
    if p then b else UNDEFINED

  def flatMapDefined[B](f: A => IterableOnce[B] | UNDEFINED.type): Vector[B] =
    as.flatMap: x =>
      f(x) match
        case UNDEFINED => Nil
        case y: IterableOnce[B] => y

  def mapDefined[B](f: A => B | UNDEFINED.type): Vector[B] =
    as.flatMap: x =>
      f(x) match
        case UNDEFINED => Nil
        case y: B => y :: Nil
```
Or, as a third possibility, `mapDefined`, `flatMapDefined` and `applyFilter` could simply generate ASTs that represent their applications.

## Discussion

We can pick old or new scheme depending on the availability of `mapDefined`, `flatMapDefined`, and `applyFilter` _vs_`withFilter` on the first generator collection. If there are no conditionals, the new scheme would generate exactly the same code as under points (1) and (2) of this SIP. 

The upside of this scheme is that we get nice untupled vals even in the presence of conditionals. The downside is that instead of `withFilter` we need three new methods to implement that scheme. On the other hand, `withFilter` turns out to be quite hard to implement since it has to fuse with any following `map`, `flatMap`, `withFilter`, and `foreach` call. So while the interface comprising three methods instead of one is more complex, the  actual implementation might be simpler. 

## Formal translation rules for the new scheme. 

The syntax of for expressions is described as follows:
```
forExpr ::= for F yield E
```
where
```
F ::= x = E; F
      x <- E; G
G ::= []
      x = E; G
      if E; G
      x <- E; G
```
The translation scheme is defined with semantic brackets `{...}`. There are two closely related forms, depending whether we see a prefix `F` up to the first generator or a general sequence `G` following it:

 - `{ for F yield E }c`    where c = undefined
 - `{ for G yield E }c`    where c is a reference to the generator preceding the G sequence

The translation is defined as follows:
```
{ for [] yield E }c              =  E
{ for x = Ex; G yield E }c       =  val x = Ex; { for G yield E }c
{ for if Ep; G yield E}c         =  c.applyFilter(Ep)({ for G yield E }c)
{ for x <- Ex G yield E }c       =  val c1 = Ex; c1.BIND(x => { for G yield E }c1)    (c1 fresh)

      where BIND = flatMapDefined if isGen(G), isFilter(G)
                 = mapDefined     if !isGen(G), isFilter(G)
                 = flatMap        if isGen(G), !isFilter(G)
                 = map            if !isGen(G), !isFilter(G)

                 isFilter(if E; S)
                 isFilter(x = E; S)      if  isFilter(S)

                 isGen(x <- E; S)
                 isGen(x = E; S)         if isGen(S)
                 isGen(if E; S)          if isGen(S)

**Comment by lihaoyi** [comment]
I like the `mapDefined`/`flatMapDefined`/`applyFilter` encoding.

As far as I can tell, it basically is the "minimal" encoding necessary to represent the for comprehension. The various Some/None/Empty/Singleton/etc. things we discussed earlier are collapsed relatively neatly into those three operations.

And as a result of being the minimal encoding of the for comprehension, it seems to be the most direct as well: the for comprehension becomes this set of nested function calls. No tuple packing/unpacking weirdness like the status quo, no ad-hoc if-else-empty conditionals like some of the alternate proposals.

I don't think it's any more complex than what we have now. Rather, it takes a bunch of complexity in the desugaring and moves it into three core methods, allowing the desugared code to be far simpler and allowing freedom to implement those three core methods in a wider variety of ways

File: content/better-fors.md [hunk 3]
```diff
@@ -0,0 +1,381 @@
...
+    in Scala we would have to write
+
+    ```scala
+    val a = largeExpr(b)
+    for
+      b <- doSth(a)
+      x <- combineM(a, b)
+      yield x
+    ```
+
```
lihaoyi, line 54: Elsewhere, we discussed how making the final `map` optional would be a breaking change. But what if we went the haskell route, and only removed the final `map` if the final `yield` is also removed?


    ```scala
    val a = largeExpr(b)
    for
      b <- doSth(a)
      combineM(a, b)
    ```

1. This is currently invalid syntax, so no backwards compat concerns
2. We get to remove the useless `map` from the generated code
3. We *also* get to remove the useless `yield` from the source code!
4. The correspondence between the `map` and `yield` is direct. No "magic" map-removal depending on the shape of the final expression
5. The final source code is shorter and more concise
6. The final source code looks more like other languages with similar syntax, e.g. Haskell's `do` or F#'s `return!` in computation expressions

KacperFKorban: I did consider this change, but I didn't manage to get it to work in the parser. (I tried going in the [reflect-like direction](https://github.com/lampepfl/dotty/compare/main...dotty-staging:dotty:monadic-blocks#diff-ff13049a579bd41563e00e4391bcf7a18d9345b754683c44b2d3e971f3fd150a) to get better syntax for this case)
But if we were to be able to distinguish a pattern from an expression, then yet another problem is automatically solved -- we can get rid of the unnecessary `_ <-` with monadic 'unit' expressions.

One workaround from a [pre-SIP](https://contributors.scala-lang.org/t/pre-sip-improve-for-comprehensions-functionality/3509) is to have a `yield <- monadicExpr` syntax.

odersky:
About the translation of for expressions with ifs. I omitted before patterns in generators and in particular `case p <- expr` generators which translate to `withFilter` applications. Here are translation rules that handle these as well:
```
{ for [] yield E }c              =  E
{ for p = Ep; G yield E }c       =  val p = Ep; { for G yield E }c
{ for if Ep; G yield E}c         =  c.applyFilter(Ep)({ for G yield E }c)
{ for p <- Ep; G yield E }c      =  val c1 = Ep; c1.BIND{ case p => { for G yield E }c1 }    (c1 fresh)
{ for case p <- Ep; G yield E }c =  { for $x <- Ep; if $x match case p => true case _ => false; p = $x@RuntimeChecked; G yield E }c
```
For symmetry could also add case binders `case p = e`, which could translate analogously to `case p <- e`:
```
{ for case p = Ep; G yield E }c  =  { for $x = Ep; if $x match case p => true case _ => false; p = $x@RuntimeChecked; G yield E}c
```

odersky:
About final `map` elision. I think I now prefer the original proposal (3) over the alternatives to change the syntax. Side effecting maps should be extreme outliers, and should not be something we wish to keep at all costs. So I believe changing the spec is fine.

content/better-fors.md
bishabosha: would it be controversial to reuse `return` as in
```scala
for x <- List(1,2,3) return List(x, x)
```

bjornregnell: I would be against this for several reasons, including: 1) yet another syntax to learn, but worse: 2) my experience is that many learners have all sorts of connotations on the `return` keyword with all sorts of misconceptions about how it works in Scala so I think it is best to say "don't use return" and for some it takes a while not to end everything with return if they are used to that...

lihaoyi:
I'm not sure side effecting `map`s are such outliers. At the very least, the `map` is needed to force the result of a previous lazy `withFilter` call. That surely must count as a side effect? 

There are also the various Future/Streaming libraries where the final `map` does affect the topology of the computation graph, albeit in a probably benign way

odersky:
`withFilter` won't be a problem since it is only used in cases like this:

```scala
for 
  x <- xs
  if p(x)
yield 
  x
```
But in that case we would not drop the `map`.

KacperFKorban:
I think that the desugaring proposed by @odersky will result in a cleaner scheme than my suggested improvements. The only downside is that it is more disruptive, since it requires implementing new functions.

So what should be the steps now?
Do we draft an alternative SIP for this approach and let the SIP committee decide?
Do we consider this SIP as a short-term solution and wait for the stdlib to unfreeze with the new scheme? (I think that at least (1) from this SIP would be nice to have decently quickly)

odersky:
I am open merging the proposals into one, but we could also have two different ones. As far as I can see, the original SIP scheme is a subset of my proposed extensions, so it could go in separately. On the other hand, maybe it's more economic to discuss everything together.

I believe even the extended the SIP can in principle be adopted without waiting for stdlib to be unfrozen. It's just that its improvements would be more useful if they could be used with stdlib collections.

bishabosha:
> That is a breaking change similar to the ones discussed in this proposal, as we have no guarantees that withFilter is pure, even though by convention is typically is.

good point, its only a source breaking change

lrytz:
> Untyped translation

> We can pick old or new scheme depending on the availability of `mapDefined`, `flatMapDefined`, and `applyFilter` _vs_`withFilter` on the first generator collection

So the first generator would be typed to figure out which desugaring to apply?

lihaoyi:
The consensus in the SIP committee was that we would prefer the more aggressive change, if it is possible in the short-medium term, rather than making an incremental change now and then possibly making another change a short time later. But that is all contingent on whether or not the more aggressive change, i.e. the desugaring proposed by @odersky, is viable.

> So the first generator would be typed to figure out which desugaring to apply?

@odersky this point is worth discussing. AFAIK, one of the goals of many of these desugarings is to make them purely syntactic. The alternate proposal you made would require typechecking the first generator in order to decide how to perform the desugaring.

1. Is that going to be an implementation issue? Given the desugaring has always been syntactic rather than type-driven, this  would be a major shift in how/when such a desugaring can take place, and it seems like something that may have require significant changes in the compiler implementation.

2. Is it possible to workaround the requirement for typechecking? e.g. could we provide some kind of transparent inline extension methods `mapDefined`/`flatMapDefined`/`applyFilter` that expand to the previous desugaring, such that the compiler can naively desugar to `mapDefined`/`flatMapDefined`/`applyFilter`, and have subsequent typechecking decide whether a defined implementation exists, or whether we use the transparent inline extensions to fall back to the current behavior using the current methods?

odersky:
@lihaoyi Yes, I think something like that will be needed. Using inline extension methods for that is an interesting idea!

lrytz:
> 1. Given the desugaring has always been syntactic rather than type-driven

I beliee we had a similar situation in Scala 2.8 where `filter` was replaced by `withFilter`, but `filter` was still supported?

odersky:
> I believe we had a similar situation in Scala 2.8 where filter was replaced by withFilter, but filter was still supported?

Yes indeed. At the time we did not have inline extension methods, so we had to base it on membership. 

With inline extension methods, it's still a bit tricky. Here's a possible scheme:

 1. Instead of desugaring a for expression directly wait until typing. When hitting the first generator of a for expression that's followed by a filter, try to expand it with `mapDefined` or `flatMapDefined` . If that works, expand the rest of the for expression with the new scheme.
 2. Otherwise, try to translate with `withFilter`. If that works, expand the rest of the for expression with the old scheme. (including the improvements in the core proposal).
 3. If neither works, report any type errors with the new scheme.

With extension methods we can retro-fit the new scheme to existing libraries such as stdlib.

bishabosha:
@kyouko-taiga I also updated the stage to implementation to make it a valid state

KacperFKorban:
FYI I updated the implementation of the original (first) schema with some more tests and rebased it onto `main`: https://github.com/dotty-staging/dotty/tree/improved-fors
I also created a draft implementation of the syntax changes in the `scalameta` parser: https://github.com/KacperFKorban/scalameta/tree/better-fors

What are the next steps here? Is there an official summary of the SIP committee decision?

lihaoyi:
@KacperFKorban apologies i forgot to update this. This proposla was accepted for implementation and experimentation, where it will be available behind a flag for a period before a second vote to confirm it

KacperFKorban:
@lihaoyi no worries.
In that case, here is a PR with the experimental implementation: https://github.com/scala/scala3/pull/20522

KacperFKorban:
Hi, I have another proposition for improving for-comprehensions.
I don't know if I should ask this here or create a new discussion, so if this is the wrong place, let me know.

I want to fix the problem that arises from the use of the following code:

```scala
//> using scala 3.3.3
//> using lib "dev.zio::zio:2.1.5"

import zio.*

def loop: Task[Unit] =
  for
    _ <- Console.print("loop")
    _ <- loop
  yield ()

@main
def run =
  val runtime = Runtime.default
  Unsafe.unsafe { implicit unsafe =>
    runtime.unsafe.run(loop).getOrThrowFiberFailure()
  }
```

This kind of effect loop is pretty commonly used in Scala FP programs and often ends in `yield ()`.

The problem with the desugaring of this for-comprehensions is that it leaks memory because the result of `loop` has to be mapped over with `_ => ()`, which often does nothing.

A possible solution that I want to suggest and possibly add to the `betterFors` language extension is to remove these unnecessary `map` calls generated from `yield ()` when the last binding also has type `Unit`.

A possible approach could be to add sticky keys to the `map` calls generated from for-comprehensions and add a new phase after. That phase would then check for every such `map` call if the argument is equivalent to `_ => ()` and the previous binding is also of type `Unit`.

A naive PoC: https://github.com/KacperFKorban/dotty/commit/31cbd4744b9375443a0770a8b8a9d16de694c6bb

kyouko-taiga:
Since the design of this SIP has been accepted, additional changes should either be part of a revision or proposed as a separate SIP (possibly after a Pre-SIP discussion). Given the current status if this SIP, I think that the latter option would be the most appropriate.

KacperFKorban:
@kyouko-taiga Sure, in that case, I'll create a pre-SIP and try to gather some feedback.

unkarjedy:
Are there any expected changes in the implementation of this SIP in 3.6.*/3.7.*?
Is there any understanding when this feature could be released?

KacperFKorban:
There most likely will be changes to the implementation, but they are only briefly sketched on a draft branch right now. I will do the full implementation, once this (https://github.com/scala/improvement-proposals/pull/99) change to the SIP is accepted.

PR #80, opened 2024-02-20
SIP-NN - Scala 3 Macro Annotations

nicolasstucki:


content/macro-annotations.md
```diff
@@ -0,0 +1,349 @@
...
+
+The `MacroAnnotation` annotation trait would be defined as follows:
+
+```scala
+package scala.annotation
+import scala.quoted.*
+
+/** Annotation classes that extend this trait will be able to transform the annotated definition. */
+trait MacroAnnotation extends StaticAnnotation:
+  /** Transforms a tree `definition` and optionally adds new definitions.
```
lihaoyi, line 140: Cask and MainArgs are perhaps one of the more common libraries using "fake" macro annotations today, and I suspect they wouldn't be implementable using this current proposal as is

That because they require that multiple `@annotated` methods be considered to generate a single combined routing object. In MainArgs you can have multiple `@main` methods, and in Cask you can have multiple `@cask.get`/`@cask.post`/etc. routes. 

In both cases, we currently have a expression-macro that looks at the class body, looks for all the `@annotated def`s within it, and generates a single expression that references all of them.

Is there some way we can support that sort of use case?

content/macro-annotations.md
```diff
@@ -0,0 +1,349 @@
...
+
+
+### Feature Interactions
+
+
+#### Typer
+As we transform the definitions after pickling, the typer only needs to be able to type the reference to the macro annotation.
+
+#### Separate/incremental compilation
+As we transform the definitions after pickling, separate/incremental compilation is not affected by the macro annotations.
```
sjrd, line 288: What if a macro annotated member of a `trait` generates a new `private val/var/lazy val` in the trait? That would still affect classes extending that trait, in a way that incremental compilation should be aware of.

content/macro-annotations.md
```diff
@@ -0,0 +1,349 @@
...
+definitions will align with source and typer as it does for non-annotated classes.
+
+
+## Alternatives
+
+### Change the order of evaluation of nested macro annotations
+There might be reasons to evaluate the inner macro annotation first, then the outer one.
+In that case, we have two options: we reverse the order of evaluation, or we add the ability to specify the order of evaluation of each macro annotation.
+
+### Allow new definitions to be macro annotated
```
sjrd, line 319: I don't understand what this means.

content/macro-annotations.md
bishabosha: maybe an aggregate version that only returns new trees to go at the end of the surrounding scope?

content/macro-annotations.md
nicolasstucki: This refers to new definitions created by the macro expansion.

content/macro-annotations.md
nicolasstucki: I am not sure how we would support these. They might need to be disallowed.

content/macro-annotations.md
nicolasstucki: Maybe we could have a macro annotation `@flask`  that looks for normal annotations `@app.route(...)` in its scope 

```scala
@flask def app: Flask = Flask()

@app.route("/")
def index() = "Index Page"

@app.route("/hello")
def hello() = "Hello, World"
```

it could then generate a `main` as follows
```scala
class app {
  <static> def main(args: Array[String]): Unit = {
    app
      .registerRoute("/", index)
      .registerRoute("/hello", hello)
}
```

nicolasstucki: We might also have it as
```scala
@flask val app: Flask = Flask()
```
and modify it to
```scala
val app: Flask = 
  Flask()
    .registerRoute("/", index)
    .registerRoute("/hello", hello)
```

nicolasstucki: Another approach that would follow closer the module initialization semantics of the Python version might be 
```scala
object app extends Flask {

  @route("/")
  def index() = "Index Page"

  @route("/hello")
  def hello() = "Hello, World"
}
```

where the `@root` macro annotation assumes it is inside a `Flask` instance and generates something like

```scala
object app extends Flask {

  val _ = (this: Flask).registerRoot("/" , index)
  def index() = "Index Page"

  val _ = (this: Flask).registerRoot("/hello", hello)
  def hello() = "Hello, World"
}
```

soronpo:
Please open a thread on contributors.scala-lang.org to gain community feedback, and link it to this proposal.

nicolasstucki:
> Please open a thread on contributors.scala-lang.org to gain community feedback, and link it to this proposal.

Here is the contributors thread: https://contributors.scala-lang.org/t/scala-3-macro-annotations-sip-63-discussions/6593

content/macro-annotations.md
nicolasstucki: I believe that macro annotations might not be the best tool to get an elegant solution to this particular problem. Based on these examples, __inline traits__ ([report](https://infoscience.epfl.ch/record/303591?ln=en), [prototype](https://github.com/scala/scala3/pull/17329)) are a much better fit. 

<details><summary>Cask with inline traits</summary>

```scala
object MinimalApplication extends MainAutoRoot {
  @get("/")
  def hello() = "Hello World!"

  @post("/do-thing")
  def doThing(request: Request) = request.text().reverse

  // generated code
  // def routes: Route = Seq(
  //   Route("hello", hello),
  //   Route("doThing", doThing, ...),
  // )
}
```

```scala
inline trait MainAutoRoot {
   def routes: Route = generateRoutes // macro `generateRoutes` finds the routes defined in the class that extends `Routes`
}

trait MainRoot {
  def roots: Roots
  def main(...) = // do something with routes
}
```

</details>  

> 1. Who defines the generateDefinition logic? Is it on the enclosing template? Or some property of bar or qux?

In an inline trait, it would be the trait itself.

> 2. If we need to generate multiple definitions, e.g. we have an object foo that contains both @cask.gets and MainArgs @main methods, how do we ensure the @cask.gets all get passed to Cask's generateDefinition logic, while all the MainArgs @mains get passed to MainArgs' generateDefinition logic?

With inline traits it becomes simpler because there is a single point of expansion for the definition. The definition expansion can collect all (non-macro) annotated definitions in scope.

lihaoyi: @nicolasstucki I think using an inline trait sounds reasonable. But if that's our stance, then that basically cuts out `Example 3: Main annotation` from the use cases of macro annotations: it doesn't make sense to say macro annotations provide a way to provide a worse version of MainArgs. We'll need to find other use cases to replace it, to make sure that macro annotations have enough motivating use cases to be worth it

lihaoyi:
One question that came up in the last SIP meeting is whether or not it's possible to implement https://github.com/scala/improvement-proposals/pull/78 using macro annotations, and what it would look like if we did. @nicolasstucki do you have any insights there? We already have full implementations and test cases in the https://github.com/com-lihaoyi/unroll/ repo, so it should be possible to try it out using macro annotations and see if it can be made to work

nicolasstucki:
> One question that came up in the last SIP meeting is whether or not it's possible to implement https://github.com/scala/improvement-proposals/pull/78 using macro annotations, and what it would look like if we did. @nicolasstucki do you have any insights there?

I wrote and [example implementation](https://github.com/scala/scala3/pull/19997) of a strip down version of https://github.com/scala/improvement-proposals/pull/78 for unrolling the last parameters of a `def`. This implementation shows that we have access to all the we need to construct the definition new definition. 

However, we still have some limitations in the current implementation, such as missing the parameter macro annotations (we are woking on a prototype and should have something soonish). We might also me missing some helper functions in the reflection API to help us deal with this transformation more elegantly (for example a simple way to get the `default` arguments).

I have not tried with constructors yet. We will have a similar situation for class constructors. But, __case class constructors parameter unrolling might not fit the current specification of macro annotations__. As the annotation is on the parameter, the transformed tree will be the constructor of the case class, this only allows us to add methods in the case class itself but not in the compassion object.

content/macro-annotations.md
nicolasstucki: While it does not work for the particular API design choices of `MainArgs`, it allows for many other kinds of main annotation use cases. For example, it can be used to implement the generalized main annotation in https://github.com/scala/scala3/pull/19937.

nicolasstucki:
@jchyb and  @hamzaremmal will take over the development of this SIP.

smarter:
This proposal was discussed at the last SIP meeting, but there was concern that the current proposal is currently too limited and did not have enough usecases as-is:
- The vote for accepting it lead to a 4/4 split.
- The committee unanimously voted against rejecting the proposal.

In particular, the fact that existing libraries like cask and mainargs cannot be ported to work as-is without requiring extra annotations seems problematic. If we look at the cask example from @lihaoyi  again:

```scala
// Cask
object MinimalApplication extends cask.MainRoutes{
  @cask.get("/")
  def hello() = {
    "Hello World!"
  }

  @cask.post("/do-thing")
  def doThing(request: cask.Request) = {
    request.text().reverse
  }

  initialize()
}
```

The current proposal would require an extra annotation on the `object` definition to gather all the method-level annotations. 

To avoid this extra level of boilerplate, I'd like to suggest we introduce "inheritable macro annotation", concretely if we define MainRoutes in cask as:
```scala
class caskGen extends InheritableMacroAnnotation { ... }
@caskGen trait MainRoutes { ... }
```
Then we should desugar:
```scala
object MinimalApplication extends cask.MainRoutes
```
into:
```scala
@caskGen object MinimalApplication extends cask.MainRoutes
```

I believe inheritable macro-annotations could be useful in many situations. For example, if I want to memoize instances of an open class hierarchy, I'd like to be able to define:
```scala
class memo extends InheritableMacroAnnotation { ... }
@memo trait Base
```
so that anyone inheriting from `Base` automatically gets a memoized class without having to remember to put the annotation on the class.

odersky:
We want to see how much of the use cases would be better addressed with inline traits. 

See: https://github.com/scala/scala3/pull/20254
and also: https://github.com/scala/scala3/pull/21950

for current draft designs

smarter:
Based on these documents, I think that inline traits could be helpful but do not cover all usecases. My understanding of inline traits (discarding how they interact with type specialization which I think is orthogonal here), is that I can define something like
```scala
inline def macroCallA() = ...
inline def macroCallB() = ...
inline trait Foo:
  macroCallA()
  def foo = macroCallB()
class Bar extends Foo:
   otherStuff()
   def ... = ...
```
and that this expands to something like:
```scala
class Bar:
  macroCallA()
  def foo = macroCallB()
  otherStuff()
   def ... = ...
```

In other words, the macro calls that appear inside an inline trait don't get executed in the inline traits (because in an inline context we don't recursively inline), but in the class that extends that inline trait. I believe this is enough to cover usecases like cask, but this doesn't seem to be enough to handle usecases like memoization.

To memoize class construction, I need to  modify the `apply` (and possibly `fromProduct`) method defined in its companion object, with inline traits this would mean that instead of writing:
```scala
@memo case class Foo(...)
```
I would have to write:
```scala
case class Foo(...)
object Foo extends Memo
```
which is significantly more boilerplate in practice. This could be solved if we had some mechanism such that extending a trait from a class would automatically extend a "companion trait" from the companion object, an idea which has come up a fw times on contributors already: https://contributors.scala-lang.org/t/pre-sip-companion-traits/6555


<hr/>

Another usecase presented in the SIP which doesn't seem to be addressed by inline traits is macro annotations on defs such as:
```scala
@memoize
def fib(n: Int): Int =
  if n <= 1 then n else fib(n - 1) + fib(n - 2)
```
which desugars in the SIP into:
```scala
private val fibCache$1: scala.collection.concurrent.Map[Int, Int] = ...
def fib(n: Int): Int = ...
```

Here there is no class to extend (`def fib` could even be in a local context, or at the top-level), so inline traits don't seem like they would fit.

lihaoyi:
I think the `@memoize` use case is fine, but as I mentioned before what is lacking is some real use cases in libraries or frameworks to really prove out the design of the language feature.

Almost all the use cases I'm aware of that naively you would reach for macro annotations, would not be supported by this feature:

1. Adding members that are visible to the typer, e.g. `@Lenses` from https://github.com/optics-dev/Monocle or other forms of optics, since the proposal does not allow us to add new members

2. com-lihaoyi-style "fake macro annotations" from Cask or MainArgs, since the proposal does not have any mechanism for "aggregating" the generated code in an idiomatic way

3. `@enableIf` from https://github.com/ThoughtWorksInc/enableIf.scala or `@ifdef` from https://github.com/eed3si9n/ifdef are also not supported, because it these proposed annotations have to run after typer

4. `@unroll` from https://github.com/scala/improvement-proposals/pull/78 decided against using these annotations, because it could provide a better API with inline annotations on each paramter

If the language feature is limited to supporting toy use cases like Scala3's `@main` or `@memoize`, I don't think that's nearly enough when you consider the use cases that someone would naively imagine macro annotations _should_ support (e.g. based on their experience in Scala 2). Even if we can't support _all_ library use cases, we should aim to support at least _some_ of them, but as far as I can tell this proposal is currently 0/4 in terms of supporting what people are doing with macro annotations and compiler plugin annotations

threeseed:
Would add [Slinky](https://slinky.dev) to the list.

I ended up having to write an [SBT Plugin](https://github.com/shadaj/slinky/pull/736) to transform the code in order to get this (barely) working on Scala3. 

Given how common annotations are in Python, Java etc and well loved by beginners you would think they would get more appreciation in Scala.

PR #81, opened 2024-03-11
SIP-64: Improve the Syntax of Context Bounds and Givens

odersky:


content/typeclasses-syntax.md
```diff
@@ -0,0 +1,380 @@
...
+```
+Since we don't have a name for the `Monoid` instance of `A`, we need to resort to `summon` in the body of `reduce`:
+```scala
+def reduce[A : Monoid](xs: List[A]): A =
+  xs.foldLeft(summon Monoid[A])(_ `combine` _)
+```
+That's generally considered too painful to write and read, hence people usually adopt one of two alternatives. Either, eschew context bounds and switch to using clauses:
+```scala
+def reduce[A](xs: List[A])(using m: Monoid[A]): A =
+  xs.foldLeft(m)(_ `combine` _)
```
johnynek, line 72: m.unit

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,380 @@
...
+    def unit: A
+```
+and a `reduce` method defined like this:
+```scala
+def reduce[A : Monoid](xs: List[A]): A = ???
+```
+Since we don't have a name for the `Monoid` instance of `A`, we need to resort to `summon` in the body of `reduce`:
+```scala
+def reduce[A : Monoid](xs: List[A]): A =
+  xs.foldLeft(summon Monoid[A])(_ `combine` _)
```
johnynek, line 67: `summon[Monoid[A]].unit`

johnynek (commented):
To me, these are two separate changes that can be considered completely independently:

1. Context bound changes
2. Given syntax changes

I think it would be valuable to discuss, implement and approve these two independently.

JD557:
I think something is off in this branch. The [Named Tuples](https://github.com/scala/improvement-proposals/pull/72/commits) commits (from `fd964a9` to `d649f6e`) are included here as well.

I imagine this was intended to only start at `267b8ff`?

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,380 @@
...
+Aggregate context bounds like `A : X : Y` are not obvious to read, and it becomes worse when we add names, e.g. `A : X as x : Y as y`.
+
+**Proposal:** Allow to combine several context bounds inside `{...}`, analogous
+to import clauses. Example:
+
+```scala
+  trait:
+    def showMax[X : {Ordering, Show}](x: X, y: X): String
+  class B extends A:
+    def showMax[X : {Ordering as ordering, Show as show}](x: X, y: X): String =
```
johnynek, line 107: I think we should add a discussion (apologies if I missed it) of the fact that some using/givens don't have a single parameter to dispatch on. For such cases we would still employ the using variant. To me I'd like to see a case that we could completely kill the using variant. To do that, it would be good to have a catalog of examples that don't fit. For instance, does Builder/CanBuildFrom work with this approach? From cats, MonadError won't work with this, I don't think. In any case taking a few such examples like that would strengthen this.

odersky:
@JD557 yes, indeed. I force pushed as an independent separate branch, incorporating fixes to the two typos pointed out by 
@johnynek.

The proposal could be split further into two or three independent areas, but there are also connections between the parts:

 - Both context bounds and new givens use `as` to introduce an optional name.
 - Implementing context bounds for type members relies on deferred givens.

So one logical progression could be (1) deferred givens replacing abstract givens (2) new given syntax (3) context bound changes. But the motivation why the new given syntax is harmonious comes in part from the fact that it is in agreement with names for context bounds.

> think we should add a discussion (apologies if I missed it) of the fact that some using/givens don't have a single parameter to dispatch on. For such cases we would still employ the using variant. To me I'd like to see a case that we could completely kill the using variant. 

I don't think that's possible or desirable.  In my world view type classes are a kind of types for types. That means they can only refer to a single type. Multi-parameter type classes are really constraints passed as context. So "multi-parameter type class" is already a misnomer.  The name was invented in Haskell because Haskell does not have a general context passing mechanism, all has to be force-fitted into the typeclass paradigm. And of course there are also bits of context that don't constrain any type parameters. So it seems natural to keep using clauses for these cases, and reserve context bounds for true type classes.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,380 @@
...
+ - The syntax for multiple context bounds is not very clear or readable.
+ - The existing syntax for givens is unfortunate, which hinders learning and adoption.
+ - Abstract givens are hard to specify and implement and their syntax is easily confused
+   with simple concrete givens.
+
+These pain points are worth fixing on their own, independently of any other proposed improvements to typeclass support. What's more, the changes
+are time sensitive since they affect existing syntax that was introduced in 3.0, so it's better to make the change at a time when not that much code using the new syntax is written yet.
+
+
+ 1. Named tuples are a convenient lightweight way to return multiple results from a function. But the absence of names obscures their meaning, and makes decomposition with _1, _2 ugly and hard to read. The existing alternative is to define a class instead. This does name fields, but is more heavy-weight, both in terms of notation and generated bytecode. Named tuples give the same convenience of definition as regular tuples at far better readability.
```
rssh, line 42: Is this from this SIP?
  It's unclear how named tuples are connected with context bounds and givens.

odersky: Oops, no that was an oversight. Deleted.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+  def unit = 0
+```
+
+The underlying principles are:
+
+ - A `given` clause consists of the following elements:
+
+    - An optional _precondition_, which introduces type parameters and/or using clauses and which ends in `=>`,
+    - the implemented _type_,
+    - an optional name binding using `as`,
```
lihaoyi, line 243: Perhaps the meta-rule here is:

1. `<prefix> as <name>` with the name on the right is used where the name is usually optional (import renames, pattern-match name bindings, given names)

2. `<keyword> <name> <suffix>` style (`val`, `def`, `object`, `class`, etc.) with the name on the left is used where the name is mandatory and important

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+  type Element : Ord
+
+class SortedSet[A : Ord] extends Sorted:
+  type Element = A
+```
+The compiler expands this to the following implementation.
+```scala
+trait Sorted:
+  type Element
+  given Ord[Element] = compiletime.deferred
```
sjrd, line 180: Here is my main concern with this proposal. I think this expansion is dangerous and will create compatibility issues. Not backward incompatibility at the language, but intrinsic, synchronous incompatibility between libraries.

The core issue is that this desugaring introduces "nameless" members whose name actually matters a lot. There is no precedent for that in the language.

Consider for example
```scala
trait Babar[T]
given Babar[Int] with {}
given Babar[String] with {}

trait BaseA[A] {
  given Babar[A] = deferred
}

trait BaseB[A] {
  given Babar[A] = deferred
}

class Child extends BaseA[Int] with BaseB[String]
```
Now you need to implement two `given_Babar_A` that have nothing in common but that happen to share a generated name.

You could argue that this is already problematic with `given` definitions today, but at least they have an explicit definition. Once we automatically generate `given Ord[Element]` from `type Element : Ord`, and then `given_Ord_Element` from that, there is a multi-step, non-obvious relationship between an innocuous `type` definition and auto-generated names that clash.

Another example would be
```scala
type Foo : {package1.Bar, package2.Bar}
```
which would immediately result in clashing members. Generalize it to `type Foo : package1.Bar` in one super trait and `type Foo : package2.Bar` in another supertrait, and things becomes even more obscure.

There are so many ways that such a scheme is going to go wrong. At the level of the compilation scheme, we're approaching the amount of danger of value classes. Value classes also expose issues that arise in similar situations: where a generic in a superclass cannot be instantiated to some value classes because their bridges double-clash.

---

Since the core issue is that a generated name acquires a strong semantic meaning, I think we can fix this by disallowing anonymous things here. We could demand that deferred givens have a name, and that includes demanding them for context bounds on type members.

Even then, some issues remain: what is the visibility of the generated member? Does it match the visibility of the type member? What if the type member is `package[something]` but the implementing class is outside `something`?

Overall, this proposal introduces for the first time adding invisible members to open traits that actually matter in subtraits and subclasses. We have no precedent for that. There are a zillion issues that could arise from such a new situation, and they need to be explored much more carefully than what the current SIP text suggests.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+
+Changing something introduced just recently in Scala 3 is not fun,
+but I believe these adjustments are preferable to let bad syntax
+sit there and fester. The cost of changing should be amortized by improved developer experience over time, and better syntax would also help in migrating Scala 2 style implicits to Scala 3. But we should do it quickly before a lot more code
+starts migrating.
+
+Migration to the new syntax is straightforward, and can be supported by automatic rewrites. For a transition period we can support both the old and the new syntax. It would be a good idea to backport the new given syntax to the LTS version of Scala so that code written in this version can already use it. The current LTS would then support old and new-style givens indefinitely, whereas new Scala 3.x versions would phase out the old syntax over time.
+
+
+### 7. Abolish Abstract Givens
```
sjrd, line 260: Should this be 6?

sjrd (commented):
Overall 1, 2, 3 and 5 look good to me.

I have serious concerns about 4.

I am sympathetic to 7 (6?) but it's not clear that a library could migrate off of abstract givens without breaking its own binary API.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+ - `given T as x = deferred` is the analogous new syntax, which is more powerful since
+    it allows for automatic instantiation.
+ - `given T = deferred` is the anonymous version in the new syntax, which was not expressible before.
+
+**Benefits:**
+
+ - Simplification of the language since a feature is dropped
+ - Eliminate non-obvious and misleading syntax.
+
+The only downside is that deferred givens are restricted to be used in traits, whereas abstract givens are also allowed in abstract classes. But I would be surprised if actual code relied on that difference, and such code could in any case be easily rewritten to accommodate the restriction.
```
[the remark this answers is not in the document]
odersky: No it would break the binary API. The thing is, I am not sure there is any use case in the wild where abstract givens are used in abstract classes. People usually reach for traits anyway.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+
+The question is how these bounds are expanded. Context bounds on type parameters
+are expanded into using clauses. But for type members this does not work, since we cannot refer to a member type of a class in a parameter type of that class. What we are after is an equivalent of using parameter clauses but represented as class members.
+
+**Proposal:**
+Introduce a new way to implement a given definition in a trait like this:
+```scala
+given T = deferred
+```
+`deferred` is a new method in the `scala.compiletime` package, which can appear only as the right hand side of a given defined in a trait. Any class implementing that trait will provide an implementation of this given. If a definition is not provided explicitly, it will be synthesized by searching for a given of type `T` in the scope of the inheriting class. Specifically, the scope in which this given will be searched is the environment of that class augmented by its parameters but not containing its members (since that would lead to recursive resolutions). If an implementation _is_ provided explicitly, it counts as an override of a concrete definition and needs an `override` modifier.
```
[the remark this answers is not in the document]
odersky: > I find it awkward that we introduce behavior that is very different between traits and classes. There is no precedent for anything like this in the language. 

There _is_ precedent. Trait parameters are resolved in the next enclosing class.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+
+Sometimes unconventional syntax grows on you and becomes natural after a while. But here it was unfortunately the opposite. The longer I used given definitions in this style the more awkward they felt, in particular since the rest of the language seemed so much better put together by comparison. And I believe many others agree with me on this. Since the current syntax is unnatural and esoteric, this means it's difficult to discover and very foreign even after that. This makes it much harder to learn and apply givens than it need be.
+
+**Proposal:** Things become much simpler if we introduce the optional name instead with an `as name` clause at the end, just like we did for context bounds. We can then use a more intuitive syntax for givens like this:
+```scala
+given Ord[String]:
+  def compare(x: String, y: String) = ...
+
+given [A : Ord] => Ord[List[A]]:
```
bjornregnell, line 217: What does the arrow mean here?

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+given Monoid[Int]:
+  extension (x: Int) def combine(y: Int) = x + y
+  def unit = 0
+```
+If explicit names are desired, we add them with `as` clauses:
+```scala
+given Ord[String] as stringOrd:
+  def compare(x: String, y: String) = ...
+
+given [A : Ord] => Ord[List[A]] as listOrd:
```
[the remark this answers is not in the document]
bjornregnell: Currently the arrow can be used and then it means something different:
```scala
Welcome to Scala 3.4.1 (17.0.6, Java OpenJDK 64-Bit Server VM).

scala> given [T]:(T => String) = x => x.toString
def given_T_to_String[T]: T => String
```
The first time I saw the new arrow syntax in this thread, I thought it was a given for FunctionN, so there could perhaps be some confusion. How should we relate the new syntax to givens for functions? Have you considered alternatives?

bjornregnell:
This proposal us up for the next SIP meeting on May 24th which is less than a week and I think we should prepare the discussion here as soon as possible.

In general I think the ambition of this SIP is great. I find the first parts of the new syntax and expansion schemes for naming context bounds, multiple context bounds, context bounds for type members reasonable and with support from the compiler to rewrite old syntax and a long grace period before removal I think it is doable.

However, the controversial/"risky" part of this proposal is the change of syntax for givens, and the main concern is language stability versus language improvement, Or in other words: is the improvement to givens worth the costs of migration and potential risk of decreased willingness to go for Scala Next.

@odersky  It would be great if the below issues can be clarified in the proposal or in this comment thread, as input to the meeting:

1. If we introduce a new syntax for givens, how would the deprecation scheme relate to the LTS roadmap? Would the current 3.3-LTS given syntax still available in the next 3.x-LTS ? (Will 3.7 be then next LTS?)
2. What parts of the new syntax for given can be back-ported to 3.3-LTS so that the ones stuck on that LTS still can use the new syntax in parallell with the old syntax?
3. I'd like an overview of the different kinds of givens in a before-after-table or similar, as I think it is still difficult to get the big picture of all the kinds of given and the different syntax variants over time.

Assuming something like (or similar):

```scala
case class MyType(i: Int)

case class MyGenType[A](a: A)

trait MyOldTypeClass[T]:
  ...

trait MyNewTypeClass:
  type Self
  ...

```

I think it would be great if you include a summary overview for the proposed new given syntax, perhaps like this (or similar) in the proposal:

| kind of given | current  | proposal |
|--|--|--|  
| named given value | given x: MyType = MyType(42) |  given MyType as x = MyType(42)|
| anonymous given value | given MyType = MyType(42) | no change |
...etc. 

As there are many types of givens, some available [here](https://docs.scala-lang.org/scala3/reference/contextual/index.html), I think it is valuable if there is a such a systematic walk-through of all combinations so we don't miss any feature interaction and also can get an overview of how big the syntax changes are. In the current proposal only a few of all possible kinds of givens are shown in the new syntax and it is not clear to me if more variants are impacted.

I think the new arrow syntax using `=>` is the most controversial/radical change so its important to include explanations of interpretations/expansions for that in the actual proposal (some rationale is currently available in the review comment only). I find this rather difficult to read: `given [A : Ord] => Ord[List[A]]: ...` and I wonder if a keyword instead of `=>` such as `for` or similar could make it easier to grok? Have you considered alpha keyword alternatives instead of a symbol?

bjornregnell:
Another question I have:
> The only downside is that deferred givens are restricted to be used in traits, whereas abstract givens are also allowed in abstract classes.

Can deferred givens be allowed also in abstract classes? If no, why not?

bjornregnell:
> It would complicate things a bit

Well, I'm inclined to favor avoidance of unnecessary exceptions so that rules are sinmpler. But if the implementation is tricky, the relaxation of the restriction on deferred givens can be added later.

odersky:
I added a new section that does a systematic comparison between current and proposed syntax.

content/typeclasses-syntax.md
```diff
@@ -356,6 +356,256 @@ I believe that a default naming convention for witnesses will be very beneficial
...
+
+**Proposal, Anonymous**
+
+```scala
+  // Simple typeclass
+  given Ord[Int]:
+    def compare(x: Int, y: Int) = ...
+
+  // Parameterized typeclass
+  given [A: Ord] => Ord[List[A]]:
```
johnynek, line 449: I wonder if we could "regularize" this syntax... `[A: Ord] => Ord[List[A]]` looks like a polymorphic function with a context... something like: `[A] => Ord[A] ?=> Ord[List[A]]` So, if we allowed polymorphic context functions with this syntax, `[A: Ord] => Ord[List[A]]` could be a valid type interpreted the same as `[A] => Ord[A] ?=> Ord[List[A]]`.

So, using something close to current syntax:
```scala
given [A: Ord] => Ord[List[A]]:
  [A] => (ord: Ord[A]) ?=> Ord[List[A]]:
    def compare(x: List[A], y: List[A]) = ...
```
now, the given line and the following line is duplicated. If we had a rule that a context function can be written as:
```scala
trait SomeTrait:
  def next(i: Int): Int

val fn: Ctx ?=> SomeTrait:
  //.. we can directly implement SomeTrait without, new, etc...
  next(i: Int) = i + skipSize(using Ctx)
```

Last, we could imagine that if there is a given polymorphic context function, in scope, then it can be used.

If we had the above, this given syntax would fall out of it directly and make the language more regular.

I think the least possible exceptions to syntax helps people learn and read code.

bjornregnell:
> I added a new section that does a systematic comparison between current and proposed syntax.

Many thanks! I think the added systematic comparison is very nice to have.

I found typos as reported below (or I'm confused...):

1. Under "Current, Anonymous" `// Alias with using clause` there should be a constructor application, i.e. missing `()` should be added as in:
```scala
 // Alias with using clause
  given [A](using Ord[A]): Ord[List[A]] =
    ListOrd[A]()
``` 
2. Same missing () under "Proposal, Anonymous", should be:
```scala
  // Alias with using clause
  given [A](using Ord[A]) => Ord[List[A]] =
    ListOrd[A]()
```
3. Same missing () under "Current, Named", should be:
```scala
  // Alias with using clause
  given listOrd[A](using Ord[A]): Ord[List[A]] =
    ListOrd[A]()
```
4. Same missing () under "Proposal, Named", should be:
```scala
  // Alias with using clause
  given [A](using Ord[A]) => Ord[List[A]] as listOrd =
    ListOrd[A]()
```
5. There is a name missing (listOrd) in the "Current, Named", under `// Typeclass with using clause` should be:
```scala
  // Typeclass with using clause
  given listOrd[A](using Ord[A]): Ord[List[A]] with
    def compare(x: List[A], y: List[A]) = ...
```
6. There is one colon too much here under "Current, Named":
```scala
  // Parameterized typeclass
  given listOrd: [A: Ord]: Ord[List[A]] with
    def compare(x: List[A], y: List[A]) = ...  
```
should be
```scala
  // Parameterized typeclass
  given listOrd[A: Ord]: Ord[List[A]] with
    def compare(x: List[A], y: List[A]) = ...
```

bjornregnell:
@odersky An idea I got after your reasoning about that the important think should come first. In the propsed arrow syntax case the important thing comes last; the most important thing is the actual given thing, not the type param precondition. 

If we have the context bound inside the type param, the actual given will be the important thing and we can do away with the `=>` symbol sallad, like so:
```scala
// Parameterized typeclass
  given Ord[List[A : Ord]]:
    def compare(x: List[A], y: List[A]) = ...
```
What do you think about that?

content/typeclasses-syntax.md
lenguyenthanh: Am I understand correctly, your proposal is rewritten this:

```scala
  given [A: Ord] => Ord[List[A]]:
    def compare(x: List[A], y: List[A]) = ...
```
to this:

```scala
  given Ord[A] ?=> Ord[List[A]]:
    def compare(x: List[A], y: List[A]) = ...
```

If that's correct, I really like the idea, because it's just context functions again. `=>` here is a bit awkward imho.

EugeneFlesselle:
@odersky as mentioned during yesterday's lab meeting, here is an extended example comparing the old and proposed syntax which could be added to the proposal.

<table>
<tr>
<th> New syntax </th>
<th> Old syntax </th>
</tr>
<tr>
<td>

```scala 3
import scala.language.experimental.{modularity, clauseInterleaving}

trait TC:
  type Self

/** A value with a witness of its type conforming to some type class. */
sealed trait Container[Interface <: TC]:
  /** The type of the contained value. */
  type Value : Interface as witness
  /** The contained value. */
  val value: Value



object Container:

  /** Wrapes a value of type `V` into a `Container[I]` 
   *  provided a witness that `V is I`. */
  def apply[I <: TC](v: Any)[V >: v.type : I] = new Container[I]:
    type Value >: V <: V
    val value: Value = v
  

  /** An implicit constructor for `Container[I]` from `V`. */
  given constructor[I <: TC, V : I]: Conversion[V, Container[I]] = apply

end Container


trait Showable extends TC:
  extension (self: Self) def show: String

object Showable:

  given String is Showable:
    extension (self: String) def show: String = self

  given Int is Showable:
    extension (self: Int) def show: String = s"My int is $self"

  given Double is Showable:
    extension (self: Double) def show: String = s"My double is $self"

  given EmptyTuple is Showable:
    extension (self: EmptyTuple) def show: String = "EmptyTuple"
  given [H : Showable, T <: Tuple : Showable] => H *: T is Showable:
    extension (self: H *: T) def show: String = 
      s"${self.head.show} andThen ${self.tail.show}"

end Showable

def showAll(xs: Container[Showable]*): Seq[String] =
  xs.map(_.value.show)


@main def Test =

  // Constructing
  Container[Showable]("Hello")[String]
  Container[Showable]("Hello")
  val x: Container[Showable] = Container("Hello")

  // Destructing
  assert(x.witness.show(x.value) == "Hello")
  assert(x.value.show == "Hello")

  showAll(Container("Hello again"), Container(23), Container("bye"))


trait Serializable extends TC:
  type Output
  extension (self: Self) def serialized: Output

object Serializable:
  type To[O] = Serializable { type Output = O }

  given Int is Serializable as IntIsSerializableToByte:
    type Output = Byte
    extension (self: Int) def serialized: Byte = self.toByte

  given Int is Serializable as IntIsSerializableToString:
    type Output = String
    extension (self: Int) def serialized: String = self.toString

end Serializable


trait Monoid extends TC:
  def unit: Self
  extension (x: Self) def + (y: Self): Self

object Monoid:

  given String is Monoid:
    def unit: String = ""
    extension (x: String) def + (y: String): String = s"$x ~ $y"

  given Tuple is Monoid:
    def unit: Tuple = EmptyTuple
    extension (x: Tuple) def + (y: Tuple): Tuple = x ++ y

end Monoid


def serializeCombine[O : Monoid as m](xs: List[Container[Serializable.To[O]]]): O =
  val ys: List[O] = xs.map(_.value.serialized)
  ys.fold(m.unit)(_ + _)



sealed trait Container2[Interface1 <: TC, Interface2 <: TC]:
  /** The type of the contained value. */
  type Value : { Interface1 as witness1, Interface2 as witness2 }
  /** The contained value. */
  val value: Value




object Container2:
  def apply[I1 <: TC, I2 <: TC](v: Any)[V >: v.type : {I1, I2}] = 
    new Container2[I1, I2]:
    type Value >: V <: V
    val value: Value = v

  
  
  given constructor[I1 <: TC, I2 <: TC, V : {I1, I2}]: Conversion[V, Container2[I1, I2]] = apply
end Container2

def showAllPos(xs: Container2[Showable, Numeric]*): Seq[String] =
  xs.filter(_.value.toInt > 0).map(_.value.show)
```

</td>
<td>

```scala 3
import scala.language.experimental.clauseInterleaving




/** A value with a witness of its type conforming to some type class. */
sealed trait Container[Interface[_]]:
  /** The type of the contained value. */
  type Value
  /** The contained value. */
  val value: Value

  given witness: Interface[Value]

object Container:

  /** Wrapes a value of type `V` into a `Container[I]` 
   *  provided a witness that `V is I`. */
  def apply[I[_]](v: Any)[V >: v.type](using w: I[V]) = new Container[I]:
    type Value >: V <: V
    val value: Value = v
    given witness: I[V] = w

  /** An implicit constructor for `Container[C, V]` from `V`. */
  given constructor[I[_], V : I]: Conversion[V, Container[I]] = apply

end Container


trait Showable[T]:
  extension (self: T) def show: String

object Showable:

  given Showable[String] with
    extension (self: String) def show: String = self

  given Showable[Int] with
    extension (self: Int) def show: String = s"My int is $self"

  given Showable[Double]:
    extension (self: Double) def show: String = s"My double is $self"

  given Showable[EmptyTuple] with
    extension (self: EmptyTuple) def show: String = "EmptyTuple"
  given [H : Showable, T <: Tuple : Showable]: Showable[H *: T] with
    extension (self: H *: T) def show: String = 
      s"${self.head.show} andThen ${self.tail.show}"

end Showable

def showAll(xs: Container[Showable]*): Seq[String] =
  xs.map(_.value.show)


@main def Test =

  // Constructing
  Container[Showable]("Hello")[String]
  Container[Showable]("Hello")
  val x: Container[Showable] = Container("Hello")

  // Destructing
  assert(x.witness.show(x.value) == "Hello")
  assert(x.value.show == "Hello")

  showAll(Container("Hello again"), Container(23), Container("bye"))


trait Serializable[T]:
  type Output
  extension (self: T) def serialized: Output

object Serializable:
  type To[O] = [T] =>> Serializable[T] { type Output = O }

  given IntIsSerializableToByte: Serializable[Int] with
    type Output = Byte
    extension (self: Int) def serialized: Byte = self.toByte

  given IntIsSerializableToString: Serializable[Int] with
    type Output = String
    extension (self: Int) def serialized: String = self.toString

end Serializable


trait Monoid[T]:
  def unit: T
  extension (x: T) def + (y: T): T

object Monoid:

  given Monoid[String] with
    def unit: String = ""
    extension (x: String) def + (y: String): String = s"$x ~ $y"

  given Monoid[Tuple] with
    def unit: Tuple = EmptyTuple
    extension (x: Tuple) def + (y: Tuple): Tuple = x ++ y

end Monoid


def serializeCombine[O](xs: List[Container[Serializable.To[O]]])(using m: Monoid[O]): O =
  val ys: List[O] = xs.map(_.value.serialized)
  ys.fold(m.unit)(_ + _)



sealed trait Container2[Interface1[_], Interface2[_]]:
  /** The type of the contained value. */
  type Value
  /** The contained value. */
  val value: Value

  given witness1: Interface1[Value]
  given witness2: Interface2[Value]

object Container2:
  def apply[I1[_], I2[_]](v: Any)[V >: v.type](using w1: I1[V], w2: I2[V]) = new Container2[I1, I2]:
    type Value >: V <: V
    val value: Value = v
    given witness1: I1[V] = w1
    given witness2: I2[V] = w2

  given constructor[I1[_], I2[_], V : I1 : I2]: Conversion[V, Container2[I1, I2]] = apply
end Container2

def showAllPos(xs: Container2[Showable, Numeric]*): Seq[String] =
  xs.filter(_.value.toInt > 0).map(_.value.show)
```

</td>
</tr>
</table>

odersky:
@bjornregnell Unfortunately that would be ambiguous. If there are no context bounds then
```scala
given List[Ord[A]]
```
could mean: for all `A`, `List[Ord[A]]` or it could mean: `List[Ord[A]]` for a specific class named `A`.

odersky:
@EugeneFlesselle This is a nice example, that mostly shows savings from having context bounds as members, which translate into deferred givens, instead of having to code up everything yourself using abstract givens. 

It does use more syntax than in this SIP. In particular, infix `is` and Self` types are not proposed in this SIP. But I think it should be easy to adapt the example to use parameterized type classes.

odersky:
> Shouldn't context bounds work also with function lambda syntax?

Yes, this would be another useful generalization.

sjrd:
Here is one concrete example of my concerns with the type bounds on abstract type members:
```scala
package strawman {
  package coll {
    trait Sequence[A]

    package immut {
      trait Sequence[A]
    }
  }
}

package package1 {
  trait Foo:
    type MySeq : strawman.coll.Sequence
}

package package2 {
  trait Bar:
    type MySeq : strawman.coll.immut.Sequence
}

package app {
  class ConcreteSeq
  given strawman.coll.Sequence[ConcreteSeq] as IsColl with {}
  given strawman.coll.immut.Sequence[ConcreteSeq] as IsImmutColl with {}

  class Foobar extends package1.Foo with package2.Bar:
    type MySeq = ConcreteSeq
}
```
gives
```
-- [E164] Declaration Error: tests/run/hello.scala:44:2 ------------------------
44 |  class Foobar extends package1.Foo with package2.Bar:
   |  ^
   |error overriding given instance given_Sequence_MySeq in trait Foo of type strawman.coll.Sequence[Foobar.this.MySeq];
   |  given instance given_Sequence_MySeq of type strawman.coll.immut.Sequence[Foobar.this.MySeq] has incompatible type
   |
1 error found
```
The *double* indirection to give a name that is actually semantically meaningful from anonymous things truly makes compatibility scenarios impossible to predict for anyone but compiler writers.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,373 @@
...
+
+The using clause in class `SortedSet` provides an implementation for the deferred given in trait `Sorted`.
+
+**Benefits:**
+
+ - Better orthogonality, type parameters and abstract type members now accept the same kinds of bounds.
+ - Better ergonomics, since deferred givens get naturally implemented in inheriting classes, no need for boilerplate to fill in definitions of abstract givens.
+
+**Alternative:** It was suggested that we use a modifier for a deferred given instead of a `= deferred`. Something like `deferred given C[T]`. But a modifier does not suggest the concept that a deferred given will be implemented automatically in subclasses unless an explicit definition is written. In a sense, we can see `= deferred` as the invocation of a magic macro that is provided by the compiler. So from a user's point of view a given with `deferred` right hand side is not abstract.
+It is a concrete definition where the compiler will provide the correct implementation.
```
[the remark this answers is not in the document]
sjrd: I am still completely baffled by this line of thought. Magic things are fine when they do things that normal things cannot do, but still *where* normal things can do them. Example of good magic things:
* Primitives
* `uninitialized`
* `compiletime.ops`

The proposed `deferred` has non-local consequences that go far beyond what we can explain as a macro, even magic. On the contrary, it *completely* matches what modifiers have the power to do: `final` has the power to non-locally prevent overriding, for example. `deferred` as a modifier would be completely in line with other modifiers.

content/typeclasses-syntax.md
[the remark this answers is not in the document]
sjrd: Then, as I mentioned before offline (but I can't find a written trace online), it is simply not acceptable to ever get rid of abstract givens. If there is a library that uses one today, it will never be able to upgrade its compiler version, no matter what source rewritings it attempts to do.

This is a big no-no. Removing them would not only break backward source compat (which we *can* do), it would break our promises of binary compatibility due to second-order effects.

We can *deprecate* them, with the usual options to silence the deprecation, but we can never remove them.

content/typeclasses-syntax.md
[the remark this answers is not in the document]
sjrd: OK. I still think it is weird but it is acceptable.

sjrd:
To summarize my thoughts in one line per proposed item ahead of the meeting:

1. Naming context bounds: good ✅
2. New Syntax for Aggregate Context Bounds: good ✅
3. Expansion of Context Bounds: good ✅
4. Context Bounds for Type Members, Deferred Givens
  a. Context Bounds for type Members: strong concerns for compatibility scenarios ❌
  b. Deferred Givens: syntax concerns, should be a modifier, but otherwise OK ⚠️
5. Cleanup of Given Syntax: I don't want to fight it ✅
6. Abolish Abstract Givens: can never remove under any circumstance ❌ (deprecation that can be silenced is fine)

odersky:
> Here is one concrete example of my concerns with the type bounds on abstract type members:

I notice that if you change it to
```scala
package strawman {
  package coll {
    trait Sequence[A]

    package immut {
      trait Sequence[A] extends coll.Sequence[A]
    }
  }
}
```
it compiles. So, to summarize the conditions that make it fail:

 - We need two unrelated typeclass traits with the same name but in different packages (that's _already_ a big code smell). 
 - We need two traits that each have a context bound one of these typeclass traits on the same abstract type.
 - We need a class that mixes in these two traits.

In that case we are indeed stuck. We simply can't mix the two traits in one class. But there are much simpler and more common scenarios where the same happens:

```scala
trait Foo:
  def foo(): Int

trait Bar:
  def foo(): String

class C extends Foo, Bar // unrecoverable error
```
So, yes, since Scala does not have a general member renaming mechanism (since th JVM does not support that), there are traits that can't be mixed in together. But I believe the case with deferred gives is so articifial that it does not materially change that situation.

odersky:
For removing abstract givens: We can do it slowly and only after a survey of existing libraries. Let's say no project in open CB uses abstract givens. Can't we remove it (it would stay around under a source flag for sure). 

If we are worried about Tasty compatibility: The Tasty format would still support them in the whole 3.x series. But can't we drop them from source at some point?

bjornregnell:
> > Shouldn't context bounds work also with function lambda syntax?
> 
> Yes, this would be another useful generalization.

I think context bounds in polymorfic function types like `val f: [A: Ordinal] => A => ...` should be included in this SIP as it seems like an uncontroversial and intuitive generalisation. The actual implementation can come in due course; by including it in this SIP, we have a more coherent, regular syntax update.

bjornregnell:
Here is my summary comment as input to the meeting:
* There are typos in the nice comparison summary as reported here that I think should be fixed: https://github.com/scala/improvement-proposals/pull/81#issuecomment-2124769512

* I'm in favor of the proposed new *naming context bound* using the `as` keyword. It's in line with import renaming.
* I'm in favor of proposed new syntax for *aggregate context bounds* using braces as it's easier to read.
* I'm in favor of generalizing abstract *type members to allow context bounds* as it improves regularity and it is a surprise if it's not possible to do that.
* I'm not sure I understand all the intricacies of *abstract givens and deferred givens*, but it seems more intuitive to me that it is a modifier, and I don't really get how "a story of a magic method" is better than a modifier; a newcomer just needs to learn the syntax anyway and if it is not for some reason valid with the existing keyword `abstract` as in a `abstract given ...` modifier then I think a `deferred` modifier makes more sense than a magic method, with a good doc explanation of why `deferred given ...` is needed and useful. In the proposal example there is even an `override given ...``for the deferred one, so I'd argue that a modifier is more regular as it is the same syntax position. But all this seems like very intricate details and I can accept any of a modifier or a magic method if others think there are prevailing arguments for any of them. For me its ok to keep experimenting with this if we accept the SIP.
* I'm mostly OK with *cleaning up the given syntax* and I like `as` better than double-colon, esp. for anonymous givens. The price for the change in badwill is probably worth it, but there needs to be a really good communication effort with the community to explain in clear terms why it is needed and better on a balance. I am hesitant towards the new `=>` syntax but after contemplating alternatives I could not find anything better, and there is the precedent in the analogy with `=>` in pattern matching. But I have an itch (or hunch) that the interaction with function types and polymorfic function types and context function types need more investigation, see also comment by [@johnynek above](https://github.com/scala/improvement-proposals/pull/81#discussion_r1608678684) but that can come during the experimental phase if we accept this SIP. 
* I think allowing context bounds in polymorfic function types and its de-sugaring into context function types should be included in the proposal to make it more complete.

sjrd:
> For removing abstract givens: We can do it slowly and only after a survey of existing libraries. Let's say no project in open CB uses abstract givens. Can't we remove the feature then? (it would stay around under a source flag for sure).
> 
> If we are worried about Tasty compatibility: The Tasty format would still support them in the whole 3.x series. But we should be able to drop them from the source language at some point.

That is first-order TASTy compatibility. The second order I'm talking about is that we have to allow, *somehow*, libraries to upgrade to newer versions of the compiler without breaking their own compatibility (other than requiring users to also upgrade their compiler). If it requires source changes, fine; if it requires changing compiler flags, fine; but there needs to be a way.

>  (it would stay around under a source flag for sure).

This would qualify as acceptable, indeed.

content/typeclasses-syntax.md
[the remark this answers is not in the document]
sjrd: Taking my reasoning further, there are other practical consequences here.

### Scaladoc

As a user of a trait containing a deferred given, when I extend it, it is completely different for me if that it is a deferred given or a concrete given with any other body. That means Scaladoc must show the *deferred* aspect of it. This is not something Scaladoc does for things after `=`: the part *after* the `=` is an implementation detail, not an API. (Unless the thing is an `inline` itself, but *being* a macro is completely different than *calling* a macro.)

### Implementation under separate compilation

Likewise, and for the same reason, the compiler also needs to look at the deferred property of things coming from separate compilation. So for practical reasons it also makes no sense for it to be on the implementation side of the `=`. It is clearly part of the ABI contract (compare ABI here with API in my Scaladoc point).

lrytz:
IIRC, there were some arguments on the pre-sip discussion about limits with the new `T: C as c` and explicit `using` paramters still being required to express certain cases? Or maybe it was about the improved syntax for given definitions? Maybe someone remembers or has a pointer.

odersky:
@lrytz I am not sure. We need using parameters for things that are not typeclasses. Or if one wants to insist on a using clause at a different place than where it would be inserted automatically. Was there something else?

bjornregnell:
@sjrd As a manager of this SIP, perhaps you can summarize here its current status and result from our last SIP-meeting. I'm not sure the status label here on github is correct in relation to previous votes. I also think it would be good if you ping the [Contributors-thread](https://contributors.scala-lang.org/t/pre-sip-improve-syntax-for-context-bounds-and-givens/6576/97) and explain the current status to the community in order to invite experimentation and feedback. 

I think the latest edits by @odersky is a great improvement in explaining the rationale for the new syntax, making it easier to understand how to apply the prameterization `[A : B] => ...` variant if you know the rational.

bjornregnell:
@odersky Perhaps the proposal should be clear about the corner case of givens for function types, even if they are rare or even unrecommended, also the variant of a given for a polymorfic function type. I guess it boils down to having parenthesis in the right places, and I think it would be good to state those rules in the proposal.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,657 @@
...
+
+The proposed set of changes removes awkward syntax and makes dealing with context bounds and givens a lot more regular and pleasant. In summary, the proposed changes are:
+
+ 1. Allow to name context bounds with `as` clauses.
+ 3. Introduce a less cryptic syntax for multiple context bounds.
+ 4. Allow context bounds on type members which expand to deferred givens.
+ 5. Introduce a more regular and clearer syntax for givens.
+ 6. Eliminate abstract givens.
+
+These changes were implemented as part of a  [draft PR](https://github.com/lampepfl/dotty/pulls/odersky)
```
soronpo, line 653: The link to the PR is missing.

sjrd:
> @sjrd As a manager of this SIP, perhaps you can summarize here its current status and result from our last SIP-meeting. I'm not sure the status label here on github is correct in relation to previous votes.

It's not easy to write a summary that encompasses the various concerns that were raised. Here is a summary of the summary, though:

* There seems to be a majority of members who think the changes are good in principle.
* There seems to be agreement on the first few changes.
* The syntax of the last few things is highly controversial: we have differing opinions on how to interpret the *meaning* of things, and that has an impact on whether we perceive the *syntax* to reflect that meaning well or not.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,657 @@
...
+    ListOrd[A]()
+
+  // Concrete class instance
+  given context: Context // this would be a change of meaning from abstract given
+
+  // Abstract or deferred given
+  given context: Context = deferred
+
+  // By-name given
+  given context: => Context = curCtx
```
sideeffffect, line 634: I'm not sure if this is the right place to post this, or whether it's perhaps too late. But I was wondering, whether we could have an alternative proposal which looks like this.

The benefit I see it this is that it keeps the syntax very regular and intentionally resembles definition of methods (but with `given` instead of `def`).
```suggestion
  // Simple typeclass
  given intOrd: Ord[Int]:
    def compare(x: Int, y: Int) = ...

  // Parameterized typeclass
  given listOrd[A: Ord]: Ord[List[A]]:
    def compare(x: List[A], y: List[A]) = ...

  // Typeclass with using clause
  given listOrd[A](using Ord[A]): Ord[List[A]]:
    def compare(x: List[A], y: List[A]) = ...

  // Simple alias
  given intOrd: Ord[Int] = IntOrd()

  // Parameterized alias
  given listOrd[A: Ord]: Ord[List[A]] =
    ListOrd[A]()

  // Alias with using clause
  given listOrd[A](using Ord[A]): Ord[List[A]] =
    ListOrd[A]()

  // Concrete class instance
  given context: Context // this would be a change of meaning from abstract given

  // Abstract or deferred given
  given context: Context = deferred

  // By-name given
  given context: => Context = curCtx
```

bjornregnell:
To xref from contributors: https://contributors.scala-lang.org/t/pre-sip-improve-syntax-for-context-bounds-and-givens/6576/117?u=bjornregnell

The new suggestion there is to require explicit () on concrete class instance givens as in:
```scala
// Concrete class instance  (named, colon-style)
  given context: Context()

// Concrete class instance  (named, as-style)
  given Context() as context

// Concrete class instance  (unnamed)
  given Context()
```

bjornregnell:
Another xref from contributors: https://contributors.scala-lang.org/t/pre-sip-improve-syntax-for-context-bounds-and-givens/6576/124?u=bjornregnell

The new suggestion is to automatically expand context bounds on higher-kinded types.

bishabosha:
> The new suggestion by @bishabosha is to automatically expand context bounds on higher-kinded types.

Slight correction - to infer type bounds from the `Self` type of the context bound

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,657 @@
...
+### 6. Cleanup of Given Syntax
+
+A good language syntax is like a Bach fugue: A small set of motifs is combined in a multitude of harmonic ways. Dissonances and irregularities should be avoided.
+
+When designing Scala 3, I believe that, by and large, we achieved that goal, except in one area, which is the syntax of givens. There _are_ some glaring dissonances, as seen in this code for defining an ordering on lists:
+```scala
+given [A](using Ord[A]): Ord[List[A]] with
+  def compare(x: List[A], y: List[A]) = ...
+```
+The `:` feels utterly foreign in this position. It's definitely not a type ascription, so what is its role? Just as bad is the trailing `with`. Everywhere else we use braces or trailing `:` to start a scope of nested definitions, so the need of `with` sticks out like a sore thumb.
```
Ichoran, line 234: The `:` feels utterly familiar in this position to me.

```scala
def foo()[A](using Ord[A]): Ord[List[A]]
```

What is foreign about that?  We write this ubiquitously!  It's simply the return type.

What is weird is that we're both declaring the return type and defining an anonymous class that implements it and asking for an instance to be created _all with the same `: Type with` construct.  That _is_ a lot of lifting for not very much syntax.

The non-given way we'd do that is by leveraging type inference

```scala
def foo[A](using Ord[A]) = new Ord[List[A]]:
  ...
```

bishabosha:
```scala
given [A](using Ord[A]): Ord[List[A]] with
```
the problem is that it isn't a return type, its an extends clause in disguise (here is the desugar):
```scala
implicit def Ord_List[A](using Ord[A]): Ord_List[A] = new Ord_List[A]

class Ord_List[A](using Ord[A]) extends Ord[List[A]]:
  def compare(x: List[A], y: List[A]) =
```

Ichoran: Okay, but do we actually _want_ this to be a `given_Ord_List[A]` as opposed to an `Ord[List[A]]`?  If I were aliasing, I wouldn't have declared a new given-specific type.  This is pretty weird, because I can

```scala
var x = summon[Ord[List[Int]]]
```

and depending on whether I used an alias or created it de-novo, I get a different return type.

That's kind of...baffling.  You can _still_ explain the `given ... with` as doing a `transparent` as well, if you want to say that `: Ord[List[A]]` is doing its normal job (I'd still say it is).  So that _still_ doesn't seem to be the weird part to me.

Anyway, I don't understand why one would want to expose in this fashion what looks like implementation details.

bishabosha: because the with can introduce new refinements

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,819 @@
...
+ - Overall, I find this version a bit more cumbersome to the one with `as`, but i could live with it (since I don't usually recommend to write named givens anyway).
+
+### Alternative: Reinforce Similarity with Function Types
+
+A reservation against the new syntax that is sometimes brought up is that the `=>` feels strange. I personally find the `=>` quite natural since it means implication, which is exactly what we want to express when we write a conditional given. This also corresponds to the meaning of arrow in functions since by the Curry-Howard isomorphism function types correspond to implications in logic.
+Besides `=>` is also used in other languages that support type classes (e.g.: Haskell).
+
+As an example, the most natural reading of
+```scala
+given [A: Ord] => List[Ord[A]]
```
bishabosha, line 651:
```suggestion
given [A: Ord] => Ord[List[A]]
```

content/typeclasses-syntax.md
Ichoran: The question still stands as to whether you want to expose this in a given.  Because you can't easily name the type, you can't declare that you intend to get those refinements.  You only get them accidentally when you ask for the named type from a context where the refinement has visibility.

Depending on the details of behavior of an un-nameable type seems like an antipattern.

Note that this is not the case for vals in Scala 3.4.2:

```scala
scala> trait T { def foo: Int }
// defined trait T
                                                                                
scala> val x = new T:
     |   val foo = 5
     |   def bar = true
     | 
val x: T = anon$1@52021702
                                                                                
scala> x.bar
-- [E008] Not Found Error: -----------------------------------------------------
1 |x.bar
  |^^^^^
  |value bar is not a member of T
1 error found
```

Not true for defs:

```scala
     |   val foo = i
     |   def bar = i < 0
     | 
def f(i: Int): T
                                                                                
scala> f(3).bar
-- [E008] Not Found Error: -----------------------------------------------------
1 |f(3).bar
  |^^^^^^^^
  |value bar is not a member of T
1 error found
```

So that the refinement is visible is an inconsistency that perhaps should be removed.  Do we crucially depend upon it for anything?

bishabosha: actually this is identical to implicit object in scala 2 - and the name is visible in scope even for anonymous, or just don't make it anonymous?

Ichoran: That it parallels Scala 2 is less important, I think, than being consistent with Scala 3.  (Note that there was a behavior change for defs and vals; you used to get a return type that was the supertype with a structural type over the refinements.)

odersky:
The discussion on contributors gave a clear preference for 

 - expressing optional names using prefix `x:` instead of suffix `as x`. 
 - using the alternative syntax that makes givens more like functions. 

I propose to change the SIP to reflect these preferences as the standard proposal and to change the implementation accordingly. Are there any objections to this?

Ichoran:
I am happy with it though I suspect it is not the better solution for most people due to lesser familiarity with curried function syntax than with method syntax.

Because the `: ... with` problem still exists, except `: ... with` got renamed `=> ... :` [(explanation)](https://contributors.scala-lang.org/t/pre-sip-improve-syntax-for-context-bounds-and-givens/6576/144?u=ichoran), I would also change the discussion of `with`.  You still have something that looks like it is in type position but isn't a type except a new type is synthesized for it so it is a type after all albeit one you can't name yourself but can see when it's inferred (see discussion in text).

In

```scala
given myOrdList: [A] => Ord[A] => Ord[List[A]]:
  def compare...
```

what is that first `:` after `myOrdList`?  The thing on the right is, as the discussion says, not a type ascription.

If we don't want to do anything about this, the discussion with `with` can simply say that since we open anonymous class blocks via `:` not `with` (i.e. `new X:` not `new X with`), we want to parallel that here.

bishabosha:
do we need the double colon?
```scala
given myOrdList[A] => Ord[A] => Ord[List[A]]:
  def compare...

given [A] => Ord[A] => Ord[List[A]]:
  def compare...

given byname() => Ord[Int] = ...

given () => Ord[Int] = ... // anon byname
```

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,819 @@
...
+type Comparer = [X: Ord] => (x: X, y: X) => Boolean
+val less: Comparer = [X: Ord as ord] => (x: X, y: X) =>
+  ord.compare(x, y) < 0
+```
+
+The expansion of such context bounds is analogous to the expansion in method types, except that instead of adding a using clause in a method, we insert a context function type.
+
+For instance, type and val above would expand to
+```scala
+type Comparer = [X] => (x: X, y: X) => Ord[X] ?=> Boolean
```
johnynek, line 211: is this right? to me, `[X: Ord] => (X, X) => Boolean` should expand to `[X] => Ord[X] ?=> (X, X) => Boolean`. Why is the context function getting interwoven in the middle of `(X, X) => Boolean`?

johnynek:
The proposal still has examples like:
```scala
given [A: Ord] => Ord[List[A]]:
  def compare(x: List[A], y: List[A]) = ???
```
to me, a very important concept in programming is the substitution principle. When I look at this, I want to be able to leverage substitution:

```scala
// this type should be well defined as [A] => Ord[A] ?=> Ord[List[A]]
type OrdList = [A: Ord] => Ord[List[A]]

// and we should be able to use it in a given:
given OrdList:
  def compare(x: List[A], y: List[A]) = ...

// or with a value:

val myOrder: [A] => Ord[A] ?=> Ord[List[A]] = ...

given OrdList:
  myOrder
```

Are those substitutions valid?

odersky:
> How would the syntax look like if one wants to name (one of the) dependencies?

It's (A)

> [A] => (aOrd: Ord[A]) => Ord[List[A]]:

And yes, that's a legal function type in Scala. It's already used the syntax for a dependent function type such as

```scala
(x: C) => x.Elem
```

odersky:
@johnynek 

> Are those substitutions valid?

They are not. The arrow really means implication. The analogy with function types serves as a help to remember the syntax, not to be taken literally. I argue this is not a problem since nobody should want to define givens for functions, that's generally a very dubious thing to do. If you really want to implement a give instance of a function type, you can use the type alias as you have shown, or put the type in parentheses.

johnynek:
In my opinion, defining a given context function seems pretty natural. In fact, that is what we are doing with `given [A: Ord] => Ord[List[A]]` in my opinion. So, I would rather simplify the number of contexts and say that is exactly what we are doing in this case.

odersky:
> In my opinion, defining a given context function seems pretty natural. In fact, that is what we are doing with given [A: Ord] => Ord[List[A]] in my opinion. 

Yes, that's certainly the inspiration. But the fine print does not work out. Given parameters can carry modifiers or annotations but function parameters can't. That's why I was careful to write "it should _look like_ a function type", not "it should _be_ a function type". Still, I think the analogy with function types is useful for learning and intuition. Most people won't even notice the details where they differ.

odersky:
@Ichoran, all: I did a major update of the proposal which now puts the revised syntax first.

bjornregnell:
@odersky I think the summary needs updating as the `as` syntax is still there etc.

bjornregnell:
> Given parameters can carry modifiers or annotations but function parameters can't.

Hmmm. 
@odersky These implicit function values *do* work with both a modifier (final, inline) and an annotation (at unchecked):
```scala
Welcome to Scala 3.4.2 (21.0.3, Java OpenJDK 64-Bit Server VM).
                                                                                    
scala> final given f: (Int => Int) = x => x + 1
lazy val f: Int => Int

scala> @unchecked given f: (Int => Int) = x => x + 1
lazy val f: Int => Int
                                                                                    
scala> inline def z(using inline f: Int => Int) = f(0)
def z(using f: Int => Int): Int

scala> def z(using @unchecked f: Int => Int) = f(0)
def z(using f: Int => Int): Int
                                   
```

odersky:
@bjornregnell I was referring to parameters not the given definitions themselves. For instance this is OK:

```scala
given (@unchecked x: T) => U = ...
```
But this is not:
```scala
val v: (@unchecked x: T) => U = ...
```
Here, `(x: T) => U` is a dependent function type, and these don't permit modifiers or annotations on their parameters.

bjornregnell:
Aha. Thanks for clarification.

So then I guess it would be an opportunity, when given function values are wanted, to first make a normal def (with all the possible glory of modifiers and annotations etc) and then use it as an eta-expanded value in the body of an alias given like so:

```scala
Welcome to Scala 3.4.2 (17.0.8.1, Java OpenJDK 64-Bit Server VM).
                                                                                
scala> inline transparent def f(@unchecked inline x: Int) = x
def f(x: Int): Int
                                                                                
scala> given (Int => Int) = f
lazy val given_Int_to_Int: Int => Int
                                                                                
scala> def z(using f: Int => Int) = f(0)
def z(using f: Int => Int): Int
                                                                                
scala> z
val res0: Int = 0

```

odersky:
@bjornregnell Yes, that's what I would recommend in this case as well.

content/typeclasses-syntax.md
```diff
@@ -0,0 +1,685 @@
...
+   or a template body.
+
+**Examples:**
+
+Here is an enumeration of common forms of given definitions in the new syntax. We show the following use cases:
+
+ 1. A simple typeclass instance, such as `Ord[Int]`.
+ 2. A parameterized type class instance, such as `Ord` for lists.
+ 3. A type class instance with an explicit context parameter.
+ 4. A type class instance with a named eexplicit context parameter.
```
bjornregnell, line 324: typo: "eexplicit" -> explicit

bjornregnell:
This SIP is up for vote on Friday Sep 27th. As one of the assignees I hereby state my recommendation to the Committee: 

After discussions and feedback [here on Contributors](https://contributors.scala-lang.org/t/pre-sip-improve-syntax-for-context-bounds-and-givens/6576/176) and also updates to the proposal accordingly, I hereby recommend to vote **`Yes`**, as I find that the pros of improvements in ergonomics and regularity outweigh the cons of changing the given-syntax. This also paves the way for coming SIP:s of improvement of type class ergonomics.

bjornregnell:
@sjrd @Kordyjan What are your recommedations to the voters?

sjrd:
I recommend to vote **yes** as well.

PR #93, opened 2024-08-19
Clause interleaving is accepted

bishabosha:


kyouko-taiga:
Oh, thanks for pointing that out. I didn't expect there are so many manual interventions involved in the babysitting of these PRs.

PR #94, opened 2024-08-26
Fix typo

pWydmuch:
I think a small typo managed to sneak in,  shouldn't it be 'c.population' rather than 'p.population'


SethTisue:
thank you!

PR #97, opened 2024-10-21
SIP-67 - Strict equality pattern matching

mberndt123:
Hi there,

I'd like to use the `strictEquality` feature for the improved type safety it provides, but currently find it too inconvenient to use due to an unfortunate interaction with pattern matching. This SIP is my attempt to fix that.
There have been no comments in the [Pre-SIP thread](https://contributors.scala-lang.org/t/pre-sip-better-strictequality-support-in-pattern-matching/6781/6) for the past two weeks, and it's a very small (though impactful) change to the language, so I felt it was time to submit it.

Best regards
Matthias


soronpo:
Thank you for the proposal. Here is the feedback from the SIP Committee:
* We would like to see a different proposal that special-cases enumerations in both pattern matching and equality operations so that enumerations are considered to have a `CanEqual` derivation and the compiler would pick up the default equality if no explicit `derives CanEqual` is given. This will take care of old libraries that do not have `CanEqual` derivation.
* In general, we do not think its good to have a different behavior between `==` and pattern match equality. Do you have such a use-case?

jducoeur:
> Have you seen this type of code in the wild?

Not sure offhand, but it seems like it would be unsurprising when constructing a DSL interpreter.  (Where the enum is a kind of expression, and the leaf is an instantiation of one expression type.)

The interpreter for my own QL language is still on Scala 2, but I could see myself trying to build it along those lines if it was Scala 3, so I don't think that's just a strawman.

soronpo:
> Not sure offhand, but it seems like it would be unsurprising when constructing a DSL interpreter.

Still I'm not sure `enum` is the correct way to do it in Scala 3. It's not meant as a replacement for all kinds of case class hierarchies.

KristianAN:
> > I would not consider this to be a valid use of enum. Have you seen this type of code in the wild?
> 
> I have seen ADTs with functions inside (Hope its ok to consider enum as a nicer syntax for ADT)

From the Scala 3 book.
> Algebraic Data Types (ADTs) can be created with the enum construct, so we’ll briefly review enumerations before looking at ADTs.

Personally I have almost exclusively written Scala 3 and would use enums for this kind of ADT (with a function). If that is not the correct way to do it, then the correct way must be elusive.

jducoeur:
Yeah, agreed.  My understanding of the `enum` feature was that it was more or less exactly to reify the pattern that the community has settled on for ADTs over the years.  (With enumerations per se being sort of the degenerate case, but far from the whole story.)

Saying that only some ADTs count, and other reasonably well-formed ones don't seems kind of un-Scala to me.  A key element of Scala is that functions are values; intuitively, I would expect them to work here like any other value type.

SystemFw:
I have to agree with @jducoeur , what's the actual argument against functions in `enum`? 
`enum` already goes some way into trying to capture some of the more complex ADT patterns that have little relationship with enums, for example allowing `extends` to encode GADTs, having a value in there that happens to be a function seems like a lot more "normal" in comparison.

The GADT mention is not accidental, the main use case for that is encoding commands:
```scala
enum Cmd[A] { 
  case Read() extends Cmd[String]
  case Write extends Cmd[Unit]
}
```
and command-like GADTs will often have functions in them to allow sequencing:
```scala
enum Cmd[A] { 
  case Read() extends Cmd[String]
  case Write extends Cmd[Unit]
  case FlatMap[O, A](fa: Cmd[O], f: O => Cmd[A]) extends Cmd[A]
}
```

soronpo:
Let me clarify. If you want function arguments that should not be part of the pattern match, then these should come as a second argument block of the case, IMO.
```scala
enum Cmd[A]:
  case Read() extends Cmd[String]
  case Write extends Cmd[Unit]
  case FlatMap[O, A](fa: Cmd[O])(val f: O => Cmd[A]) extends Cmd[A]
```
If you want them to be part of the pattern match, you put them in the first block and they can the same equality treatment like all the other arguments.

JD557:
I admit that I never played with strict equality, so this might not make much sense, but:

> Now you might say, OK, then maybe we can make the CanEqual instance available only when all the fields in the enum have a CanEqual instance? I. e. make CanEqual[Foo, Foo] available only when CanEqual[Int => Int, Int => Int] is available? We can simulate this:

Do we need a `CanEqual[Foo, Foo]` for the pattern match? Wouldn't a `CanEqual[Foo, Foo.Bar.type]` suffice?

Now, from what I can tell based on some quick tests, the compiler **really** wants a `CanEqual[Foo, Foo]` for pattern matching, but maybe this could be changed?

mberndt123:
Hi @JD557,

> Do we need a `CanEqual[Foo, Foo]` for the pattern match? Wouldn't a `CanEqual[Foo, Foo.Bar.type]` suffice?
> 
> Now, from what I can tell based on some quick tests, the compiler **really** wants a `CanEqual[Foo, Foo]` for pattern matching, but maybe this could be changed?

Yes, I had this in mind and I was going to propose it – you beat me to it. It can't be done with just a magic `CanEqual` instance, but if we slightly tweak the behaviour of `strictEquality` I think it should work. When either the scrutinee or the pattern or either side of an `==` comparison is a stable identifier, the compiler should require that identifier's singleton type as the corresponding type parameter to the `CanEqual`.

I. e. 
```
def bla(foo: Foo) =
  foo match
    case Foo.Bar => 0 // requires CanEqual[foo.type, Foo.Bar.type]
    case Foo.Baz(f) => f(0) // uses unapply
```

Then all we would need is a magic `CanEqual` instance that spawns when the least upper bound of the two types is a sealed or enum type and either side is the singleton type of a `case object` (for sealed types) or singleton `case` (for enums). That seems like a workable approach.

mberndt123:
> Let me clarify. If you want function arguments that should not be part of the pattern match, then these should come as a second argument block of the case, IMO.
> 
> […]
> 
> If you want them to be part of the pattern match, you put them in the first block where they have same equality treatment like all the other arguments.

First of all, I'd like to address the question of "in the wild" examples of functions within ADTs. This is definitely something that people do, e. g.
https://github.com/typelevel/cats/blob/1cc04eca9f2bc934c869a7c5054b15f6702866fb/free/src/main/scala/cats/free/Free.scala#L219
https://github.com/typelevel/cats-effect/blob/eb918fa59f85543278eae3506fda84ccea68ad7c/core/shared/src/main/scala/cats/effect/IO.scala#L2235

I think this is perfectly good code. It should be possible to use `enum` for these types, and it is today – `strictEquality` shouldn't break that. I also see no reason to compel people to rewrite this to a style where the function goes in the second parameter list. In fact, I think it is _worse_ that way because `==` is supposed to tell you whether two things are the same – but if you just ignore the function because it comes in the second parameter list, you're simply going to get a broken `equals` method that says that two things are the same even though they aren't. It simply shouldn't be possible to perform an equality check with anything other than the singleton cases. For those, we do know that we can get the check right.

mberndt123:
I've pushed a new revision that is based on a minor change to `enum` pattern matching and a magic `CanEqual` instance. @soronpo please let me know what you think

soronpo:
@mberndt123 Hey Mathias, thank you again for the amendments. We discussed the SIP today and we strive for a more general approach: A case class or enum case will be considered to have `CanEqual` derivation when all arguments (recursively) have `CanEqual` derivation. What do you think?

mberndt123:
Hi Oron,

here is my feedback to the committee:

## Singleton typing of `enum case`

The committee has not addressed point 1 of the proposed specification, which is to tweak the behaviour of pattern matching with regard to singleton `enum` `case`s, i. e. this part of the proposal:
> Tweak the behaviour of pattern matching with regards to singleton `enum` `case` patterns. When a singleton `enum` `case` (like `Nat.Zero` in the above example) is used as a pattern, it should be considered  to have that `case`'s singleton type, not the `enum` type. E. g. the above `def +` currently requires a `CanEqual[Nat, Nat]`, and should require a `CanEqual[Nat.Zero.type, Nat]` in the future. This is already the case for `case object`s today. In expressions, singleton `enum` `case`s should continue to have the `enum` type, not the singleton type.

 This change is crucial, as for reasons that were already discussed above, it isn't possible to solve this problem only by providing a magic `CanEqual` instance, irrespective of the exact rules of when that `CanEqual` instance should spawn (unless you want to allow nonsensical instances like `CanEqual[Foo, Foo]`).

Here are two examples to clarify the problem.

Example 1 ([Scastie](https://scastie.scala-lang.org/avPr49JDRKyjt28sVu6MaQ))
```scala
import scala.language.strictEquality
enum Foo:
  case Bar
  case Baz(f: Int => Int)

object Foo:
  given CanEqual[Foo.Bar.type, Foo] = CanEqual.derived

def bla(f: Foo) =
  f match
    case Foo.Bar => 0
    case Foo.Baz(a) => a(0)
```
Example 2 ([Scastie](https://scastie.scala-lang.org/vT8aG9KFS2SIOyZWDbIEGQ)):
```scala
import scala.language.strictEquality
sealed trait Foo
object Foo:
  case object Bar extends Foo
  case class Baz(f: Int => Int) extends Foo

  given CanEqual[Foo.Bar.type, Foo] = CanEqual.derived

def bla(f: Foo) =
  f match
    case Foo.Bar => 0
    case Foo.Baz(a) => a(0)
```

The only difference between these two is that the first uses `enum` to declare the ADT and the second uses a `sealed trait`. Yet, Example 1 does not compile, whereas Example 2 does. Note that both contain the line `given CanEqual[Foo.Bar.type, Foo] = CanEqual.derived` in order to emulate the magic `CanEqual` instance that will eventually be provided by default.

## The magic `CanEqual` instance
The committee has addressed point two of the specification, which is to have a magic `CanEqual` instance, and requested to change the rules for when this `CanEqual` instance should spawn:
> A case class or enum case will be considered to have `CanEqual` derivation when all arguments (recursively) have `CanEqual` derivation (singletons therefore have `CanEqual`).

I am going to try and emulate the specified behaviour by spelling out explicitly the `CanEqual` instances that would be spawned according to this specification.
For `sealed trait` ([Scastie](https://scastie.scala-lang.org/fU4xQzpSRa2IL68CkK43JA)):
```scala
import scala.language.strictEquality

sealed trait Foo
object Foo:
  case object Bar
  case class Baz(f: Int => Int)

  // The specification only mentions `case class`, not `case object`.
  // Sorry, `Foo.Bar`, no `CanEqual` for you :-(
  object Baz:
    given (using CanEqual[Int => Int, Int => Int]): CanEqual[Foo.Baz, Foo.Baz] = CanEqual.derived
```

For `enum` ([Scastie](https://scastie.scala-lang.org/wcJsrEjoTNWX9O7cpUa1NA)):
```scala
import scala.language.strictEquality

enum Foo:
  case Bar
  case Baz(f: Int => Int)

object Foo:
  given CanEqual[Foo.Bar.type, Foo.Bar.type] = CanEqual.derived
  given (using CanEqual[Int => Int, Int => Int]): CanEqual[Foo.Baz, Foo.Baz] = CanEqual.derived
```
This is my best understanding of the wording that the committee supplied.

`def bla` from Example 2 above requires a `CanEqual[Foo.Bar.type, Foo]` instance to compile. The committee's proposal only provides a `CanEqual[Foo.Bar.type, Foo.Bar.type]` and hence does not enable Example 1 to compile. It thus fails at what the SIP is meant to achieve. By contrast, my own version of the magic `CanEqual` does support this case. 
The committee's proposal also does not enable most `==` comparisons between `enum`s because those require `CanEqual` instances for the `enum` type rather than the individual `case`s. 

A perhaps more lenient reading of the committee's wording would be that a magic `CanEqual[A, B]` instance is available when `A` is a subtype of `B`, and `A` is a `case object`,  or a `case class` with fields of types `T_1`, ..., `T_n`, and all of `CanEqual[T_1, T_1], …, CanEqual[T_n, T_n]` are available (plus equivalent wording for `enum`, which I can't be bothered to spell out right now).

I've tried to emulate that too ([Scastie](https://scastie.scala-lang.org/bgoDh59cShCiJwwQZDtIUQ))
```scala
import scala.language.strictEquality

sealed trait Foo
object Foo:
  case object Bar extends Foo
  case class Baz(f: Int => Int) extends Foo

  given [A >: Foo.Bar.type]: CanEqual[Foo.Bar.type, A] = CanEqual.derived
  given [A >: Foo.Baz](using CanEqual[Int => Int, Int => Int]): CanEqual[Foo.Baz, A] = CanEqual.derived
```
And while this does enable `def bla` to compile, it also allows all sorts of nonsense such as `summon[CanEqual[Foo.Bar.type, Int]]` (because `CanEqual[Foo.Bar.type, Any]` is available and contravariance is a thing). So this is not a way to fix it. 

Thinking about this made me realize that my own specification is also not restrictive enough and allows for nonsense ([Scastie](https://scastie.scala-lang.org/NfP45MKzRkOxsbwwQ2DKrg)).
```scala
import scala.language.strictEquality

sealed trait Foo
object Foo:
  case object Bar extends Foo
  case class Baz(f: Int => Int) extends Foo

  given CanEqual[Foo.Bar.type, Foo | X] = CanEqual.derived
  given (using CanEqual[Int => Int, Int => Int]): CanEqual[Foo.Baz, Foo | X] = CanEqual.derived

sealed trait X
object X:
  case object Y extends X

summon[CanEqual[Foo.Bar.type, X]] // ew 🤢
```
I wrote it that way to enable pattern matching for _unions_ of `sealed` and/or `enum` types, but apparently that can't be done safely in this way. I'll tighten my proposal accordingly. This is not a huge problem because unions of `sealed` types can be dealt with by nesting the `match`es. 

So perhaps the rule should be: A magic `CanEqual[A, B]` instance is available when `A` is a subtype of `B`, and `B` is a `sealed` type, and `A` is a `case object`,  or a `case class` with fields of types `T_1`, ..., `T_n`, and all of `CanEqual[T_1, T_1], …, CanEqual[T_n, T_n]` are available (plus equivalent wording for `enum`). So now you can compare `sealed` types to their derived `case class`es and `case objects`. But is that actually useful? Well maybe… But I would say that to _properly_ generalize it, you need to make it work for the `sealed` types themselves too, not just for the `case class`es.

So we're going to need an additional rule: `CanEqual[A, A]` is also available if `A` is a `sealed` type, and that for all types `T_1, …, T_n` that extend it, a `CanEqual[A, T_1]` is available. So for the `Nat` type, it would look more or less like this ([Scastie](https://scastie.scala-lang.org/2Jipf8ipRhatkdxh2Ct8DA))
```
import scala.language.strictEquality

sealed trait Nat:
  def +(that: Nat): Nat =
    this match
      case Nat.Zero => that
      case Nat.Succ(x) => Nat.Succ(x + that)

object Nat:
  case object Zero extends Nat
  case class Succ(n: Nat) extends Nat

  given (using CanEqual[Nat.Zero.type, Nat], CanEqual[Nat.Succ, Nat]): CanEqual[Nat, Nat] = CanEqual.derived

  given CanEqual[Nat.Zero.type, Nat] = CanEqual.derived
  given (using CanEqual[Nat, Nat]): CanEqual[Nat.Succ, Nat] = CanEqual.derived

summon[CanEqual[Nat, Nat]]
```

But… That doesn't work either because `Nat` is recursive and so we're caught in a vicious circle 🙄
So it gets even more complicated: `CanEqual[A, A]` is available if `A` is a `sealed` type, and that for all types `T_1, …, T_n` that extend it, a `CanEqual[A, T_1]` is available under the assumption that `CanEqual[A, A]` is available.

mberndt123:
Hi Oron,

I have considered the committee's proposal and would like to present my conclusions.

1. When applied to the motivating `Nat` example in the SIP, the committee's proposal would make `CanEqual[Nat.Zero.type, Nat.Zero.type]` available (it would not make `CanEqual[Nat.Succ, Nat.Succ]` available since its only field is of type `Nat` and a `CanEqual[Nat, Nat]` is not available). However, `def +` from that example requires a `CanEqual[Nat, Nat]` instance to compile. Therefore, the committee's proposal does not solve the problem which this SIP is intended to solve. If `Nat` had been specified as a sealed trait hierarchy, the pattern match in `def +` would require a `CanEqual[Nat.Zero.type, Nat]`, which is also not available, so it doesn't work for this case either.
1. The committee's proposal also does not make the `==` operator work for enum types, because comparing these with `==` again requires a `CanEqual` instance for the `enum` type, not for the individual cases.
1. A modified version of the committee's proposal that makes `CanEqual` available for enum types iff all of their cases' fields' types have a `CanEqual` will also not fix pattern matching as it cannot work with the [`enum Foo` example from above](https://github.com/scala/improvement-proposals/pull/97#issuecomment-2498101045).

What the committee's proposal does do is make `==` work for case classes. However, I have some observations about this too:

4. For deeply nested data structures, the recursive search for `CanEqual` instances would lead to an increase in compile times. This can not happen in my own proposal as it only provides a magic `CanEqual` instance for case objects and singleton enum cases – the fields never enter the picture. That is not a bug, it is a feature.
1. It is not immediately obvious if and how it would work for recursive definitions such as `case class Nat(n: Option[Nat])`
1.  I am also concerned about maintainability. Imagine you use a library that defines some case class. You use `==` all over the place to compare it, and it works due to the magic `CanEqual` instance for case classes. One day the library author decides to change the case class. Maybe he changes one of the fields' types to an opaque type for better type safety. Now there is no longer a `CanEqual` available for that field and your code no longer compiles. And if this happens in some complex, deeply nested data structure, figuring out where that breakage is coming from is going to be a nightmare.
1. Making the "magic" `CanEqual` instance available only when `CanEqual` is available for the fields' types makes this a safer option than simply adding `derives CanEqual` to the case class, because it won't make `CanEqual` available for types that shouldn't have it, whereas `derives CanEqual` will. This encourages developers to _never_ use `derives CanEqual`, making all of the above problems _even worse_.

For all of these reasons, I think "fully automatic" typeclass derivation (i. e. typeclass derivation that works without an explicit `derives Something` clause) is a bad idea in general, and the language/standard library shouldn't set a bad example in that regard.

To sum up: the committee's proposal doesn't solve the problem that the SIP is designed to solve. It doesn't make the `==` operator work for enum types either. It does make `==` work for case classes, but with many downsides that should be avoided. I therefore do not consider the committee's proposal viable. I had already pointed out in [my first comment](https://github.com/scala/improvement-proposals/pull/97#issuecomment-2498101045) that “it's impossible to make this work correctly by providing a magic CanEqual instance”. This continues to be true and is the reason why my revised proposal also includes language regarding the behaviour of pattern matching, which is required to enable a solution based on a `CanEqual` instance.

Thinking about this issue once more made me realize that my own revised proposal has a safety problem. It mentions a "union or intersection of one or more `sealed` or `enum` types". This was meant to allow pattern matching against unions of different `sealed` or `enum` types without nesting `match`es, but it can also be used to perform equality tests that don't make sense and should be prevented. Since the revised proposal does not have any other upsides over the original proposal and also failed at its other goal of being acceptable to the committee, I've decided to stop pretending and reverted the SIP proposal to its original state.

ADT pattern matching and equality tests are separate things, and the fact that Scala has conflated them is a historical mistake. The best way to move forward now is to untangle them as far as possible: simply don't require a `CanEqual` instance for those cases that correspond to traditional ADT pattern matching, and require it for those cases where it makes sense, like matching a `Vector` against `Nil` or an `Int` against a named constant.

Best Regards
Matthias Berndt

soronpo:
We had a SIP discussion, and we'll vote on it next time, but I was just thinking about it and the specification needs to account for any singleton that is a sub-type of the matched type, which unreachability already does check for.
I think this should compile (but currently under your spec it does not):
```
import scala.language.strictEquality
def foo(x: Any): Unit =
  x match
    case 5 =>
    case "hello" =>
```
I don't know what to think about the function type example now.

mberndt123:
Hey @soronpo

Yes, you're right, that code does not compile under the rules I've proposed: `Any` cannot be equality-compared to an `Int` or a `String`.
I think that's fine. The whole point of `-language:strictEquality` is to increase type safety, and `Any` is the antithesis of that: it completely throws any type information out of the window. Therefore I don't think that this feature needs to support that use case, and nor is this the kind of code that the language should encourage or be optimized for.
The vast majority of reasonable use cases should be covered by the `CanEqual` instances that are available by default like `CanEqual[String, String]` or even `CanEqual[Long, Int]`. 

I also don't think that special-casing literals is a good solution here because it should always be possible to use a named constant instead of a literal with a minimum of fuss. IOW, if it works for the pattern `"hello"`, it should also work for the pattern `H` where `val H = "hello"`. At that point, Strings and Ints are just values like any other and should be treated according to the same rules.

It should also be pointed out that strings, unlike case objects and singleton enum cases, are not singletons.
```scala
val a = "a" * 10
val b = "aa" * 5
a eq b // false
a == b // true
```
So the philosophical reasoning described in the SIP does not apply to Strings.

bjornregnell:
I agree with @mberndt123 that Any-comparisons are dubious and the whole point of strictEquality is making risky business not compile. However, I think it is very important to have good error messages so that a learner can understand *why* such a match compiles. 

I'd like to see a future where structEquality is the default, but to get there we need to make it *really ergonomic* and error messages, docs and other help to the user/learner is critical to making that happen. After all safety is next to scalability the unique selling-point of Scala and we should always prioritize sound and ergonomic language features that improve safety, IMHO.

soronpo:
> The whole point of -language:strictEquality is to increase type safety

I think the point is for *equality*. Pattern matching does not fall under this category beyond reachability concerns, IMO.

bjornregnell:
But `equals` is called in pattern matching with constant patterns, so it's an equality test?

bjornregnell:
Aha, thanks @soronpo for the clarification.

bjornregnell:
@mberndt123 Perhaps update the proposal to upfront clarify the scope of being `==` and not `equals` ? When re-reading the intro etc I'm not sure this is absolutely clear.

sjrd:
Pattern matching uses `==`. It doesn't use `equals`. Proof:
https://scastie.scala-lang.org/A0eckzOaRxaygBFoCMiyWA
```scala
locally {
  val x: Any = 5L
  println(x == 5)      // true
  println(x.equals(5)) // false
  println(x match {    // yes
    case 5 => "yes"
    case _ => "nope"
  })
}
```

mberndt123:
> @mberndt123 Perhaps update the proposal to upfront clarify the scope of being `==` and not `equals` ? When re-reading the intro etc I'm not sure this is absolutely clear.

Hi Björn,

What @soronpo wrote there is not a clarification, it is his opinion, and it is in fact exactly where the controversy is. Currently, pattern matching against a constant (whether it's a literal or a case object or just a `val`) requires a `CanEqual` instance. @soronpo believes that this should be dropped entirely. For reasons that I've laid out in [in the SIP](https://github.com/scala/improvement-proposals/pull/97/files#diff-925644a27cee51dbcd229d7d9230cafeb1cf7ea890c66671551d76d3115ecbcdR125), I don't think that this is sufficiently safe and I would rather drop the `CanEqual` requirement only for cases that correspond to traditional ADT pattern matching in statically typed languages like Haskell, OCaml or Rust.

bjornregnell:
AHA!! Thanks also to @mberndt123 and @sjrd for yet more clarifications. Maybe it's just me who didn't get the actual controversy, but perhaps the intro to the SIP can be improved by referring to `==` versus `equals` and later state what is at stake in terms of compiling or not for a busy reader...

I think I'm leaning towards the stricter end of the design space of the future language - it is after all called *strict* equality.

soronpo:
> Pattern matching uses `==`. It doesn't use `equals`. Proof: https://scastie.scala-lang.org/A0eckzOaRxaygBFoCMiyWA
> 
> ```scala
> locally {
>   val x: Any = 5L
>   println(x == 5)      // true
>   println(x.equals(5)) // false
>   println(x match {    // yes
>     case 5 => "yes"
>     case _ => "nope"
>   })
> }
> ```

What I meant is that I think it should be equals and therefore not affected by strict equality

soronpo:
This SIP was approved to move into the implementation stage. Congrats @mberndt123 ! 
We think that it's important for future readers to add a couple of examples of what is and isn't affected by this SIP.
My own view is that we can eliminate the need for CanEqual inside of pattern matching, but we can do this "in baby steps", and this is a step in the right direction.
Regarding the implementation, would you like to try an implement this SIP?

SethTisue:
> We think that it's important for future readers to add a couple of examples of what is and isn't affected by this SIP.

I think this would help both the committee, and the users who want to try the changes out. Users especially have trouble reading and digesting spec-ese.

It's useful not only to have examples of what the proposal _does_ change, but what it doesn't. The full story of the proposal is: here's what already worked even before the proposal; here's examples that the proposal does fix; here's examples that don't work and it might be surprising but we're certain the behavior is correct; here's examples that don't work and we might have hoped to fix but it's out of scope for now but maybe a subsequent SIP could address it.

I know that may sound like a lot, but I'm not imagining tons of examples, just key ones that would make the SIP clear to everyone.

mberndt123:
Thanks @soronpo and @SethTisue.

I've added a handful of examples, I hope that clarifies things

bracevac:

Thanks!

nilskp:
New to this SIP. I always thought Scala would implement strict equality akin to Javascript, with a triple equals `===`. I'm sure this has been discussed, but perhaps someone can point me to where?

soronpo:
> New to this SIP. I always thought Scala would implement strict equality akin to Javascript, with a triple equals `===`. I'm sure this has been discussed, but perhaps someone can point me to where?

Such discussions would usually occur on https://contributors.scala-lang.org/ but I don't remember such discussion.

jducoeur:
> New to this SIP. I always thought Scala would implement strict equality akin to Javascript, with a triple equals ===. I'm sure this has been discussed, but perhaps someone can point me to where?

Note that `===` is implemented in some libraries (notably Scalactic) -- for those who want to opt into that, it's long been available.

Strict equality is much further-reaching than that, though, intentionally -- it's not just opt-in on an ad hoc call-site basis, it *prevents* you from handling equality in an ad-hoc way, and enforces some discipline instead.  (Still deciding how much I like it in practice, but I appreciate the principle.)

mberndt123:
Hi there,

I've updated this SIP to incorporate feedback that I've gathered. I’d appreciate it if the committee could take another look!
@soronpo

SethTisue:
@soronpo @Gedochao it's a bit unfortunate that we skipped a SIP meeting or two, as this could very likely have been approved by the committee in time for 3.9.0-RC1

but since the user feedback has been uniformly positive, and since it only affects an experimental feature, perhaps it's actually not too late for 3.9?

soronpo:
> @soronpo @Gedochao it's a bit unfortunate that we skipped a SIP meeting or two, as this could very likely have been approved by the committee in time for 3.9.0-RC1

You mean approve the update *as experimental*? I don't think we need even a meeting for that since changes being made are an inherent part of the experimental phase. However, we cannot promote it to a stable feature before 3.10.

SethTisue:
> we cannot promote it to a stable feature before 3.10

right, agreed, we're just trying to see if https://github.com/scala/scala3/pull/25850 can make 3.9

sorry, I should have been clearer

SethTisue:
Core agreed today that the implementation can land in 3.9, assuming no new doubts arise at the April 24 SIP meeting.

Reasoning: the feature is experimental anyway, and the implementation changes don't appear risky.

soronpo:
Great news @mberndt123 !

bjornregnell:
@mberndt123 One minor thing: In example 1 there is reference to a "compiler bug" -- for the record and for later reading when this is merged it would be good if you could insert a link to an actual issue on the compiler so it will be easy to eventually see when it is fixed for the ones landing on the sip text in any kind of future...
```
Welcome to Scala 3.9.0-RC1-bin-20260423-5d6b9a4-NIGHTLY-git-5d6b9a4 (25.0.2, Java OpenJDK 64-Bit Server VM).
                                                                                                                              
scala> def foo(vector: Vector[Int]) =
         vector match
           case Nil => 0
           case _ => 1
       
1 warning found
def foo(vector: Vector[Int]): Int
-- [E030] Match case Unreachable Warning: ------------------------------------------------------------------------------------
3 |    case Nil => 0
  |         ^^^
  |         Unreachable case
                                                                                                                              
scala> foo(Vector())
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$ (file:/home/bjornr/.cache/coursier/v1/https/repo.scala-lang.org/artifactory/maven-nightlies/org/scala-lang/scala-library/3.9.0-RC1-bin-20260423-5d6b9a4-NIGHTLY/scala-library-3.9.0-RC1-bin-20260423-5d6b9a4-NIGHTLY.jar)
WARNING: Please consider reporting this to the maintainers of class scala.runtime.LazyVals$
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
val res0: Int = 0
```
(res0 should be 1...)

bjornregnell:
...and many thanks for your thorough investigations and well-written SIP :)  @mberndt123

mberndt123:
@bjornregnell
I haven't been able to find a bug report for this, so I've made one (https://github.com/scala/scala3/issues/25933). I also added links to the implementation PRs.
Thank you for your feedback.

soronpo:
Hey @mberndt123 this SIP was accepted to be shipped 🥳. We only require two minor clarifications:
1. The PR title should be changed to match the SIP document title.
2. The non-case object amendment isn't clearly visible in the document. The specification part mainly mentions case objects and not regular objects.

Thank you for all the work!

mberndt123:
Thanks @soronpo, I've made the requested changes. I'll make a PR in scala/scala3 soon to enable this by default.

Gedochao:
Enabled by default in https://github.com/scala/scala3/pull/26424

julienrf:
Should the PR be merged?

PR #99, opened 2024-11-22
Change for SIP-62: betterFors

KacperFKorban:
A proposal of a new approach for implementing change `3` from SIP-62: betterFors.

TLDR:

Don't drop the trailing `map` when desugaring. Instead, mark the last `Apply` node with a `TrailingForMap` attachment and move the optimization to a later phase (after typing). This way, we can make sure that the resulting type of the `for` doesn't change.

### DIFF
--- a/content/better-fors.md
+++ b/content/better-fors.md
@@ -10,10 +10,11 @@ title: SIP-62 - For comprehension improvements
 
 ## History
 
-| Date          | Version            |
-|---------------|--------------------|
-| June 6th 2023 | Initial Draft      |
-| Feb 15th 2024 | Reviewed Version   |
+| Date          | Version                |
+|---------------|------------------------|
+| June 6th 2023 | Initial Draft          |
+| Feb 15th 2024 | Reviewed Version       |
+| Nov 21th 2024 | Addendum for change 3. |
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


KacperFKorban:
I suppose that the manager stays the same. ping @lihaoyi

lihaoyi:
Ping @sjrd

content/better-fors.md
```diff
@@ -333,6 +334,60 @@ doSth(arg).map { a =>
...
+def run =
+  val runtime = Runtime.default
+  Unsafe.unsafe { implicit unsafe =>
+    runtime.unsafe.run(loop).getOrThrowFiberFailure()
+  }
+```
+
+This kind of effect loop is pretty commonly used in Scala FP programs and often ends in `yield ()`.
+
+The problem with the desugaring of this for-comprehensions is that it leaks memory because the result of `loop` has to be mapped over with `_ => ()`, which often does nothing.
```
He-Pin, line 387: Will it then be desugared to two `foreach`? as the `_` will never be used anyway.

KacperFKorban: In this case, it should end up with just one `flatMap`. It will eliminate the last `map` call. This is also what bm4 would do in this situation.

lihaoyi:
Just a bump on this @sjrd, or if you have anyone else on the compiler side who is able to review this. I don't know enough about Dotty internals to give this a proper review myself

sjrd (approved):
Looks like a very good fix to the issues we encountered.

odersky:
The implementation has been merged in https://github.com/scala/scala3/pull/22619.

PR #101, opened 2024-12-20
SIP-69: Existential containers

kyouko-taiga:


content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+trait Polygon extends TypeClass:
+  extension (self: Self)
+    def area: Double
+
+given Square is Polygon: ...
+given Hexagon is Polygon: ...
+```
+
+Defining `Polygon` as a type class rather than an abstract class to be inherited allows us to retroactively state that squares are polygons without modifying the definition of `Square`.
+Sticking to subtyping would require the definition of an inneficient and verbose wrapper class.
```
SethTisue, line 48:
```suggestion
Sticking to subtyping would require the definition of an inefficient and verbose wrapper class.
```

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+
+Alas, type classes offer limited support for type erasure–the eliding of some type information at compile-time.
```
SethTisue, line 50: you mean elided at runtime, I think?

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+largest(List(Square(), Hexagon()))
+// error: No given instance of type Polygon{type Self = Square | Hex} was found for a context parameter of method largest
+```
+
+The call to `largest` is illegal because, although there exist witnesses of the `Polygon` and `Hexagon`'s conformance to `Polygon`, no such witness exists for their least common supertype.
+In other words, it is impossible to call `largest` with an heterogeneous sequence of polygons.
+
+## Proposed solution
+
+The problems raised above can be worked around if, instead of using generic parameters with a context bound, we use pairs bundling a value with its conformance witness.
```
SethTisue, line 67:
```suggestion
The problems raised above can be worked around if, instead of using generic parameters with a context bound, we use pairs bundling each value with its conformance witness.
```

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+```scala
+trait AnyPolygon:
+  type Value: Polygon as witness
+  val value: Value
+
+def largest(xs: Seq[AnyPolygon]): Option[AnyPolygon] =
+  xs.maxByOption((a) => a.witness.area(a.value))
+```
+
+The type `AnyPolygon` conceptually represents a type-erased polygon.
```
SethTisue, line 79: I don't understand why we're talking about type erasure here. Maybe it makes sense somehow, but it's not explained sufficiently for me to follow it.

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Compatibility
+
+The change in the syntax does not affect any existing code and therefore this proposal has no impact on source compatibility.
+
+The semantics of the proposed feature is fully expressible in Scala.
+Save for the implicit addition of `.value` on method selection when the receiver is an instance of `Containing[C]`, this proposal requires no change in the language.
+As a result, it has no backward binary or TASTy compatibility consequences.
+
+### Feature Interactions
```
SethTisue, line 148:
```suggestion
### Feature interactions
```

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+### Other concerns
+
+This document has been written under the experimental modularity improvements for Scala 3.
+Although the proposed feature is fully expressible without those changes, the encoding of existential containers can only work with the "old" (i.e., the one currently used in production) or "new" type class style.
+
+### Open questions
+
+One problem not addressed by the proposed encoding is the support of multiple type classes to form the interface of a specific container.
+For example, one may desire to create a container of values whose types conform to both `Polygon` _and_ `Show`.
+We have explored possible encodings for such a feature but decided to remove them from this proposal, as support for multiple type classes can most likely be achieved without any additional language change.
```
SethTisue, line 164: Is the intent to answer this more fully before the proposal comes to a vote? Or do we expect it to remain an open question until some future iteration after the SIP lands?

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+
+Another open question relates to possible language support for shortening the expression of a container type and/or value.
```
SethTisue, line 166: Ditto on whether this might get filled in soon, or it's definitely a "someday" thing?

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Related work
+
+Swift support existential containers.
```
SethTisue, line 170:
```suggestion
Swift supports existential containers.
```

content/existential-containers.md
kyouko-taiga: I did mean compile-time but the sentence can be phrased differently. The point is that you can't easily write a list of things that have the same type class. Meanwhile, it is easy to write a list of things that have the same upper bound.

content/existential-containers.md
kyouko-taiga: I realize I may simply have a different definition of "type erasure". I will rephrase.

content/existential-containers.md
kyouko-taiga: I would propose that as a future proposal, if necessary. That said it is likely that the SIP committee would not have to formally weigh in this addition because it would only relate to the way existential containers are encoded in the library, without any other change to the language.

content/existential-containers.md
kyouko-taiga: That's a someday thing but, again, these open questions can hopefully be answered without introducing more changes to the language.

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+  xs.maxByOption(_.area)
+
+largest(List(Square(), Hexagon()))
+// error: No given instance of type Polygon{type Self = Square | Hex} was found for a context parameter of method largest
+```
+
+The call to `largest` is illegal because, although there exist witnesses of the `Polygon` and `Hexagon`'s conformance to `Polygon`, no such witness exists for their least common supertype.
+In other words, it is impossible to call `largest` with an heterogeneous sequence of polygons.
+
+## Proposed solution
```
lihaoyi, line 65: This section is under-specified. Reading it, I have no idea what the proposed solution is. 

* Probably a lot of the text from `Specification` should actually be up here
* We need several more use case examples, which are less contrived than `Square` `Rectangle` and `Polygon`

Someone should be able to read the proposed solution have _zero idea how it is implemented_, and still get the general idea of what the proposed language feature is about. Right now, that is not the case, so the proposal is incomplete

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+In contrast, to avoid possible undesirable complications, this proposal does not suggest any change to the subtyping relation of Scala.
+
+Rust also supports existential containers in a similar way, writing `dyn P` to denote a container bundling some value of a type conforming to `P`.
+Similar to Swift, existential containers in Rust are considered supertypes of the types conforming to their bound.
+
+
+A more formal exploration of the state of the art as been documented in a research paper presented prior to this SIP [2].
+
```
lihaoyi, line 194: This proposal is missing a section on `Alternatives`. I can think of a few myself:

* What if we just add a `Containing[T]` type to the standard library; people could then use it without needing a language/compiler change at all?

* What about implicit conversions and implicit constructors? Those are widely used today, often referred to as the "magnet pattern"

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+  xs.maxByOption((a) => a.witness.area(a.value))
+```
+
+The type `AnyPolygon` conceptually represents an arbitrary polygon.
+It consists of a pair containing some arbitrary value as well as a witness of that value's type being a polygon.
+We call this pair an _existential container_, as a nod to a similar feature in Swift, and the remainder of this SIP explains how to express this idea in a single, type-safe abstraction.
+
+### Specification
+
+Existential containers are encoded as follows:
```
lihaoyi, line 85: It's not clear to me which part of this specification is user-land library code, and which section is compiler/language-level changes. Perhaps we can split it explicitly into two sections?

content/existential-containers.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Open questions
+
+One problem not addressed by the proposed encoding is the support of multiple type classes to form the interface of a specific container.
+For example, one may desire to create a container of values whose types conform to both `Polygon` _and_ `Show`.
+We have explored possible encodings for such a feature but decided to remove them from this proposal, as support for multiple type classes can most likely be achieved without any additional language change.
+
+Another open question relates to possible language support for shortening the expression of a container type and/or value.
+
+## Related work
```
lihaoyi, line 168: Apart from Swift, what other languages have a feature like this? Does Haskell have something similar? Or Rust? Both of those languages have typeclass-like features and presumably would encounter the same issues around e.g. hetoregenous collections. How do they solve it, or do they not?

content/existential-containers.md
kyouko-taiga: I can write other examples but I disagree that the one in the document is contrived. It also generalizes very easily to any situation involving a set of classes conforming to the same type class. Any instance of a situation where the "upper bound" of a set of types is defined retroactively would look like the proposed example.

```
def lookup(key: List[Containing[Hashable]]) = ???
def zipAll[E](xss: List[Containing[Iterator]{ type Element = E }]) = ???
```

Instead of justifying existential containers, however, I propose to clarify that the SIP is about sugaring the selection of a method rather than proposing the containers themselves, as those can be defined in library space.

lihaoyi: You really need to justify both the language changes and library support. The language change is useless without the library support, but most people would not be familiar with this library-level technique.

`List[Containing[Hashable]]` is definitely a lot more concrete than `Seq[Containing[Polygon]]`. We need more examples like that. 

* I've found use for the equivalent of `Seq[Containing[upickle.default.Writer]]` myself, which contains a list of things that can be converted to JSON

* Scalatags currently uses methods taking `(args: Frag*)` with implicit conversions from various types to `Frag`. it could conceivable instead use `(args: Containing[Fragable]*)` with a `Fragable` typeclass to render the value to HTML. Is that better or worse? Why?

jducoeur:
Comment from the peanut gallery: I agree with Haoyi that the structure of the proposal is currently a little confusing.  Specifically, since it jumps directly to implementation details, it's very unobvious what the userland code would look like.  I'd recommend putting a motivating example of "this is what the end result would look like" before getting into the details of the encoding of `Containing`.

That said, I love the meat of this proposal -- it's reifying exactly the pattern that I've had to build by hand a number of times for various systems (and which I've been occasionally teaching to my teams), so it would be a real win IMO if we had an official and concise solution.

content/existential-containers.md
```diff
@@ -204,3 +253,36 @@ This overhead drops below 10% on larger, more realistic benchmarks [2].
...
+
+```scala
+import language.experimental.{clauseInterleaving, modularity}
+
+/** A type class. */
+trait TypeClass:
+  type Self
+
+/** A value together with an evidence of its type conforming to some type class. */
+sealed trait Containing[Concept <: TypeClass]:
```
bishabosha, line 269: does it only work with this upper bound or can it be `{ type Self <: AnyKind }`

content/existential-containers.md
bishabosha: > support for multiple type classes can most likely be achieved without any additional language change.

you suggest that you can synthesize a single `witness: Polygon & Show` that can resolve extension methods from either without a language change?

content/existential-containers.md
```diff
@@ -0,0 +1,288 @@
...
+
+## References
+
+1. Stefan Wehr and Peter Thiemann. 2011. JavaGI: The Interaction of Type Classes with Interfaces and Inheritance. ACM Transactions on Programming Languages and Systems 33, 4 (2011), 12:1–12:83. https://doi.org/10.1145/1985342.1985343
+2. Dimi Racordon and Eugene Flesselle and Matt Bovel. 2024. Existential Containers in Scala. ACM SIGPLAN International Conference on Managed Programming Languages and Runtimes, pp. 55-64. https://doi.org/10.1145/3679007.3685056
+
+
+## Appendix
+
+The following is a possible implementation of existential containers.
```
smarter, line 259: Here's a version that doesn't rely on the experimental modularity language import (and doesn't require the typeclass to extend `TypeClass`):
```scala
/** A value together with an evidence of its type conforming to some type class. */
sealed trait Containing[TC[_]]:
  /** The type of the contained value. */
  type Value: TC as witness

  /** The contained value. */
  val value: Value

object Containing:

  /** Wraps a value of type `V` and a corresponding typeclass instance `TC[V]` into a `Containing[TC]`. */
  def apply[TC[_]](v: Any)[V >: v.type](using TC[V]): Containing[TC] =
    new Containing[TC]:
      type Value >: V <: V
      val value: Value = v
```

content/existential-containers.md
```diff
@@ -0,0 +1,288 @@
...
+
+def shortest[A: Sizeable, B: Sizeable](a: A, b: B): Containing[Sizeable] =
+  if b.size < a.size then Containing(a) else Containing(b)
+```
+
+Further motivation for existential containers in Scala have been described in a research paper [2].
+
+### Specification
+
+Assuming the existence of an abstraction named `Containing[TC]` for representing containers pairing an arbitrary value with a witness of its conformance to some type class `TC` in the standard library, the compiler injects the selection of the `value` field implicitly when a method of `Containing[TC]` is selected.
```
smarter, line 131: Since this is in the `Specification`section, can I assume that "the compiler injects ..." is in fact part of what this SIP specifies? If so, it should be fleshed out:
- How does this interact with other adaptations, including implicit conversions and apply insertion?
- In which condition does this adaptation triggers exactly? Is Containing treated specially by the compiler or can we (and should we) generalize this?

Otherwise, this should be in a separate section (example or appendix)

PR #105, opened 2025-02-28
SIP-70 - Flexible Varargs

lihaoyi:


content/flexible-varargs.md
```diff
@@ -0,0 +1,350 @@
...
+Apart from the manual workflows described above, one alternative is to use implicit conversions
+to a target type, i.e. the "magnet pattern". For example:
+
+```scala
+case class Summable(value: Seq[Int])
+implicit def seqSummable(value: Seq[Int]) = Summable(value) 
+implicit def singleSummable(value: Int) = Summable(Seq(value))
+implicit def optionSummable(value: Option[Int]) = Summable(value.toSeq)
+
+def sum(x: Int*) = x.sum
```
JD557, line 254: I suspect this was supposed to be
```suggestion
def sum(x: Summable*) = x.sum
```

to trigger the implicit conversions.

lihaoyi: Thanks!

som-snytt:
On the pattern side, the old gotcha is https://github.com/scala/bug/issues/7623 (which has a lint on Scala 2). That is where two pat vars are satisfied by a single Seq, i.e., there is a kind of misalignment. (The example may temper ambition for patterns.)

Atry:
Do you want to cite https://github.com/scala/improvement-proposals/pull/41 ? There are a lot of previous discussions on this feature.

prolativ:
I started to wander if this SIP could help resolve this issue: https://github.com/scala/scala3/issues/18009
In short, the problem there is that an invocation of a curried method added by a refinement on a subtype of `Selectable`, e.g.
```scala
mySelectable.foo(args1*)(args2*)
```
should get desugared to something like
```scala
mySelectable.applyDynamic("foo")(args1*, args2*).asInstanceOf[Foo]
```
which is now illegal, but would be when this SIP gets implemented

odersky (approved):
What's proposed:

 1. Constructing varargs with multiple splices xs*, interspersed with normal values
 2. Deconstructing sequences with a single vararg pattern that can be somewhere in the middle of the sequence, not just at the end.

These are clear quality of life improvements. The construction part is the more important one, and it should be straightforward to implement. The pattern matching part is probably also easy to implement, but it might need some care to get good performance for linear sequences. E.g you have a long list and match with `List(x, y, ys*, 2, 3)` you want to avoid multiple trips to the end of the list to match the `2` and `3`. Nevertheless, it looks quite doable.

lrytz:
I also think this SIP is a very nice improvement.

You mention / suggest to use `IArray`, which is an opaque type alias and not part of the collection hierarchy. `sum(IArray.newBuilder[Int].addOne(1).result()*)` wraps the resulting array using `IArray.wrapIntArray`, which converts it to an `immutable.ArraySeq`. So it's probably better to use `ArraySeq` directly. Or maybe `Vector`?
  - `ArraySeq`: can wrap a primitive array, minimal memory overhead
  - `Vector`: better if the collections is later used for appending / prepending, concatenation etc. Memory and computational overhead is small. Primitives are boxed.

I'm not sure if the `newBuilder` call needs an explicit type argument, or if we can have that inferred by the typer. In Scala 3:

```scala
scala> collection.immutable.ArraySeq.newBuilder.addOne(1).result()
-- [E172] Type Error: ----------------------------------------------------------
1 |collection.immutable.ArraySeq.newBuilder.addOne(1).result()
  |                                        ^
  |                                        No ClassTag available for Any
1 error found
```

Interestingly, the same line works on Scala 2, the compiler infers `newBuilder[Int]`. Is that a known limitation @odersky?

If an explicit type argument is needed, the SIP should specify how that type is obtained.


------

For patterns, we'll need to update the spec to say how `case E(p1, ..., pN, s*, u1, ..., uM)` translates. I think the section is https://scala-lang.org/files/archive/spec/3.4/08-pattern-matching.html#pattern-sequences, the spec seems not to be fully up to date though, it doesn't reflect https://github.com/scala/scala3/pull/11240.

This SIP doesn't change repeated parameters (of case classes), a repeated parameter will still be last in a case classs definition. An `unapplySeq` declaration should also continue to have the corresponding return type `(T1, ..., Tm, Seq[S])`.

So the spec needs to map the new pattern shape onto to current case class parameter list shape (or unapplySeq return type shape). Note that `T1, ..., Tm` can be arbitrary types, all different from each other and different from `S`. So I'm not sure the proposal of translating `case E(p1, ..., pN, s*, u1, ..., uM)` to `case VarargsMatcher(Seq(p1, ..., pN), s, Seq(u1, ..., uM))` would work.

lihaoyi:
> You mention / suggest to use `IArray`, which is an opaque type alias and not part of the collection hierarchy. `sum(IArray.newBuilder[Int].addOne(1).result()*)` wraps the resulting array using `IArray.wrapIntArray`, which converts it to an `immutable.ArraySeq`. So it's probably better to use `ArraySeq` directly. Or maybe `Vector`?
  - `ArraySeq`: can wrap a primitive array, minimal memory overhead
  - `Vector`: better if the collections is later used for appending / prepending, concatenation etc. Memory and computational overhead is small. Primitives are boxed.


The motivation for `IArray` was to try and decouple the language feature from the concrete implementation of the Scala collections library: `IArray` seems like a relatively simple collection with good performance characteristics. In particular, I wanted to avoid using `Seq` which defaults to `List` which is from the start a lot less cache-friendly than `IArray`. `Vector` seems reasonable, but it has the downside of being a pretty specific collection in the standard library and AFAIK something the Scala language was not aware of up to this point.

> This SIP doesn't change repeated parameters (of case classes), a repeated parameter will still be last in a case classs definition.

Yes, definitions are unchanged. Not just for `case class`es, but for method `def`s as well, they all can only have one `vararg*` parameter at the end

> An `unapplySeq` declaration should also continue to have the corresponding return type `(T1, ..., Tm, Seq[S])`.

Yes that is right. The trailing `Seq[S]` can be broken up into a smaller `Seq` in the middle and loose elements before/after it during destructuring, but these elements all continue to be homogeneously typed as `S`, and the signature of the `unapplySeq` does not need to change

odersky:
There is now an [implementation](https://github.com/scala/scala3/pull/23855) of most of the proposal.

One doubt I am having is spreads of `Option`. We do not allow an option to be passed individually to a vararg parameter. I.e. the following would not work:
```scala
  def foo(x: Int*) = x.sum
  val xo = Some(2)
  f(xo*)
```
That's an error since the argument of a spread operator must be a `Seq` or an `Array`. Do we really want to change that? That would affect normal spreads not just spreads embedded in a longer list. There might be use cases for embedded options, but for a single spread passed to a vararg it feels a bit weird. Alternatively, we can demand an explicit `toList`, i.e.
```scala
  f(xs.toList*)
```
would work today. I'd be interested in getting the committee member's opinion on this aspect before I try to implement it.

lihaoyi:
Perhaps a fourth alternative is that we add a new marker trait `trait VarargUnpackable[T]` that `Option` and `Seq` extend, and make `foo(bar*)` work on any `VarargUnpackable`. `VarargUnpackable` could have whatever abstract methods are necessary to efficiently participate in the flexible varargs desugaring

That would avoid hardcoding `Option` and `Seq`, and have the upside that user-specified types will be able to hook into this syntax without having to `extend Seq[T]` themselves which we all know is very error prone and discouraged.

Atry:
Why not just let `Option` and `Seq` extend a newly introduced `trait OrderedIterable[+A] extends Iterable[A]`?

odersky:
I think all this goes to far. Varargs are Seq's. I see no reason to complicate things further here. Arrays can be passed to Seq's since there is an implicit conversion. The point of SIP 70 is to stay within this framework and allow multiple spreads. My PR provides that. Options unfortunately are an irregular case since there is no conversion from option to Seq (and for good reason!). I think it is much cleaner to keep that principle even if some use cases require an explicit `.toList`. 

Here's another way to put it: I can have a vararg function and a spread argument to it:
```scala
def f(x: int*)
f(xs*)
```
I can convert the vararg to a `Seq` and drop that the spread:
```scala
def f(x: Seq[Int])
f(xs)
```
That works, always. I have done refactoring many times. But allowing options in spreads breaks it. 

Generally varargs are tricky to get right since they need to be treated early during type inference when types are not yet known completely. That compounds complexity. So we need to keep it simple.

Ichoran:
Maybe we can add `.cc` as a method that turns a non-collection type into a collection.  `name.cc*` doesn't seem too bad.  Even `x.cc*` doesn't seem too bad to me.  `x.toList*` is a lot more effort.

If we allowed a spread type then we could use a unary `*` suffix operator to simply allow `x.*` while the ordinary collection would be `x*`.  As it is, the vararg type isn't first-class, though, so you can't use it this way.

lihaoyi:
We can probably merge the `Seq`-only implementation first and follow up with the `Option` handling later: follow up discussion in the SIP meeting, follow up PR, or follow up SIP entirely. No need to block the meat of this proposal on an edge case when the bulk of it is uncontroversial

odersky:
Let's discuss this point later today in the SIP meeting.

content/flexible-varargs.md
```diff
@@ -0,0 +1,402 @@
...
+val coll = Seq(foo*, bar*, qux, baz)
+```
+
+## Reference Implementation
+
+A reference implementation of this is shown below, but the language is allowed to implement
+it in any way that returns a `Seq` that has the same elements as the reference implementation:
+```scala
+// User Code
+val total = sum(0, numbers1*, numbers2*, 4) // 10
```
odersky, line 178: The code generated by the implementation for this snippet is
```scala
sum(
  ArraySeqBuilder.ofInt(2 + numbers1.length + numbers2.length)
    .add(0)
    .addArray(numbers1)     
    .addSeq(numbers2)
    .add(4).result)
```

odersky (commented):
Implementation matches proposal except that Option spreads are not implememnted

bjornregnell:
> Maybe we can add `.cc` as a method that turns a non-collection type into a collection. `name.cc*` doesn't seem too bad. Even `x.cc*` doesn't seem too bad to me. `x.toList*` is a lot more effort.

We already have `name.toSeq`, which is not that bad, I think...

bracevac:

Thanks!

sjrd:
FTR, this now ships as an experimental feature in Scala 3.8.

PR #109, opened 2025-04-23
New SIP: Allow fully implicit conversions in Scala 3 with `into`

odersky:


vascorsd (commented):
Personally, when I initially heard about this feature I had a gut feeling I didn't like it, but after reading it carefully I think I changed my mind. It's surprisingly interesting.

I feel the feature and the easiness of the syntax will be abused heavily by some users and probably new users since they will get into contact with it pretty early in their lives in scala since eventually this will all end up in signatures of the standard library almost everywhere and users tend to copy the patterns they see in their language std libs since that's what they believe are good patterns or writing code in the language.

It's gonna for sure generate a lot of contention to explain to users why they should or should not use this instead of typeclasses and why should they even think about them if they have all of this mechanism. Would be interesting to see a deeper exploration on that, the limits and when to choose this encoding and how to guide users to good choices and specially to avoid using `into` all over the place.

Guess we'll have to wait and see who it all pans out.

content/into.md
```diff
@@ -0,0 +1,306 @@
...
+
+We can support implicit conversions to `Modifier`s simply by making `Modifier` an `into` trait:
+```scala
+into trait Modifier ...
+```
+This means implicit `Conversion` instances with `Modifier` results can be inserted without requiring a language import.
+
+Here is a simplified example:
+```scala
+trait Modifier
```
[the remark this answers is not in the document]
bjornregnell: > the syntax will be abused heavily by some

I think it is very difficult to draw general conclusions about this upfront, but my guess is rather that `ìnto` will *not* to any large extent be "abused" in terms of creating unsafe or slow code , as the syntax strikes a good balance in the difficult trade-off among verbosity, expressiveness, ergonomics and clear communication of intent. 

Compared to the old `implicit` modifier, which was used as a mechanism for many completely different intents, this new focused feature retains expressiveness while being clear about intent. We get the signal of "danger" with the Conversion trait but can still opt-in for ergonomics of both library construction and usage, if the use case so requires. 

The big problem that this feature solves is that, IMHO, we should not stay on a trajectory where degraded expressiveness and ergonomics of automatic conversions make certain use cases unfeasible in practice or in principle.

content/into.md
```diff
@@ -0,0 +1,306 @@
...
+  List[into[Keyword]]("if", "then", "else")
+```
+This type-checks since the target type of the list elements is the type parameter of the `List.apply` method which is explicitly instantiated to `into[Keyword]`. On the other hand, if we continue the example as follows we get an error:
+```scala
+  val ifKW: into[Keyword] = "if"
+  val ys: List[into[Keyword]] = List(ifKW, "then", "else")
+```
+Here, the type variable of `List.apply` is not explicitly instantiated
+when we check the `List(...)` arguments (it is just upper-bounded by the target type `into[Keyword]`). This is not enough to allow
+implicit conversions on the second and third arguments.
```
bjornregnell, line 237: Perhaps explain here in simpler terms *why* it is "not enough"? And perhaps give an example that show how to make it type check? Also you sometimes use "expected type" and sometimes "target type" - is there a reason for this? I think "expected type" is easier to understand without specific type inference knowledge, while "target typing" is a more advanced concept...

content/into.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Syntax Changes
+
+```
+LocalModifier     ::=  ...  |  ‘into’
+```
+
+`into` is a soft modifier. It is only allowed on traits and classes.
+
+## Compatibility
```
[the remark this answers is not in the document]
odersky: Changing back or forth it is binary and Tasty compatible. But dropping it will mean that user code that relies on implicit conversions into the trait will get a warning.

content/into.md
odersky: I use "target" in "conversion target types" since its shorter than "expected types of conversions" and lends itself better to a formal definition.

content/into.md
```diff
@@ -0,0 +1,306 @@
...
+Here, the conversion from `String` to `Iterable[Char]` is applied on the results of `flatMap`'s function argument when it is applied to the elements of `xs`.
+
+### Vararg arguments
+
+When applied to a vararg parameter, `into` allows a conversion on each argument value individually. For example, consider a method `concatAll` that concatenates a variable
+number of `IterableOnce[Char]` arguments, and also allows implicit conversions into `IterableOnce[Char]`:
+
+```scala
+def concatAll(xss: into[IterableOnce[Char]]*): List[Char] =
+  xss.foldLeft(List[Char]())(_ ++ _)
```
johnynek, line 121: This is a quadratic cost but a not uncommon kind of mistake. Can we use something as an example that isn't an anti pattern? For instance prepending with a foldRight for instance, or `xss.flatten.to(List)`

content/into.md
```diff
@@ -0,0 +1,319 @@
...
+```scala
+  def ++ (elems: into[IterableOnce[A]]): List[A] =
+    val buf = ListBuffer[A]()
+    for elem <- elems.iterator do // no `.underlying` needed here
+      buf += elems
+    buf.toList
+```
+Inside the `++` method, the `elems` parameter is of type `IterableOnce[A]`, not `into[IterableOne[A]]`. Hence, we can simply write `elems.iterator` to get at the `iterator` method of the `IterableOnce` class.
+
+Specifically, we erase all `into` wrappers in the local types of parameter types that appear in covariant or invariant position. Contravariant `into` wrappers are kept since these typically are on the parameters of function arguments.
```
smarter, line 164: I think the fact that the type of `into` is meaningful from the outside (because of the lower-bound) but needs to be erased inside (because the lower-bound doesn't let you call any method on it) is more confusing than helpful. Either `into` behaves like a normal citizen of the type system and I can understand if something typechecks from looking at the definition of `into` (e.g. if it's `type into[T] = T`), or it's outside the type system (and then it seems like it really is more like a keyword like `=>`).

content/into.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Why Two Different Schemes?
+
+Can we make do with just one scheme instead of two? In practice this would be difficult.
+
+Let's first take a look the `Expr` example, which uses into-as-a-constructor. Could it be rewritten to use into-as-a-modifier?
+This would mean we have to add `into` to the whole `Expr` enum. Adding it to just `Const` is not enough, since `Add` and `Neg` take `Expr` arguments, not `Const` arguments.
+
+But we might not always have permission to change the `Expr` enum. For instance, `Expr` could be defined in a lower level library without implicit conversions, but later we want to make `Expr` construction convenient by eliding `Const` wrappers in some higher-level library or application. With `into` constructors, this is easy: Define the implicit conversion and facade methods that construct `Expr` trees while taking `into[Expr]` parameters.
+With `into` modifiers there is no way to achieve the same.
```
smarter, line 261: This could also be solved by allowing `into` as a modifier on type aliases. Then, the higher-level library is free to define:
```scala
into type intoExpr = Expr
```

content/into.md
```diff
@@ -0,0 +1,319 @@
...
+type ToIterator[T] = into[IterableOnce[T]]
+```
+and then `++`, `flatMap` and other functions could use this alias in their parameter types. The effect would be the same as when `into` is written out explicitly.
+
+## Second Scheme: `into` as a Modifier
+
+The `into` scheme discussed so far strikes a nice balance between explicitness and convenience. But migrating to it from Scala 2 implicits does require major changes since possibly a large number of function signatures has to be changed to allow conversions on the arguments. This might ultimately hold back migration to Scala 3 implicits.
+
+To facilitate migration, we also introduce an alternative way to specify target types of implicit conversions. We allow `into` as a soft modifier on
+classes, traits, and opaque type aliases. If a type definition is declared with `into`, then implicit conversions into that type don't need a language import.
```
steinybot, line 179: Could the `into` modifier be inferred if the companion object contains implicit conversions? This would make migration unnecessary in a lot of cases.

SimY4:
Just curious, why opaque alias is preferred over annotation? This seems as a good fit for altering compiler behaviour via additional context passes as annotation on a type.

I can foresee `into` type appearing in compiler errors (since it's opaque) complicating the understanding the code.

odersky:
> Just curious, why opaque alias is preferred over annotation @into? This seems as a good fit for altering compiler behaviour via additional context passed as an annotation on a type.

No it's the opposite. Annotations should not affect typing. They can affect host interop and code generation though. That's for instance why `infix` and `inline` are modifiers, not annotations. Also, annotations do not get propagated reliably through type inference. For instance they can be dropped by meets and joins.

sjrd:
FTR, I recommend to **accept** this proposal.

The motivation is well laid out. The proposed solution addresses all the concerns I've had over the years about deprecating implicit conversions. Also, I don't think we can make the proposed solution any simpler without weakening it. In other words, IMO this is an excellent proposal.

I would still like to see somewhere written the reason why `into[T] >: T`, instead of the more intuitive `into[T] <: T`. @odersky's explanation offline convinced me at the time, but it should be recorded for posterity.

content/into.md
kyouko-taiga: This particular point has been discussed during the last the SIP committee. The general observation is that information going into a public API should not be inferred; they should be conscious decisions. Hence, the `into` modifier should not be inferred.

content/into.md
```diff
@@ -0,0 +1,319 @@
...
+This is where the `into` type constructor comes in. Here is a signature of a `++` method on `List[A]` that uses it:
+
+```scala
+  def ++ (elems: into[IterableOnce[A]]): List[A]
+```
+The `into` wrapper on the type of `elems` means that implicit conversions can be applied to convert the actual argument to an `IterableOnce` value, and this without needing a language import.
+
+`into` is defined as follows in the companion object of the `scala.Conversion` class:
+```scala
+opaque type into[T] >: T = T
```
prolativ, line 58: Commenting here to keep track of the loose idea of defining this as
```scala
opaque into type into[T] >: T = T
```
(not sure of the recommended order of `opaque` and `into` modifiers though).
Would that make any sense? Could that simplify the SIP by making the type `into` less special?

prolativ: @odersky any thought on this?

content/into.md
```diff
@@ -0,0 +1,328 @@
...
+of the original `Modifier` trait has changed. In summary, upgrading Laminar to use Scala 3 conversions could keep either source compatibility or binary compatibility but not both at the same time.
+
+
+## Syntax Changes
+
+```
+LocalModifier     ::=  ...  |  ‘into’
+```
+
+`into` is a soft modifier. It is only allowed on classes, traits, and opaque type aliases.
```
prolativ, line 285: It was only now that I realized that enums are not included here. Was this intentional? @odersky

content/into.md
odersky: I don't think there's enough of a match to pursue this.

odersky:
I am not sure about status. @kyouko-taiga should be better able to answer this.

content/into.md
prolativ: I tried this snippet with the current nightly compiler (3.7.3-RC1-bin-20250619-b1da8fc-NIGHTLY):
```scala
import scala.language.experimental.into

into enum Foo:
  case Foo1
  case Foo2
```
and compiling resulted in the error:
```
[error] Only access modifiers are allowed on enum definitions
[error] into enum Foo:
[error]           ^^^
```
so it looks like this doesn't work and it wasn't clear for me if that's because of the spec or the current implementation

odersky: Yes. On second thought I think we should leave the spec and implementation as is. The rule that enums take only access modifiers is valuable, and the ability to add `into` as a modifier to enums is a bit dubious since `into` is primarily intended for migration from Scala 2 codebases. But of course there we no enums in Scala 2.

content/into.md
```diff
@@ -0,0 +1,330 @@
...
+  def ++ (elems: into[IterableOnce[A]]): List[A] =
+    val buf = ListBuffer[A]()
+    for elem <- elems.iterator do // no `.underlying` needed here
+      buf += elems
+    buf.toList
+```
+Inside the `++` method, the `elems` parameter is of type `IterableOnce[A]`, not `into[IterableOne[A]]`. Hence, we can simply write `elems.iterator` to get at the `iterator` method of the `IterableOnce` class.
+
+Specifically, we erase all `into` wrappers in the local types of parameter types that appear in covariant or invariant position. Contravariant `into` wrappers are kept since these typically are on the parameters of function arguments.
+For case classes this means that class mirrors will use the transformed version of
```
prolativ, line 165:
```suggestion
This means that class mirrors will use the transformed version of
```
This seems to apply to cases of enums as well, e.g.
```scala
enum MyEnum:
  case A(x: into[Int])
  case B
```
So maybe let's not be too specific in order not to introduce additional doubt whether cases of enums should be considered case classes or not

bracevac:

Thanks!

PR #112, opened 2025-08-15
SIP-72 Dedented Multiline String Literals

lihaoyi:


content/dedented-string-literals.md
```diff
@@ -0,0 +1,547 @@
...
+> quotation marks. The string begins on the first line after the opening quotation marks
+> (""") and ends on the line before the closing quotation marks, which means that neither 
+> of the strings below start or end with a line break:
+> 
+> A multiline string can be indented to match the surrounding code. The whitespace before the
+> closing quotation marks (""") tells Swift what whitespace to ignore before all of the other
+> lines. However, if you write whitespace at the beginning of a line in addition to what’s before
+> the closing quotation marks, that whitespace is included.
+
+### Elixer
```
jeremyrsmith, line 503: *Elixir

SimY4:
Just a thought, instead of having a new syntax construct, we can probably have yet another standard string interpolator. Can be macro to deal with the problem at compile time.

lihaoyi:
@SimY4 the proposal discusses why a string inteprolator wouldn't work as well. Specifically, it won't compose with other interpolators like `sql"..."`, it won't work in literal types like `val x: "..." = ???`, it won't work in literal-only contexts like `fastparse.StringIn()` or `@implicitNotFound("...")`

SimY4:
@lihaoyi on this tho, in terms of what's easier to evolve - a language or a library DSL, I think library is much more flexible in this regard. Libraries can chose to provide DSLs that will deal with margins differently. I did that for my own custom DSLs and it worked pretty great. Users don't have to add strip margin at the end of every string block - it just worked.

Over time you can deprecate and phase out the old DSL from the library. You can't phase it out of the language.

lihaoyi:
This now has a PR implementation in https://github.com/scala/scala3/pull/24185

PR #114, opened 2025-08-24
SIP-XX - Unpack Case Classes into Parameter Lists and Argument Lists

lihaoyi:


content/unpack.md
```diff
@@ -0,0 +1,1084 @@
...
+definition-site argument list via `*`: 
+
+```scala
+case class RequestConfig(url: String, 
+                         connectTimeout: Int,
+                         readTimeout: Int)
+
+def downloadSimple(unpack config: RequestConfig) = doSomethingWith(config)
+def downloadAsync(unpack config: RequestConfig, ec: ExecutionContext) = doSomethingWith(config)
+def downloadStream(unpack config: RequestConfig) = doSomethingWith(config)
```
jducoeur, line 31: Personal opinion: the `doSomethingWith()`s here and elsewhere are introducing more confusion than benefit.  While I had a suspicion where you were going in this proposal, this example just leaves me going, "So what?  That's exactly the same as today."

(Continued below...)

content/unpack.md
```diff
@@ -0,0 +1,1084 @@
...
+def downloadSimple(url: String,
+                   connectTimeout: Int,
+                   readTimeout: Int) = doSomethingWith(config)
+def downloadAsync(url: String,
+                  connectTimeout: Int,
+                  readTimeout: Int,
+                  ec: ExecutionContext) = doSomethingWith(config)
+def downloadStream(url: String,
+                   connectTimeout: Int,
+                   readTimeout: Int) = doSomethingWith(config)
```
jducoeur, line 72: (... continuing from above)

And here, the `doSomethingWith()` bits are downright confusing, leaving me going, "`config`?  Where does `config` come into it?"

So I'd recommend either fleshing out the example method bodies, or omitting them entirely, but this pseudo-code is causing more confusion than good.

sjrd:
I really believe you're overestimating the proportion of scenarios of a) evolve from pre-`unpack` to post-`unpack` versus b) evolve post-`unpack` to extended post-`unpack`.

Scala does not have a history of designing new replacement features so that they are binary compatible with what we had before (see for example `enum`s and `extension`s). It's not a major goal. All else being equal, it might tip the scale. But if it weakens the new design when standing on its own, such argument won't hold up.

lihaoyi:
It's not just "what we had before" for Scala as a language, but "what they had before" for every individual library and callsite.  Even after this feature is released and in the wild for a decade, everything I said still applies.

People using large flat argument sets rarely know what the "meaningful" groupings of those parameters are until *after* they have grown a long list of them which have been out there for a while. That means they will not be able to up-front define the `unpack` case classes until later, and will need to add them later, causing breakage.

You've said yourself you never use large lists of default parameters, so maybe this use case never arises for you. But it has arisen for me many times in the libraries I maintain, which is why I opened this SIP. Every example described in the proposal falls under the category above. I maintain 5-6 libraries which match this exactly. If you don't believe me and don't believe the examples in the proposal that show that this use case is common, there isn't really anything else I can say

sjrd:
> With the status quo, or even with the boxed `unpack` design @sjrd and @odersky have put forward, there is no solution.

I already mentioned the solution: a single, one-time `@binaryAPI private def`. It's that simple. We already have the universal "let me change the API but preserve bin compat" feature. There is no need to burden every other feature with that requirement.

lihaoyi:
Does `@binaryAPI private def` allow us to avoid duplicating parameter lists between the `private def` and the `case class1? The point of this feature is to avoid the need to duplicate parameter lists, so if it just changes the duplication between multiple `def`s to duplication between the `@binaryAPI private def` and the `case class`, that helps but doesn't fully solve the problem

sjrd:
No, you'll need to keep the full signature exactly as it was before you decided to migrate to `unpack`, then never touch it again. So it's a one time thing at the commit where you migrate. Similarly to other similar migrations.

It doesn't solve the duplication that you already had in your to before you migrated, but it won't accumulate *more* duplication after you migrate. Again, that's quite similar to other migration scenarios, for example when you deprecate methods in favor of new ones that are better designed: you have to keep the deprecated signature exactly as it was before your refactoring, but then you can forget about it.

PR #116, opened 2025-08-25
SIP-NN - End markers for method blocks and anonymous classes

soronpo:


content/end-markers-for-method-blocks-and-anonymous-classes.md
```diff
@@ -0,0 +1,207 @@
...
+```scala
+object Foo:
+  def apply(block: => Unit): Unit = ()
+
+Foo.apply:
+  // do something
+end apply
+```
+
+- **Implicit `apply` calls**: Use the name of the object/class that owns the `apply` method when it's called implicitly.
```
prolativ, line 166:
```suggestion
- **Implicit `apply` calls**: Use the name of the object/class instance that owns the `apply` method when it's called implicitly.
```
I believe this was intended for cases like
```scala
class Foo:
  def apply(block: => Unit): Unit = ()

val foo = new Foo

foo:
  // do something
end foo
```

prolativ:
What about curried methods as they were not mentioned explicitly here? E.g. should this work?
```scala
def foo(bar: String)(baz: String) = ???

foo("abc"):
  "xyz"
end foo
```

And similarly
```scala
class CurriedFoo(bar: String):
  def apply(baz: String) = ???
  
def foo(bar: String) = CurriedFoo(bar)

foo("abc"):
  "xyz"
end foo

**Review by odersky — COMMENTED** [review]
End markers for method calls fill an important hole. 

End markers for new are in fact already implemented. This works

```scala
val foo = 
  new Foo:
    // do something
    // overrides, vals, defs...
  end new
```
And so does this:
```scala
val foo = new Foo:
  // do something
  // overrides, vals, defs...
end foo
```
I don't think we need another case were we also allow `end new` in the second case.

I am not sure about disambiguation:

```scala
def foo = bar:
  ...
end bar
end foo
```
is this supposed to work? It would be the first instance where we allow multiple ends on the same indentation level.

content/end-markers-for-method-blocks.md
```diff
@@ -0,0 +1,226 @@
...
+- Allow an end marker that refers to the invoked method, e.g., `end test`, `end locally`, which communicates the intent and closes the application site rather than a definition. See the discussion and viewpoints in `End markers for fewer braces blocks` and follow-on threads [link1] and `More indentation syntax end marker variations` [link2].
+
+Illustrative example:
+
+```scala
+def foo(arg: Int)(block: => Unit): Unit = ???
+
+foo(42):
+  // do something
+  // more nested blocks...
```
prolativ: I think all the examples we give (except for some counterexamples) should be valid scala code (currently or after the SIP would become a part of the language). This snippet below is not one, because fewer-braces syntax requires non-empy code blocks (and comments don't count as code in this sence).
The same applies for some other examples below

soronpo:
Implementation https://github.com/scala/scala3/pull/24251

PR #117, opened 2025-09-18
initial draft for multi-level enums

esarbe:
A draft SIP for https://contributors.scala-lang.org/t/proposal-for-multi-level-enums/3344


content/multi-level-enums.md
```diff
@@ -0,0 +1,213 @@
...
+
+The above syntax desugars to an enum tree with subtype relationships:
+
+```scala
+sealed trait Animal
+
+object Animal:
+  sealed trait Mammal extends Animal
+  object Mammal:
+    case object Dog extends Mammal
```
bishabosha, line 82: to avoid getting into the weeds, because this is the wrong desugaring, its probably better to explain it in a set of recursive steps, where each `case enum` becomes an `enum` in the companion that extends the parent enum class (because an enum is a class not a sealed trait)
see the spec here: https://scala-lang.org/files/archive/spec/3.4/05-classes-and-objects.html#lowering-of-enum-definitions

esarbe: Thanks for the pointer!

I'll look into that and update the spec.

content/multi-level-enums.md
```diff
@@ -0,0 +1,213 @@
...
+### High-level overview
+
+Allow `enum` cases to contain **nested `enum` definitions**, using a consistent indentation-based syntax.
+
+```scala
+enum Animal:
+  case enum Mammal:
+    case Dog, Cat
+  case enum Bird:
+    case Sparrow, Pinguin
```
bishabosha, line 51: also good to demonstrate how parameters work (e.g. `case enum Bird(val call: String):`) and optional extends clauses

content/multi-level-enums.md
```diff
@@ -0,0 +1,213 @@
...
+  case Sparrow => "a bird"
+  case Pinguin => "a penguin"
+  case Fish    => "a fish"
+```
+
+Matching on intermediate enums is also allowed:
+
+```scala
+def isWarmBlooded(a: Animal): Boolean = a match
+  case Mammal => true     // Covers Dog, Cat
```
sjrd, line 122:
```suggestion
  case _: Mammal => true     // Covers Dog, Cat
```

content/multi-level-enums.md
```diff
@@ -0,0 +1,213 @@
...
+  case Bird   => true     // Covers Sparrow, Pinguin
+  case Fish   => false
+```
+
+Matching on a **supercase** (e.g., `Mammal`) is shorthand for matching all its subcases.
```
sjrd, line 127: This does not make sense. Matching on `Mammal` itself already has meaning, which is to test if it is the `Mammal` companion `object` itself. That cannot be changed to means `case Cat | Dog`.

sjrd (commented):
One really thorny question is what type gets inferred for term subcases:
```scala
val x = Animal.Mammal.Dog
```
What is the inferred type for `x`? Choosing `Dog.type` would be completely inconsistent with current enums. But then is it `Animal` or `Mammal`?

---

The SIP does not mention sub *class* cases at all. IMO that's missing.

content/multi-level-enums.md
esarbe: I've added an example.

content/multi-level-enums.md
esarbe: I think I need some help here. Please let me know if that's the wrong approach.

content/multi-level-enums.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### `values`, `ordinal`, `valueOf`
+
+- `Animal.values` returns all **leaf cases**: `[Dog, Cat, Sparrow, Pinguin, Fish]`
+- `Mammal.values` returns `[Dog, Cat]`
+- `Mammal.ordinal` and `Mammal.valueOf(...)` are also available
```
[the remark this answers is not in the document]
esarbe: Global ordinals are weird. Updating the spec for local ordinals.

bracevac:
Since this has been around for some time, we'll discuss it in the next SIP meeting (probably May 22, 2026).

PR #118, opened 2025-09-21
SIP-75: Allow single-line lambda after `:`

odersky:


odersky:
@prolativ Good observation! I changed the proposal and the implementation to admit curried parameter lists.

content/singleLineLambdas.md
```diff
@@ -0,0 +1,135 @@
...
+Seq(1, 2, 3).map: plus1 // error
+```
+
+## Detailed Spec
+
+A `:` means application if its is followed by one of the following:
+
+ 1. a line end and an indented block,
+ 2. a parameter section, followed by `=>` or `=>?`, a line end and an indented block,
+ 3. a parameter section, followed by `=>` or `=>?` and an expression on a single line,
```
prolativ, line 70:
```suggestion
 3. a parameter section, followed by `=>` or `?=>` and an expression on a single line,
```

content/singleLineLambdas.md
```diff
@@ -0,0 +1,135 @@
...
+```scala
+def fun(f: Int => Int => Int): Int = f(1)(2)
+
+fun: (x: Int) => y =>
+  x + y
+```
+
+In the detailed spec above, point (2) is modified as follows:
+
+2. one or more parameter sections, each followed by `=>` or `=>?`, and finally a line end and an indented block.
```
prolativ, line 98:
```suggestion
2. one or more parameter sections, each followed by `=>` or `?=>`, and finally a line end and an indented block.
```

content/singleLineLambdas.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+Single-line lambdas can be nested, as in:
+```scala
+  xs.map: x => x.toString + xs.dropWhile: y => y > 0
+```
+
+## Detailed Spec
+
+A `:` means application if its is followed by one of the following:
```
som-snytt: `s/its/it`

joan38:
We need the ability to do:
```scala
xs.collect: case e: String => e
```
(case on the same line)

And therefore:
```scala
xs.collect(case e: String => e)
```
should work too.

bjornregnell:
@joan38 The parenthesis syntax is a separate SIP: https://github.com/scala/improvement-proposals/pull/113

content/singleLineLambdas.md
```diff
@@ -0,0 +1,145 @@
+---
+layout: sip
+permalink: /sips/into.html
```
prolativ: Wrong permalink. But what should be the right one? @bracevac

bracevac: I updated the YAML to reflect the current SIP template.

bjornregnell:
Not much debate on this...

@bracevac  I think we are ready for a vote.

I recommend accept.

bracevac:
This was accepted by the SIP committee on April 24 2026, with the condition that nested single-line lambdas be excluded.

PR #127, opened 2025-12-03
Have a number based SIP url

hamzaremmal:
I find it better to have a link based on the number rather than the title. We keep also the old links as a redirect.


bracevac (approved):

SethTisue:
Nice improvement! 👏

PR #128, opened 2025-12-03
Tweak number-based permalinks

bracevac:
Support `sips/nn` which redirects to `sips/nn.html` while keeping other redirects (custom and based on the title).


hamzaremmal (commented):
I don't the redirect `/sips/:number` because the issue we are facing has to do with the configuration of the web server and we can't check it until tomorrow. I'm approving because we will at least not get a 404 like we do now.

hamzaremmal (approved):

PR #129, opened 2025-12-31
SIP-78: Relaxed De-dented Case Syntax

lihaoyi:


content/072-relaxed-dedented-case-syntax.md
```diff
@@ -0,0 +1,275 @@
...
+as partial function literals and catch blocks:
+
+```scala
+xs.map:
+case (a, b) => a + b
+case (a, b, c) => a + b + c
+
+val pf: PartialFunction[String, Int] =
+case "foo" => 1
+case "bar" => 2
```
som-snytt, line 33: For the PR, I'll look at 2.1.4 in

https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#optional-braces

where `case` modifies the condition in 2.1.2. It's saying, wherever you could have indented, you can have a `case` block at same indentation.  This example won't work yet.

content/072-relaxed-dedented-case-syntax.md
```diff
@@ -0,0 +1,306 @@
...
+
+```scala
+try ???
+catch // OK
+case (a, b) => a + b
+case (a, b, c) => a + b + c
+```
+
+This means using `case` statements can be confusing: in some scenarios you can de-dent them, in
+other scenarios you cannot, and it is not obvious which is which.
```
JD557, line 83: > it is not obvious which is which.

Seems obvious to me: If there's a `:`, you must indent them.

This fix will add a different inconsistency: after a `:`, sometimes you can de-dent the code (i.e. if there's a `case`) and sometimes you may not.

odersky: I agree, it is not clear this is more orthogonal than what we have. One analogy is established (between match and case-functions) but another rule is broken.

I am also not convinced there's a use case for this. All the examples I saw would be more legible with indentation.

The analogy with if-then-else is weak at best. `else` is aligned with `if` since the two are keywords of the same control stricture, just like `while/do`, or`try/catch`. But for case-functions, there is no leading keyword to align with.

lihaoyi: The issue with this "keyword or not" rule is that Scala in general goes pretty far in making user-definable control structures "first class", using higher order functions to do what many other languages use keywords: things like `with` blocks in Python or `try-with-resources` in Java are higher-order functions in Scala. `test("..."):` blocks, `DynamicVariable#withValue:` blocks, etc. are all similar in that regard

Thus it makes sense that the various syntactic niceties that are afforded to keywords should be afforded to these user-defined control structures. We already saw one iteration of _"we only need to provide short syntax to keywords"_ to _"we definitely need to provide it to user-defined control structures as well"_ with `fewerBraces`. 

It's definitely confusing this is allowed:

```scala
x match
case (a, b) => a + b
case (a, b, c) => a + b + c
```

But this is not:

```scala
xs.map:
case (a, b) => a + b
case (a, b, c) => a + b + c
```

Maybe you think dedented `case`s are bad in general, but you can't really argue that the latter one is any worse than the former. From a user-perspective, `x match` or `xs.map:` are both control structures, and the reasoning you propose 

> _"`match` is a keyword and `:` is technically not because it's punctuation so now you need to add another layer of indentation"_ 

Makes zero sense to a user. Just like how `while ... do` allowing indentation but `myLoop(...){...}` requiring braces was awkward enough that we ended up allowing `myLoop(...):` to use indentation as well: they're all control flow structures and should behave similarly and nobody cares if it ends in a keyword or not

lihaoyi: > But for case-functions, there is no leading keyword to align with.

With `match`/`case`, there is no leading keyword to align with either, because `match` is an infix keyword rather than a prefix keyword and so would not align with the following `case`s. But that doesn't seem to have prevented us from speccing/implementing/using de-dented `case`s in `match` blocks

odersky: That's correct, I was just arguing that any appeal to analogy with if/else does not hold water. 

The proposal aligns case-functions and match case, but breaks the rule that `:` requires an indent to mean `{...}`. So it's not clear it's a win in orthogonality to me.

som-snytt: I was not initially persuaded by the "uniformity" or "orthogonality" argument, but the "no colon after keyword" principle (now foregrounded in 1.1 of the spec) makes me think that `case` is special as a keyword. "Indent is optional for `case`." (As always, a case extends to the next case, or cases end on undent or something that is not a case.) We're only adding that `case` doesn't care if it follows another keyword or a colon. As a neophyte, maybe I also don't care.

The analogy with `if`/`else` justifies the `match` syntax (because cases are labels for flow). Then, I think, Haoyi's argument follows that for "visual harmonization" (to riff a phrase), `case` should behave as described, even for a structure that is semantically disparate (an anonymous function). (Edit: I think Haoyi is arguing something stronger about control structures, for which I invent a different term.)

It's a bit cloudy for "subcases" or what I'm calling "match guards", since you need indents there. Currently, badly aligned cases in match guards are not illegal; hopefully they induce a warning for exhaustivity. My point is that the door is already open for case indentation that is confusing or plain wrong.

JD557: > Scala in general goes pretty far in making user-definable control structures "first class", using higher order functions to do what many other languages use keywords: 

As a middle ground, maybe this could be supported for methods marked as `infix`?

Currently:
```scala
final case class Box[T](x: T):
  infix def myMap[U](f: T => U): Box[U] = Box(f(x))

// Compiles
Box(5) myMap
(x => x + 1)

// Doesn't compile
Box(5) myMap
case x => x + 1
```

If the second case was allowed, one could still define custom control structures and there wouldn't be any confusion regarding statements after a `:` always required an indent.

som-snytt:
This would be fodder for reactionaries who wish to roll back "aligned match", but I noticed that I lodged https://github.com/scala/scala3/issues/22772 which was about error recovery:

```scala
def f(i: Int) =
  i match
  case 42 =>
  i + 1
  case 27 =>
//^^^^
//unindent expected, but 'case' found
  i - 1
  case _ =>
  i
```
My previous comment here suggested that `case 42` extends to `case 27`, which is wrong because there is no closing brace. An alternative approach would suggest that aligned match must be closed by an undent. Then maybe this example just works.

content/072-relaxed-dedented-case-syntax.md
lihaoyi: `infix` is orthogonal to custom control structure or not. `Seq.map:`, `test:`, `withBlah:`, etc. are all custom control structures that are not infix. So limiting this to infix doesn't solve any of the motivating examples

lihaoyi: > The proposal aligns case-functions and match case, but breaks the rule that : requires an indent to mean {...}. So it's not clear it's a win in orthogonality to me.

I think the win in orthogonality is replacing the rule

> `case` doesn't need a level of indentation when inside `match` or `catch`

with 

> `case` never needs its own level of indentation

Similar to what we did earlier in https://github.com/scala/improvement-proposals/pull/113 and https://github.com/scala/improvement-proposals/pull/118, this moves the "special behavior" of `case` from a bunch of special case interactions with surrounding syntax (curlies, `match`, `catch`, etc.) to the belonging to the `case` keyword itself. I'd argue that that is better, since users don't think of `case` expressions as an auxiliary keyword that happens to be shared by a bunch of unrelated language features, but instead as a single language feature _"destructuring/partial function literals"_ and that should behave mostly the same regardless of enclosing syntax (curlies-or-not, indented-or-not, etc.)

This definitely wasn't the case before, and `case` statements had all sorts of odd edge cases such as requiring curlies. But that's where we're going with other `case`-syntax-related SIPs, and this one is in the same direction

kitbellew: what about these, do they require indent:
```
// .match
x.match
case ...
case ...

val partial =
  if foo then
  case ...
  case ...
  else
  case ...
  case ...
```

also, is there's a following unindented select, does it belong to last case, or the entire match:
```
x.foo1:
case ...
.foo2:
...
```

som-snytt: @kitbellew Your examples work. I'll add them to the test. The cases end at non-case, so your `foo1...foo2` are chained.

SethTisue:
I wish we were going in the exact opposite direction, namely to discourage de-dented `case` entirely; I wish it had never been supported in the first place

For three main reasons,

1) I find the readability terrible
2) I find it inconsistent with the rest of the language
3) It allows two different ways of doing things when there should only be one (unless there is a compelling reason to provide two)

som-snytt:
I just commented on a PR that adds automatic formatting.

Many of the automated edits are strictly worse (besides, obviously, using the bad style for Scaladoc asterisks). But the resulting syntax is not outlawed.

I happen to find "match-aligned case" very nice when used tastefully.

It's consistent because `case` was always a special keyword. Witness the people who needlessly put case bodies in braces (as a counter-example showing that case syntax is more beautiful when used to best effect).

Point 3 is lost in the kettle of syntax around "colon" and its conveniences. My reading of the exchange above (where Haoyi argues for uniformity and Martin argues against orthogonality) is that the burden of proof does not lie there; rather, if there is no reason to forbid the syntax (on "technical" grounds, i.e., ambiguity) then it must be allowed.

I also prefer to stack leading infix operators.

bjornregnell:
I see students trying all of these variants:
image

To them it is surprising that the colon variant does not work.

If we are doing "regularity by allowing more" then this might be considered.
Another option is "regularity by allowing less"...

(I think a non-indented sequence of cases look strange as they are "inside" the match or partial function, and "inside" means "indented or braced".)

som-snytt:
This is a less-scrutible message:
```Scala
scala> x match
         case a if a > 0 => a + 1
       case b => b
-- [E040] Syntax Error: --------------------------------------------------------
3 |case b => b
  |^^^^
  |eof expected, but 'case' found
```
Extra indent is OK:
```
scala> x match
       case a if a > 0 => a + 1
         case b => b

val res0: Int = 43
```
No kolon after keyword:
```
scala> try:
-- [E018] Syntax Error: --------------------------------------------------------
1 |try:
  |   ^
  |   expression expected but : found
  |
```
I said to myself, OK, let me restart with `-explain`:
```
scala> try:
-- [E018] Syntax Error: --------------------------------------------------------
1 |try:
  |   ^
  |   expression expected but : found
  |-----------------------------------------------------------------------------
  | Explanation (enabled by `-explain`)
  |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  | An expression cannot start with :.
   -----------------------------------------------------------------------------

scala> var x = 42
1 warning found
-- [E002] Syntax Warning: ------------------------------------------------------
1 |try:
  |^^^^
  |A try without catch or finally is equivalent to putting
  |its body in a block; no exceptions are handled.
  |-----------------------------------------------------------------------------
  | Explanation (enabled by `-explain`)
  |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  | A try expression should be followed by some mechanism to handle any exceptions
  | thrown. Typically a catch expression follows the try and pattern matches
  | on any expected exceptions. For example:
  |
  | import scala.util.control.NonFatal
  |
  | try _root_.scala.Predef.??? catch {
  |   case NonFatal(e) => ???
  | }
  |
  | It is also possible to follow a try immediately by a finally - letting the
  | exception propagate - but still allowing for some clean up in finally:
  |
  | try _root_.scala.Predef.??? finally {
  |   // perform your cleanup here!
  | }
  |
  | It is recommended to use the NonFatal extractor to catch all exceptions as it
  | correctly handles transfer functions like return.
   -----------------------------------------------------------------------------
var x: Int = 42
```
Anyway, `No kolon after keyword` is licensed under creative commons  so feel free to put it on t-shirts, syllabi, etc.

It's also free to use in an `-explain` message for the various syntax productions that may be misused.

bjornregnell:
@som-snytt We will discuss this at the SIP-meeting tomorrow; what do you think about the direction to take in terms of regularity and less surprise on what is allowed or not?

som-snytt:
@bjornregnell To summarize https://github.com/scala/improvement-proposals/pull/129#issuecomment-3751802909,

> I happen to find "match-aligned case" very nice when used tastefully.

Scala does not restrict syntax when it can, but permits syntax when it can.

Scala is "opinionated" in its semantics, such as "use `Conversion` for conversions". So it doesn't go out of its way to create ambiguities, but will add syntax to support a use case, such as `into`.

I don't know how to get more regular than `No kolon after keyword`. Even students must know what a keyword is, and what is a keyword, and what is a soft keyword.

I have come around to "`case` after NL induces indentation" (irregardless of preceding token). (I threw in an irregardless for Seth, but it illustrates that I am catholic in my prejudices.)

I may not use it often, but I suggest it falls under similar questions of alignment:
```
xs.filter(p)
.map(f)

f(xs)
 (ys)
```
where the usage is a matter of style or taste (in the narrow sense, which may be bad taste).

```Scala
xs.map:
case p =>
case q =>
```
What is significant here are the two cases.

Worth adding that (as with a `match` selector) I can move the `fun` to this position without indenting everything that follows:
```Scala
def `one for all`(xs: List[Int]) = xs.map:
  case 42 => "expected"
  case 27 => "unexpected"
  case y => s"weird $y"
```
A typical usage is when inserting a line of debug. This doesn't work:
```
def `one for all`(xs: List[Int]) =
println(s"Ah! $xs")
xs.map:
  case 42 => "expected"
```

bjornregnell:
In summary (before tomorrow's SIP-meeting): This is a debate about regularity and flexibility in syntax. Achieving regularity can be done in two ways: 1) allowing more de-dented cases and 2) dis-allowing some "strange" de-dented cases. The main argument for having syntactic flexibility is that it is more regular. And perhaps it is too much of a braking change to achieve regularity by being more restrictive.

I guess this boils down to syntax opinion, although, objectively, improved regularity is mostly good and flexibility is sometimes bad. 

I'm currently inclined to recommend keeping things as is, as I'm unsure if the wart is big enough... (I agree with @SethTisue that the currently allowed de-dented cases gives significantly less readable code and that it would have been good if they were never allowed. Indented cases are more readable, under the ((in my opinion) good) norm of 2-space indent, but also for the odd 4-space-indent cases.) 

Will we get any more information on this that would make us better post-pone it? @lihaoyi 

@bracevac  Otherwise I think we are ready for a vote.

bjornregnell:
As there are several drawbacks pointed out in the previous discussions, we today decided to postpone this for further discussions at the next SIP meeting.

odersky:
Stepping back, I think Haoyi's proposal to not indent cases makes perfect sense, if `match` is used a a prefix keyword, as in Python (analogously to `switch` in Java):

```scala
match xs 
case x :: xs => 
    ...
case Nil =>
    ...
```
This makes perfect sense. The `case` aligns with the `match`, just like the `then` and `else` align with the `if` in
```scala
if someCondition
then doSomething
else alternative
```
We have the rule that `case` does not need to be indented to emulate that behavior, even though the `match` comes after the selector in Scala. But the emulation breaks down if there is no `match`.

That said, one could ask the question why `match` is not used as a leading keyword, as it would be in other languages. No good reason really, except early Scala history (`match` started as a method on `Object`, and was promoted to keyword later).

jducoeur:
Hmm.  I don't agree with the reasoning, but I think that's partly because I still always write
```
if someCondition then
   doSomething
else
   alternative
```
I'm not sure I've ever used the alignment you're describing for `if` expressions, and at a gut level I dislike it -- I find it harder to parse mentally.

So the two examples aren't equivalent in my mind.  The `if` and `else` are properly parallel to each other, whereas the `case` clauses are subordinate to the `match`.

All very subjective, of course.  But I'd probably never use that dedented approach personally: it feels vaguely "muddied" to me.

sjrd:
These dedented formattings kind of make sense if the introduction comes first in their line. As the examples show:
```scala
match xs 
case x :: xs => 
    ...
case Nil =>
    ...
// and
if someCondition
then doSomething
else alternative
```
Yeah, OK. Fine.

But it doesn't work when there is something in front, like a `val` introduction:
```scala
val y =match xs 
case x :: xs => 
    ...
case Nil =>
    ...
// and
val y = if someCondition
then doSomething
else alternative
```
In Java or Python, the former syntax *is all you ever get*, since `if/else` and `match` are not expressions; they are only statements. In Scala, because they are expressions, and hence almost always appear to the right of something, dedenting their `then/else`s or `case`s prevents appropriate visual chunking and hence increases cognitive load.

odersky:
@sjrd I think it's again analogous to if/then/else. I would also never write
```scala
val x = if bla then
  ...
else
  ...
```
It has to be
```scala
val x = 
  if bla then
    ...
  else
    ...
```
I.e. a statement with a keyword must start a line, unless it all fits in the same line.

sjrd:
I don't understand your point, then. This SIP is precisely about allowing *more* cases where a statement with a keyword *won't* start a line.

Also, maybe you wouldn't do this with `if/then/else`, but you sure do with `match`. Unless you mean that you only write `match`es after the `=` of a `val` because `match` doesn't really start the statement (i.e., the scrutinee comes first)? So
```scala
val y = x match
  case Foo() => bar
```
is fine but
```scala
val y = match x
  case Foo() => bar
```
wouldn't be fine? That doesn't seem likely either, so I'm confused what is the rule you would set for yourself?

odersky:
My post was an argument against accepting the SIP. 

> Also, maybe you wouldn't do this with if/then/else, but you sure do with match. 

That's actually a good point. By analogy, the following could also be ok:
```scala
val x = if condition
  then one
  else two
```
I never used that style, but why not?

So the rule would be: Can use same indent if leading keyword starts a new line, needs additional indent if not. By that rule, the SIP should be rejected since there is no leading keyword in an apply-to-case-block.

To clarify: The rule I gave is a stylistic, not intended to be enforced by the language syntax and parser.

sjrd:
Ah! That makes more sense. In that case I think I agree.

bracevac:
This was rejected in the SIP meeting on May 22, 2026.

PR #133, opened 2026-04-11
SIP-79: Relaxed Null checks under `strictEquality`

mberndt123:
Hi,

I'd like to propose my solution to the following problem to the SIP committee:
https://contributors.scala-lang.org/t/strictequality-with-explicit-nulls-do-not-work-well-together/7050

### DIFF
--- a/content/079-relaxed-null-checks.md
+++ b/content/079-relaxed-null-checks.md
@@ -1,12 +1,12 @@
 ---
 layout: sip
-number: NN
+number: 79
 permalink: /sips/:number.html
 redirect_from:
   - /sips/:title.html
   - /sips/:number
-stage: pre-sip
-status: submitted
+stage: design
+status: under-review
 presip-thread: https://contributors.scala-lang.org/t/strictequality-with-explicit-nulls-do-not-work-well-together/7050
 title: Relaxed Null checks under `strictEquality`
 ---
--- a/content/079-relaxed-null-checks.md
+++ b/content/079-relaxed-null-checks.md
@@ -43,7 +43,7 @@ x match
 
 ## Proposed solution
 
-Do not require a `CanEqual` instance for the `==` and `!=` operators when one operand is `null` and the other operand's type is a supertype of `Null`. Similarly, don't require a `CanEqual` instance during pattern matching when the pattern is `null` and the scrutinee's type is a supertype of `Null`.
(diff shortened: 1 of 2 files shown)


content/079-relaxed-null-checks.md
```diff
@@ -0,0 +1,71 @@
...
+Under `strictEquality`, I propose to not require a `CanEqual` instance when an expression is tested for (in-)equality with `null` and its type is a supertype of `Null`
+This affects the `==` and `!=` operators as well as `case null` clauses in pattern matching.
+
+## Motivation
+
+With the explicit-nulls feature, Scala currently supports a form of flow typing where the type of a nullable variable is refined to a non-nullable type when a suitable non-null check is performed. However this interacts poorly with `strictEquality`, because there is currently no good way to perform a null check for certain types when `strictEquality` is enabled. The type `Int | Null` cannot be compared to `null` using `==` or `!=` because a `CanEqual[Int | Null, Null]` instance is not available. `eq` cannot be used either because that is only available on `AnyRef`, which is not a supertype of `Int | Null`.
+The currently discussed proposal to make `Null` a subtype of `AnyVal` makes `eq` an even less viable alternative.
+
+```scala
+val x: Int | Null = ...
```
soronpo, line 34: I think the example is only missing the explicit nulls and strict equality flags added, so iy will be self contained and runnable (to see the failure that is).

soronpo:
What about `null == x`?. Yes, it's a bit unusual to write, but the proposal currently ignores it.

mberndt123:
I don't think it ignores the `null == x` case. It says this:
> Do not require a `CanEqual` instance for the `==` and `!=` operators when one operand is `null` and the other operand's type is a supertype of `Null`.

It says "one operand", not "the right operand"

content/079-relaxed-null-checks.md
mberndt123: While you can import `scala.language.strictEquality`, I don't think that's possible with explicit nulls. That needs to be enabled using `-Yexplicit-nulls`, so I don't think I can make this self-contained.

soronpo: Use a scala-cli syntax

mberndt123: ok

SethTisue:
"either operand" might help head off any misunderstanding

mberndt123:
done

soronpo:
The average code writer will always favor the LHS to be the argument being compared to null. But for the sake of the reader I would recommend the example to show both just-in-case. You can even leverage to show equality in use as well. E.g. null == y && x!= null

mberndt123:
I think "either operand" is good enough. This SIP isn't exactly rocket science.

soronpo:
Today the SIP committee agreed to move this SIP forward to the experimental phase. Congrats!

PR #135, opened 2026-05-09
SIP-NN: Inline Match Types

soronpo:
Introduces 'inline X match' as a type-level analog of 'inline x match'. Cases are first-match-wins with no disjointness check; the form requires its scrutinee to be concrete at every use site, which prevents the SIP-56 soundness chain (no abstract Foo[X] in signatures means no variance-driven cross-context disagreement). Closes the "specific case first, general fallthrough" UX failure that current match types cannot express without final/sealed hierarchies or summonFrom scaffolding.

### DIFF
--- a/content/NN-inline-match-types.md
+++ b/content/NN-inline-match-types.md
@@ -1,0 +1,517 @@
+---
+layout: sip
+number: NN
+permalink: /sips/:number.html
+redirect_from:
+  - /sips/:number
+  - /sips/:title.html
+stage: pre-sip
+status: under-review
+title: Inline Match Types
+---
+
+**By: Oron Port and Claude (AI)**
+
+## History
+
+| Date          | Version            |
+|---------------|--------------------|
+| May 2026      | Initial Draft      |
+
+## Summary
+
+This SIP proposes **inline match types**, a type-level analog of `inline x match`. The syntax is
+
+```scala
+type Foo[X] = inline X match
+  case P1 => R1
+  …
+  case Pn => Rn
+```
+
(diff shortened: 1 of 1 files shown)


soronpo:
I also have a different proposal of slightly expanding regular match types to allow more sound patterns. Let's discuss this first.

Ichoran:
Yes please!  The kind of type and casting gymnastics one has to use to get regular match types to work in an inline context really adds challenges to achieving correctness and capability.  Often both suffer.

soronpo:
> Yes please! The kind of type and casting gymnastics one has to use to get regular match types to work in an inline context really adds challenges to achieving correctness and capability. Often both suffer.

Yeah, I wonder if we can deprecate regular match types from being used within inline context if this SIP is accepted.

soronpo:
> Yeah, I wonder if we can deprecate regular match types from being used within inline context if this SIP is accepted.

Actually this won't be possible for any usage position that causes problems today.

content/NN-inline-match-types.md
```diff
@@ -0,0 +1,517 @@
...
+The two compose without conflict:
+
+```scala
+transparent inline def first[X](xs: List[X]): X = xs.head
+
+type Codec[X] = inline X match
+  case Int    => IntCodec
+  case String => StringCodec
+
+val c = Codec[first(List(1, 2, 3))]   // first(...) reduces at typer to 1, of type Int
```
bishabosha, line 376: this is invalid syntax

soronpo: > this is invalid syntax

I may pushed the wrong file. I'll check when I get home.

soronpo: FWIW it should be the exact same file, but with almost all the transpanet inline sections removed.

PR #137, opened 2026-07-27
SIP-82 Inline Refinement Captures in Match Type Patterns

halotukozak:
This SIP proposes allowing match type case patterns to capture the abstract type members of a refined type directly, without going through an intermediate type alias.

Today, capturing an abstract type member requires introducing a named extractor alias:

```scala
trait Box:
  type Content

type BoxOf[c] = Box { type Content = c }

type Unbox[B <: Box] = B match
  case BoxOf[x] => x
```

This proposal allows the refinement to be written inline in the pattern instead:

```scala
type Unbox[B <: Box] = B match
  case Box { type Content = x } => x
```

----
Draft implementation: https://github.com/scala/scala3/pull/26381
Contributos thread: https://contributors.scala-lang.org/t/pre-sip-inline-refinement-captures-in-match-type-patterns


content/nnn-inline-refinement-captures-in-match-type-patterns.md
```diff
@@ -0,0 +1,221 @@
...
+
+Multiple members can be captured in a single refinement pattern:
+
+```scala
+trait KV:
+  type Key
+  type Value
+
+type PairOf[K <: KV] = K match
+  case KV { type Key = k; type Value = v } => (k, v)
```
bishabosha, line 124: so i know this one is banned for a type alias pattern (e.g. no `type KV_Key_Value[K0,V0] = KV { type Key = K0; type Value = V0 }`), so implies more power than just a desugaring - unless this rewrites to tuple type of lambdas (or nested matches if you can get the semantics right to simulate a single branch)

halotukozak: you are right. I didn't know that type alias pattern doesn't work. I thought it's a natural generalisation

halotukozak: any rewriting here implies that for a type alias pattern it's also possible

halotukozak: although, it works

```scala
trait TC:
  type A
  type B


type Match[X <: TC] = X match
  case ([a0] =>> TC { type A = a0 })[a] => X match
    case ([b0] =>> TC { type B = b0 })[b] => (a, b)


val tc = new TC:
  type A = Int
  type B = String

val tup: Match[tc.type] = (42, "ok")

println("ok")
```

halotukozak:
Fixes https://github.com/scala/scala3/issues/13416

PR #138, opened 2026-08-04
SIP-NN: Compile-time Specialization

jmesyou:
Pre-SIP thread: https://contributors.scala-lang.org/t/pre-sip-compile-time-specialization/7524/1


content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+class B extends A(true)
+```
+Is inlined to:
+
+```scala
+inline trait A(b: Boolean):
+    def foo(): Int
+
+class B extends A(true):
+    private val A$$b: Boolean = true
```
prolativ, line 247: Shouldn't this SIP specify how exactly these names are mangled (e.g. by explicitly saying that they're prefixed with the name of the inherited inlined trait followed by `$$` or something similar)? What would happen in case of inheritance from two inline traits that have the same name and define an identically named private val but are defined in different packages?

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+    private val A$$x: Int = 1
+    override def foo(): Int = if this.A$$b then this.A$$x.+(1) else 0
+```
+
+#### Inheriting Inline Traits
+Inline traits can not be extended by traits if they have parameters.
```
prolativ, line 253: Maybe something like `Inline traits with parameters cannot be extended by ordinary traits` would be a better wording? Otherwise, `they` is quite ambiguous until one actually reads the example below.

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+def foo(x: Foo[String])
+~~~
+This would normally be allowed with ordinary traits, but the process of specialization creates `Foo$sp$String` in the compilation unit of `B.scala` which is an illegal child of the sealed `Foo[String]`.
+
+We are also faced with the question of whether we should allow the following:
+```scala
+// (2)
+// A.scala
+sealed inline trait Foo[T: Specialized]
+val x = new Foo[Int]() {} // Forces creation of Foo$sp$Int and Foo$impl$Int
```
prolativ, line 351: What's the difference between `Foo$sp$Int` and `Foo$impl$Int`? The latter kind of mangled name is mentioned only in this section of the document but it needs more explanation

jmesyou: `Foo$impl$Int` is the name of the anonymous class type that represents the concrete instance of the specialized trait. I had hoped to spare the details of the implementation behind anonymous class instances from this document but I mistakenly left this artifact (':

content/NNN-compile-time-specialization.md
jmesyou: I can certainly expand upon the topic of name mangling further. Thank you for the suggestion.

content/NNN-compile-time-specialization.md
irfanstract: in Scala 2 it's `Foo$mcI$sp` or `Tuple2$mcII$sp`

prolativ: In this document it should be clearly said what is a part of the specification (to be voted on by the SIP committee) and what is just an implementation detail (and could be changed in the future if needed without breaking compatibility with code using the older implementation)

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+redirect_from:
+  - /sips/:title.html
+  - /sips/:number
+stage: implementation
+status: under-review
+presip-thread: https://contributors.scala-lang.org/t/pre-sip-foo-bar/9999
+title: Compile-time Specialization
+---
+
+**By: James You, Hamish Starling, and Timothee Andres**
```
timotheeandres, line 14: If I may :P
```suggestion
**By: James You, Hamish Starling, and Timothée Andres**
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+*/
+~~~
+This relationship is unable to fully expressed on the JVM so some specialization is lost. 
+Our approach encapsulates specializable semnantics inside traits, which permit multiple inheritance.
+
+### 2. Duplicate class fields
+~~~scala
+class A[T](val x: T)
+
+class A[Int](val x: Int) extends A(x)
```
JD557, line 78: Is this what you meant?

Or was it 
```suggestion
class A_Int(val x: Int) extends A[Int](x)
```

(this might be just me misunderstanding the way the old specialization worked).

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+/* Would require the following inheritance hierarchy when specializing for Int
+     B
+   /   \
+ A     B_Int
+   \   /
+   A_Int
+*/
+~~~
+This relationship is unable to fully expressed on the JVM so some specialization is lost. 
+Our approach encapsulates specializable semnantics inside traits, which permit multiple inheritance.
```
merlinorg, line 72:
```suggestion
Our approach encapsulates specializable semantics inside traits, which permit multiple inheritance.
```

content/NNN-compile-time-specialization.md
prolativ: Some more explanation of the issue would be necessary here.
WIth the change suggested above
```scala
class A_Int(val x: Int) extends A[Int](x)
```
the compiler complains with
```
[error] error overriding value x in class A of type Int;
[error]   value x of type Int needs `override` modifier
[error] class A_Int(val x: Int) extends A[Int](x)
[error]                 ^
```
but indeed it stops complaining when `override` is added
```scala
class A_Int(val x: Int) extends A[Int](x)
```
which seems almost like problem solved besides the fact that this doesn't seem to match the name of the section.
So was the intent really to have two distinct fields named `x` with different types and somehow differentiate between them?

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+~~~
+
+#### Anonymous Specialized Classes
+
+~~~scala
+val v = new Vec[Int] {}
+~~~
+
+Anonymous class instances acting as instances of Specialized traits are subject to the following restrictions:
+  - can extend only a single specialized trait [0],
```
prolativ, line 327:
```suggestion
  - can extend only a single specialized trait,
```
This looks like a typo or some remnants of earlier text

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+#### Variance in Specialized Type Parameters
+Specialized traits may define variance parameters e.g.:
+```scala
+inline trait MyFunction1[-T1: Specialized, +R: Specialized]
+```
+However, variance works in Scala because type parameters are erased, and so we can freely cast e.g. List[Lion] to List[Animal] at runtime. 
+Because specialized traits have a special erasure, The following cases with variance are not supported with specialization:
+
+In the contravariance case:
+~~~scala
```
prolativ, line 404: The unsupported case should be actually described rather than just exemplified

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,488 @@
...
+  rbin.recycle(100)
+
+// Yet, this erases to:
+def recycleAnInteger(rbin: RecyclingBin$sp$Int) = 
+  rbin.recycle(100)
+    
+recycleAnInteger(new RecyclingBin[Any]() {})    // RecyclingBin[Any] can be interpreted as RecyclingBin[Int] due to contravariance
+~~~
+
+We reject these cases at compilation time.
```
prolativ, line 418: This suggests that there were multiple cases mentioned but there was only one mentioned above. And there is only one described below. If this sentence was supposed to refer to both, it should be placed either before or after the two, but to me it seems redundant as it was already stated `The following cases with variance are not supported with specialization`

content/NNN-compile-time-specialization.md
jmesyou: Fixed ( :

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
+---
+layout: sip
+number: NN
+permalink: /sips/:number.html
+redirect_from:
+  - /sips/:title.html
+  - /sips/:number
+stage: implementation
+status: under-review
+presip-thread: https://contributors.scala-lang.org/t/pre-sip-foo-bar/9999
```
prolativ:
```suggestion
presip-thread: https://contributors.scala-lang.org/t/pre-sip-compile-time-specialization/7524/1
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+
+- For a first-time reader, a high-level overview of what they should expect to see in the proposal.
+- For returning readers, a quick reminder of what the proposal is about.
+-->
+
+This proposal presents a pair of complementary features, _inline_ traits and _specialized_ traits, 
+to provide compile-time specialization for Scala 3. 
+This is presented as the following features:
+
+1. New semantics when an `inline` modifier is defined for traits
```
prolativ: This might be understood as if `inline trait` was already possible without with this SIP applied but this SIP would change how such code would behave. Instead, it should be clear that it's actually a new syntactic structure carrying its own semantics (introducing a new features rather than a change)

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+2. The `Specialized` type class, intended to be use in compile-time erased context bounds to generation specializations.
```
prolativ:
```suggestion
2. The `Specialized` type class, intended to be used in compile-time erased context bounds to generation specializations. 
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+
+/* Would require the following inheritance hierarchy when specializing for Int
+     B
+   /   \
+ A     B_Int
+   \   /
+   A_Int
+*/
+~~~
+This relationship is unable to fully expressed on the JVM so some specialization is lost.
```
prolativ:
```suggestion
This relationship is impossible to express on the JVM so some specialization is lost. 
```

content/NNN-compile-time-specialization.md
prolativ: This comment still seems not to have been addressed

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+*/
+~~~
+This relationship is unable to fully expressed on the JVM so some specialization is lost. 
+Our approach encapsulates specializable semantics inside traits, which permit multiple inheritance.
+
+### 2. Unused class fields
+~~~scala
+class A[T](val x: T)
+
+class A[Int](val x: Int) extends A[Int]
```
prolativ:
```suggestion
class A_Int(val x: Int) extends A[Int]
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+~~~
+Consider the example above where the specialization `A[Int]` would need to have a field `x: Int`, but a field
```
prolativ:
```suggestion
Consider the example above where the specialization `A_Int` would need to have a field `x: Int`, but a field
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+~~~ scala
+// This is an @main method
+@main def foo(x: Int): Unit =
+  println(x)
+~~~
+-->
+
+### Inline Traits
+
+An `inline` trait is a trait defined with a `inline` modifier.
```
prolativ:
```suggestion
An `inline` trait is a trait defined with an `inline` modifier.
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+  def scalarProduct(other: Vec[T]): T =
+    require(this.length == other.length)
+    var result = num.fromInt(0)
+    for i <- 0 until length do
+      result = num.plus(result, num.times(this(i), other(i)))
+    result
+~~~
+ 
+Inline traits may be freely extended by objects, classes or other inline traits. 
+An inline trait may also be extended by an ordinary trait, but only if the inline trait does not take parameters
```
prolativ: The missing piece of information is what an inline trait itself might extend

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+
+def first(v: Vec[Int]): Int = Int.unbox(v(0))
+~~~
+
+### Specialization of Inline Traits
+
+We introduce the `Specialized` typeclass as a optional context bound on inline traits and inline methods to address the specific shortcoming with interfaces. 
+
+~~~scala
+sealed trait Specialized[T] extends compiletime.Erased
```
prolativ: If this is going to be a part of the standard library, it should be made clear, in which package this definition will live (which will affect whether it needs to be imported before use or not)

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+
+<!--
+A discussion of how the proposal interacts with other language features. Think about the following questions:
+
+- When envisioning the application of your proposal, what features come to mind as most likely to interact with it?
+- Can you imagine scenarios where such interactions might go wrong?
+- How would you solve such negative scenarios? Any limitations/checks/restrictions on syntax/semantics to prevent them from happening? Include such solutions in your proposal.
+-->
+
+Our proposal does not require any changes to existing syntax.
```
prolativ: Even though we don't need to update the EBNF grammar for the parser, allowing a new modifier for a keyword (`inline trait`) is arguably a kind of a syntax change

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+inline trait A[T](x: T):
+    val y = x
+trait B extends A[Int]
+class C extends A[Int](10), B
+~~~
+
+After inlining, `B.y` is defined in terms of `B.A$$x` (the inlined copy of the parameter accessor of `x`), but this is undefined as we don't have the parameter value in `B` due to the trait parameter passing rules that insist that `C` passes the parameters directly.
+
+#### Inline Members & Inline Traits
+Inline traits may define inline members (e.g. `inline def`, `inline val`). These defs are inlined as the body of the trait is inlined into the inherting type,
```
prolativ:
```suggestion
Inline traits may define inline members (e.g. `inline def`, `inline val`). These definitions are inlined as the body of the trait is inlined into the inherting type, 
```
Assuming this applies to `val`s as well

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+    val y = x
+trait B extends A[Int]
+class C extends A[Int](10), B
+~~~
+
+After inlining, `B.y` is defined in terms of `B.A$$x` (the inlined copy of the parameter accessor of `x`), but this is undefined as we don't have the parameter value in `B` due to the trait parameter passing rules that insist that `C` passes the parameters directly.
+
+#### Inline Members & Inline Traits
+Inline traits may define inline members (e.g. `inline def`, `inline val`). These defs are inlined as the body of the trait is inlined into the inherting type, 
+but the members themselves are not inlined and are deleted from the parent trait.
```
prolativ:
```suggestion
and they are deleted from the parent trait.
```
unless the meaningfulness of this sentence relies on some subtle distinction between a `def(inition)` and a member, which might not be obvious here

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+  inline val z = x // Not ok
+~~~
+
+#### Anonymous Specialized Classes
+
+~~~scala
+val v = new Vec[Int] {}
+~~~
+
+Anonymous class instances acting as instances of Specialized traits are subject to the following restrictions:
```
prolativ:
```suggestion
Anonymous class instances acting as instances of specialized traits are subject to the following restrictions:
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+  - can extend a single specialized trait,
+  - cannot mix in further classes or traits, and
+  - cannot contain member definitions.
+
+#### `sealed` Specialized Traits
+
+While Specialized traits may be `sealed`, the generated `$sp$` traits and implementation classes extending the `sealed` specialized trait may potentially be from another compilation unit.
```
prolativ:
```suggestion
While specialized traits may be `sealed`, the generated `$sp$` traits and implementation classes extending the `sealed` specialized trait may potentially be from another compilation unit. 
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+```scala
+// (2)
+// A.scala
+sealed inline trait Foo[T: Specialized]
+val x = new Foo[Int]() {}
+
+// B.scala
+val y = new Foo[Int]() {}
+```
+In the bytecode `x` and `y` may or may not point to the same `Foo$impl$Int` depending on if we share the generated specialized classes, but in the source code `y` extends `Foo[Int]` illegally.
```
prolativ: Here an example of an `$impl$` class jumps out of nowhere, which is confusing. If the generation of `$impl$` classes is a part of the SIP, it should be clearly explained upfront. Otherwise, if it's just an implementation detail, which should be hidden, let's not rely on it in the specification

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+
+```scala
+def foo(x: List[Double]): Unit = x match {
+  case xs: :+:[_] => println("Cons case")
+  case _: Nil[_] =>  println("Nil case")
+  // warning: non-exhaustive pattern match, missing case _: List[Double] & List$sp$Double
+}
+```
+
+We don't register the compiler generated specialized traits as children of the original specialized trait for exhaustivity checking. This is safe because these traits are synthetic and the following invariant `T <:< Foo[Int] <=> T <:< Foo$sp$Int`, so users cannot match on `Foo$sp$Int` or `Foo[Int] minus Foo$sp$Int`
```
prolativ:
```suggestion
We don't register the compiler generated specialized traits as children of the original specialized trait for exhaustivity checking. This is safe because these traits are synthetic and the following invariant holds: `T <:< Foo[Int] <=> T <:< Foo$sp$Int`, so users cannot match on `Foo$sp$Int` or `Foo[Int] minus Foo$sp$Int`
```

content/NNN-compile-time-specialization.md
```diff
@@ -0,0 +1,495 @@
...
+def recycleAnInteger(rbin: RecyclingBin$sp$Int) = 
+  rbin.recycle(100)
+    
+recycleAnInteger(new RecyclingBin[Any]() {})    // RecyclingBin[Any] can be interpreted as RecyclingBin[Int] due to contravariance
+~~~
+
+We reject this case at compilation time. 
+We impose an additional restriction 
+
+In the covariance case:
```
prolativ: Same as for the contravariant case: explain the restriction rather than just give an example

